PR #35, opened 2014-08-23
Added a reset macro to allow the buffer to be reused

wesrobb:
... without (re)allocs until the old capacity is exceeded.

Typically this is useful when you have some approximate upper limit of items. The buffer grows to a point where no more allocs are required (e.g. holding data generated during each frame of a game).


nothings:
Current plan is to make a new sblib.h which contains stretchy_buffer API but full stb_arr set, and also include stb_sdict, stb_file, stb_readdir, and a few other must-haves from stb.h

akaltar:
Is there still a plan on creating sblib.h? I'm looking for these features, and I would be glad to help.

rygorous:
There's now a data structures-only header file in `stb_ds.h`, which supports a way to reset a dynamic array without freeing the allocation (using `arrsetlen(array, 0)`), so I'm tentatively closing this PR.

PR #39, opened 2014-09-07
Added support for Portable Gray Map (PGM) and Portable Pixel Map (PPM)

kennethdmiller3:
I tried to match stye style of the rest of the library and use existing functionality where possible.
This implementation doesn't handle header comments or the ASCII variants (P2 and P3)


cheako:
I'm looking to add ascii support for my project.  Is there any reason it wasn't implemented? 

I can see that adding support for comments in the header wouldn't take much, I may start with that as it's essentially a requirement for P3 and P2.

kennethdmiller3:
Yes, though they weren't necessarily _good_ reasons:
- I only needed binary to read the game source assets I was processing at the time
- parsing ASCII image data was a lot more work than just reading N * W * H bytes into the buffer
- I assumed ASCII PGM and PPM were much less common than binary
- I didn't have any ASCII-format PGM or PPM images to work with

PR #41, opened 2014-09-09
Add .hdr file writing support

baldurk:
I've added .hdr radiance RGBE support to stb_image_write.h since it could be useful for dumping an hdr screenshot which photoshop or other editors could open, and the encoding isn't too onerous.

I tried to match the code style and follow stb guidelines, I've tested it such as I can (I can't seem to find an image viewer that will open >32768 width files to test that case but inspecting the data loaded by stb_image.h by hand looks fine), and it compiles on what I have available - gcc, clang on linux and VS2010.

If one really wanted to it'd be possible to avoid the malloc() for a scanline but it would make the RLE much more complex I suspect, so I didn't consider it worthwhile.

### DIFF
--- a/stb_image_write.h
+++ b/stb_image_write.h
@@ -24,11 +24,12 @@ ABOUT:
 
 USAGE:
 
-   There are three functions, one for each image file format:
+   There are four functions, one for each image file format:
 
      int stbi_write_png(char const *filename, int w, int h, int comp, const void *data, int stride_in_bytes);
      int stbi_write_bmp(char const *filename, int w, int h, int comp, const void *data);
      int stbi_write_tga(char const *filename, int w, int h, int comp, const void *data);
+     int stbi_write_hdr(char const *filename, int w, int h, int comp, const void *data);
 
    Each function returns 0 on failure and non-0 on success.
    
@@ -51,6 +52,10 @@ USAGE:
    formats do not. (Thus you cannot write a native-format BMP through the BMP
    writer, both because it is in BGR order and because it may have padding
    at the end of the line.)
+
+   HDR expects linear float data. Since the format is always 32-bit rgb(e)
+   data, alpha (if provided) is discarded, and for monochrome data it is
+   replicated across all three channels.
 */
 
 #ifndef INCLUDE_STB_IMAGE_WRITE_H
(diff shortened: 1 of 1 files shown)


nothings:
Can you check this, since I don't have any test cases?

"there is a bug in HDR writing code.

stb_image_write.h, line 309:
            } else if (head - runstart == 128 && rlerun == 0) {

it checks length for 128 and then passes it into
stbiw__write_nonrle_data, but this function asserts for length <=
0x7f.

Looks like this check on line 309 should be
            } else if (head - runstart == 127 && rlerun == 0) {
"

I added the assert exactly to catch this sort of thing.

baldurk:
I remember running into this when writing the code. [The spec](http://radsite.lbl.gov/radiance/refer/filefmts.pdf) isn't explicit on whether 128 is a valid length or not, it just says "A run is indicated by a byte with its highorder bit set, corresponding to a count with excess 128. A nonrun is indicated with a byte less than 128.". Which leaves 128 a bit ambiguous, it's just implied that 128 is then an invalid value corresponding to a 0 length run.

When I looked around, it seemed the convention was to treat >128 as a run, and <= 128 as a non-run. That matches the behaviour in stb_image which [treats a length of 128 as non-RLE](https://github.com/nothings/stb/blob/master/stb_image.h#L5829). At least one test file (%DXSDK%\Samples\Media\Light Probes\uffizi_cross.hdr) I found had non-runs of 128 length and photoshop seems happy with that, but at best that's just more anecdotal convention.

If you made the suggested change though, it'd still be perfectly valid AFAIK and just have slightly shorter non-runs, so that seems like a good change to me.

Sorry about that!

nothings:
While trying to test the edge cases, I found a minor bug in the rle logic (that probably never would cause wrong output) and also found it a bit too confusing for my way of thinking, so I rewrote the rle from scratch to work a different way. (Though once I could debug it, it turned out to be that most of my issues were because I had screwed up my test case.)

I tested the RLE logic pretty carefully, but I only tested one synthetic image back through the decode pieline, so If you could test the new version I'd appreciate it.

baldurk:
I tested it by saving out some render targets and things of various types that would include RLEable and non-RLEable data. It all seems to work fine, saving as well as loading via stb_image.h and external viewers. I had some compile warnings when updating to the latest stb_image.h which I listed in #74 but that's unrelated.

PR #67, opened 2015-01-06
stb_truetype: split stbtt_PackFontRanges() into 3 functions for advanced uses

ocornut:
The reasoning is: If I want to pack multiple fonts together or custom graphics data along with a font, right now the user needs to make a very customized copy of stbtt_PackFontRanges() - 107 lines, possibly breaks with update.

So I split stbtt_PackFontRanges() and created two extra functions called by it.

    int stbtt_PackFontRangesGatherRects(stbtt_pack_context *spc, stbtt_fontinfo *info, stbtt_pack_range *ranges, int num_ranges, stbrp_rect *rects);
    int stbtt_PackFontRangesRenderIntoRects(stbtt_pack_context *spc, stbtt_fontinfo *info, stbtt_pack_range *ranges, int num_ranges, stbrp_rect *rects);

stbtt_PackFontRanges is now 32 lines and is conceptually a high-level helper, contains the essential for packing a single font into a texture. Easy to copy to adapt for more advanced packing.

The new function names or parameters may be renamed if you have a more suitable idea.

### DIFF
--- a/stb_truetype.h
+++ b/stb_truetype.h
@@ -2320,14 +2320,19 @@ int stbtt_PackFontRangesGatherRects(stbtt_pack_context *spc, stbtt_fontinfo *inf
       float scale = fh > 0 ? stbtt_ScaleForPixelHeight(info, fh) : stbtt_ScaleForMappingEmToPixels(info, -fh);
       for (j=0; j < ranges[i].num_chars_in_range; ++j) {
          int x0,y0,x1,y1;
-         stbtt_GetCodepointBitmapBoxSubpixel(info, ranges[i].first_unicode_char_in_range + j,
-                                             scale * spc->h_oversample,
-                                             scale * spc->v_oversample,
-                                             0,0,
-                                             &x0,&y0,&x1,&y1);
-         rects[k].w = (stbrp_coord) (x1-x0 + spc->padding + spc->h_oversample-1);
-         rects[k].h = (stbrp_coord) (y1-y0 + spc->padding + spc->v_oversample-1);
-         ++k;
+		 int glyph = stbtt_FindGlyphIndex(info,ranges[i].first_unicode_char_in_range + j);
+		 if (glyph) {
+            stbtt_GetGlyphBitmapBoxSubpixel(info,glyph,
+                                            scale * spc->h_oversample,
+                                            scale * spc->v_oversample,
+                                            0,0,
+                                            &x0,&y0,&x1,&y1);
+            rects[k].w = (stbrp_coord) (x1-x0 + spc->padding + spc->h_oversample-1);
+            rects[k].h = (stbrp_coord) (y1-y0 + spc->padding + spc->v_oversample-1);
+		 } else {
+			 rects[k].was_packed = false;
+		 }
+		 ++k;
       }
    }
 
--- a/stb_truetype.h
+++ b/stb_truetype.h
@@ -2165,7 +2165,8 @@ int stbtt_PackBegin(stbtt_pack_context *spc, unsigned char *pixels, int pw, int 
 
    stbrp_init_target(context, pw-padding, ph-padding, nodes, num_nodes);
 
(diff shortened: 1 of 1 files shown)


nothings:
I haven't looked at the code here yet because I don't understand one of the issues. You can pack multiple fonts by making multiple calls to stbtt_PackFontRanges. It's not optimal, but it's not terrible either.

ocornut:
Alright, I had completely overlooked the fact that it could be called multiple times and stbrp_pack_rects as well). I will measure the difference between the two approaches. Dealing with CJK sets here (21 K characters) so aiming for optimal probably a good idea. But I imagine with power-of-two sized textures at that sort of size the difference won't matter in practice.

So this probably can be closed. I can use an extra call to stbrp_pack_rects() with the same context to get space from the atlas prior to adding my fonts. It works.

PS: I haven't verified but by looking at it stbtt_PackFontRanges() seems to leak the rects it allocates.

ocornut:
I found another reason for separating the functions, that is to optimally do a pass for finding a texture size and then a pass for rendering. Right now stbtt_PackFontRanges() assume enough texture space and the user can iteratively increase the texture space but that's lots of wasted time. If you are working with large number of glyphs this is becoming very meaningful.

In the first pass one can give infinite / max height to the packing algorithm, and then create the pixels buffer according to the packing output. 

Pass 1. 
call stbtt_PackBegin() with atlas size but NULL pixels.
gather rectangles and pack
Pass 2
allocate pixels
render
stbtt_PackEnd() 

It would only make sense if you use a customized version of stbtt_PackFontRanges().

ocornut:
Added two more commits. 

- The first allows passing NULL pixels (the implement what I have described above, works well).

- The second is perhaps more arbitrary (but reasonable I think) change. If stbtt_FindGlyphIndex() returns zero the glyph is not rendered and marked as not packed. Previously if you gave it a range of say 0-255 with several fonts you would end with a bunch of empty rectangles outline. When dealing with large ranges this would have a noticeable cost. 

I don't know TTF enough to tell if glyph 0 has a specific meaning or if it can exist aside from an error case.. Another approach for that last commit would be to render glyph 0 once and have multiple occurrences of it point to the same rectangle. Not too hard to implement (maybe another flag in stbrp_rect and a pass copying data after packing). I'm happy to implement this if you think this strategy makes more sense.

nothings:
Yeah, glyph id 0 is the normal rectangle 'substitute this character for missing characters' character, so the right thing would indeed be to render it once and reuse it. But have to actually do that separately for each font and font size, which would probably be tricky?

ocornut:
Can be done locally in the loop calling stbtt_GetGlyphBitmapBoxSubpixel() by recording the index of the rectangle for the first 0 glyph, per font size. That would be a dynamically allocated array but most likely 1-sized and never absurdly big, and the cost for searching in that array will always be lower than the cost of rendering another glyph 0 anyway.

First occurrence of glyph 0 for a given font size adds to array. Subsequent occurrence set a new field in stbrp_rect to refer to the index of the first rectangle (stored as a relative offset from the current rectangle). Other fields are zero-ed to be ignored by packer.

The rendering loop test for the new field, if non-zero copies the stbtt_packedchar structure.

May be overkill to handle an error case. I am personally happy with them not being rendered, we already get the feedback that a character "failed" when everything in stbtt_packedchar is zero.

ocornut:
Pretty embarrassing, the handling for missing glyph was broken. Setting 'was_packed' is unnecessary because it is overwritten by stbrp_pack_rects() anyway. It worked but uninitialized width/height fed into packer. Setting width/height to 0 seems to break it (haven't investigated), width/height 1 works but that's waste. I'll fix that properly later and will submit a new PR without all the intermediary broken commits.

ocornut:
Opened another PR https://github.com/nothings/stb/pull/72
Closing this. Don't understand how this can be updated better with GitHub.

PR #71, opened 2015-01-13
stb_image_write.h: Add support for writing through callback functions

ejulien:
I tried to respect what was done in stb_image.h as much as possible with a few modifications so that stb_image.h and stb_image_write.h can be used in the same compilation unit.

### DIFF
--- a/stb_image_write.h
+++ b/stb_image_write.h
@@ -19,16 +19,23 @@ ABOUT:
 
    The PNG output is not optimal; it is 20-50% larger than the file
    written by a decent optimizing implementation. This library is designed
-   for source code compactness and simplicitly, not optimal image file size
+   for source code compactness and simplicity, not optimal image file size
    or run-time performance.
 
 USAGE:
 
-   There are three functions, one for each image file format:
+   There are functions for each image file format and output method:
 
-     int stbi_write_png(char const *filename, int w, int h, int comp, const void *data, int stride_in_bytes);
-     int stbi_write_bmp(char const *filename, int w, int h, int comp, const void *data);
-     int stbi_write_tga(char const *filename, int w, int h, int comp, const void *data);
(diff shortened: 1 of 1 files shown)


nothings:
I think what might make more sense for this library is to support a "write to memory" path, and then if you want to redirect I/O you just write it to memory and then write it out yourself. Then the size of the API would only double, AND it would provide a write-to-memory path that some people probably want anyway.

The only downside of this would be on, say, embedded platforms that can't afford the memory. But since the size of the memory is in the same ballpark as the image itself, and you have to store the image itself, it seems like this limitation wouldn't be _that_ restrictive.

ejulien:
I too am concerned with the explosion of the API function count in the current patch.

However it looks like it is a very good (if not the best) option if you choose to support custom I/O:
* The impact on existing code is minimal.
* It supports any kind of output without the drawback you mentioned (another problem of which is memory fragmentation unless you overshoot or have a very good output size estimate).

I can rework the patch toward something less messy like:

```c++
int stbi_write_XXX(stbi_io_write_callbacks const *clbk, ...);

int stbi_write_initialize_file_callbacks(stbi_io_write_callbacks, const char *filename);
int stbi_write_dispose_file_callbacks(stbi_io_write_callbacks);
```

Would that be better? If not can you give me direction on how you'd like to have that feature done?

nothings:
Let me think about it some more, as it'd be better to not waste the work you've already done.

nothings:
I decided I've been sitting on this for long enough, so I went ahead and integrated it as is.

Brought it up to date with the current version (which mostly involved starting with 0.99 and re-porting it to your stuff), rewrote the API to get rid of the unnecessary struct, renamed all the functions (bad to break working client code by renaming the existing functions to '_to_file'), and added back in support for handling when fopen() fails,

I've tested the "to_file" paths, but I only cursorily tested the io callback path.

ejulien:
Great, I can switch back to the official fork. Thanks!

PR #72, opened 2015-01-18
stb_truetype: split stbtt_PackFontRanges() into 3 functions for advanced uses + allow stbtt_PackBegin() to take NULL pixels input

ocornut:
This is a cleaned version of https://github.com/nothings/stb/pull/67

It includes two changes:

- Part of stbtt_PackFontRanges() have been split into 2 extra functions (no functional changes)

NB: One call to stbtt_GetCodepointBitmapBoxSubpixel() has been replaced by a call to stbtt_FindGlyphIndex() followed by stbtt_GetGlyphBitmapBoxSubpixel() - which is identical to calling the earlier function. The only reason for that is to facilitate merging of changes involving different handling of glyph 0.

- stbtt_PackBegin() silently allows NULL pixels to be passed (doesn't try to memset the contents)
This is the only change of behaviour here, previously it was invalid input and would crash, now it would just ignore it. 

Those changes are to allow packing in two passes as discussed in https://github.com/nothings/stb/pull/67
One pass to do packing and extend texture size, another pass to do the actual rendering.
The changes shouldn't affect existing code, merely provide individual calls for what was all done in 1 function before.

The PR doesn't include any change of behaviour when dealing with glyph 0 (which was discussed in the other thread but will be tackled separately).


ocornut:
FIxed missing forward declaration of typedef stbrp_rect if stb_rect_pack.h wasn't included before stb_truetype.h

ocornut:
Merged with latest.

ocornut:
Thanks for merging that and adding the docs!
Btw to clarify the primary point of this is to measure the texture size. AFAIK it's not possible to decide of ideal texture size with the normal texture API. So I pass an arbitrary width (e.g. 512) and an infinite-ish (very large) height when packing, then decide on proper texture height, allocate my texture, run the rasterizer.

PR #78, opened 2015-01-30
stbte: layer name buttons grow with layer panel

### LINKED ISSUE #47: stbte_create or stbte_create_map
The declaration is
```C++
extern stbte_tilemap *stbte_create(int map_x, int map_y, int map_layers, int spacing_x, int spacing_y, int max_tiles);
```
The implementation is
```C++
stbte_tilemap *stbte_create_map(int map_x, int map_y, int map_layers, int spacing_x, int spacing_y, int max_tiles)
{
...
}
```

machinamentum:
Layername buttons will grow/shrink to fill layer panel width. Grows to a maximum of the draw width of the widest set layer name.
Also fixes #47


nothings:
Released version is slightly tweaked and chose a different version number, but effectively unchanged.

PR #87, opened 2015-03-20
Allow to replace memmove function with custom implementation

mmozeiko:
Similarly how stb_truetype.h allows you to replace any stdlib.h, math.h or memory.h function with custom implementation, allow also to replace memmove in stb_textedit.h.


nothings:
Renamed STBTE_memmove to STB_TEXTEDIT_memmove to avoid conflicts with stb_tilemap_editor, which uses the STBTE_ prefix.

PR #88, opened 2015-03-20
Allow to replace qsort function with custom implementation

mmozeiko:
Similarly how stb_truetype.h allows you to replace any stdlib.h, math.h or memory.h function with custom implementation, allow also to replace qsort in stb_rect_pack.h.


nothings:
Renamed the definition to STBRP_SORT instead of STBRP_sort for consistency with STBRP_ASSERT.

PR #89, opened 2015-03-21
added stbi_vertical_flip

nguillemot:
makes it easier to get started quickly with GL.
strawman implementation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,20 @@ attribution requirement). They may be less featureful, slower,
 and/or use more memory. If you're already using an equivalent
 library, there's probably no good reason to switch.
 
+#### Why single-file headers?
+
+Windows doesn't have standard directories where libraries
+live. That makes deploying libraries in Windows a lot more
+painful than open source developers on Unix-derivates generally
+realize. (It also makes library dependencies a lot worse in Windows.)
+
+Making them a single file makes it very easy to just
+drop them into a project that needs them. (Of course you can
+still put them in a proper shared library tree if you want.)
+
+The difference between 10 files and 9 files is not a big deal,
+but the difference between 2 files and 1 file is a big deal.
+
 #### Why "stb"? Is this something to do with Set-Top Boxes?
 
 No, they are just the initials for my name, Sean T. Barrett.
--- a/tools/README.footer.md
+++ b/tools/README.footer.md
@@ -20,6 +20,20 @@ attribution requirement). They may be less featureful, slower,
 and/or use more memory. If you're already using an equivalent
 library, there's probably no good reason to switch.
 
+#### Why single-file headers?
+
+Windows doesn't have standard directories where libraries
+live. That makes deploying libraries in Windows a lot more
+painful than open source developers on Unix-derivates generally
+realize. (It also makes library dependencies a lot worse in Windows.)
+
+Making them a single file makes it very easy to just
+drop them into a project that needs them. (Of course you can
+still put them in a proper shared library tree if you want.)
+
+The difference between 10 files and 9 files is not a big deal,
+but the difference between 2 files and 1 file is a big deal.
+
 #### Why "stb"? Is this something to do with Set-Top Boxes?
 
 No, they are just the initials for my name, Sean T. Barrett.


nothings:
Heh. I guess that's one way to do it.

nguillemot:
I was thinking of adding another flag like `stbi_convert_iphone_png_to_rgb`, but couldn't find a good place to put it that wouldn't violate the sanctity of `stbi_load_main`.

There's probably opportunities for optimization if the flipping algorithm can know about the image format. For example, if a tga image is already being stored from bottom to top, then the image doesn't have to be inverted.

nothings:
Yeah, most decoders could be changed to have an explicit stride, and then if you're doing bottom-to-top you just set the stride to negative and the starting pointer to the bottom left corner. (The fact that this DOESN'T work in OpenGL is an embarassment.)

So yeah, since the goal of stb_image 2.0 and on is to be full-featured, robust, *and* competitive efficiency, we probably shouldn't say "eh, if you want vertical-flip or BGR we make a whole separate pass over the image".

nguillemot:
Tempted to give a stab at adding vertically flipped image decoding by modifying the decoding code itself maybe this weekend or something. Like you mentioned, it's probably not always possible...

nothings:
What I actually care more about is getting the API right rather than getting the efficient implementation in there the first time.

In other words, if we just add a "postprocessing" step to the top-level API which checks for flags and vertically flips and/or does RGB->BGR (and RGBA->BGRA,ABGR,ARGB), that gets the API right, and isn't any slower than the current situation (where the user has to do such post-processing themselves).

Then we can investigate on a case-by-case basis moving that logic down into the decoders as desired.

nguillemot:
Are you thinking of adding flags like `stbi_set_unpremultiply_on_load` and `stbi_convert_iphone_png_to_rgb`? I'm worried about that design because it makes it hard to use stb_image on multiple threads...

nothings:
Can you think of a realistic circumstance where you would want a different setting for vertical-flip or rgba/bgra across multiple threads? I'd prefer not to have to change the API or add a new API function, which would be necessary to add a flags parameter with the C API .

nguillemot:
Since `req_comp` only currently stores 0,1,2,3,4 which takes 3 bits, we could switch the name of that parameter to `flags` and use the rest of the bits for other options.

    enum
    {
        STBI_default = 0,
        STBI_grey = 1,
        STBI_grey_alpha = 2,
        STBI_rgb = 3,
        STBI_rgb_alpha = 4,
        STBI_premultiply_on_load = 8,
        STBI_convert_iphone_png_to_rgb = 16,
        STBI_vertical_flip = 32,
        etc
    };

nothings:
Very ugly! In the unlikely circumstance the user needs more than one configurable variant at once, they can use STB_IMAGE_STATIC to create multiple variants each with their own set of global variables.

nguillemot:
>they can use STB_IMAGE_STATIC to create multiple variants each with their own set of global variables.

I think that passing a bitset is less ugly than having to make N different .c files that include stb_image in order to duplicate its symbols N different times... Especially since that N is fixed at compile time, so you can't launch an arbitrary number of threads.

nothings:
Sure, *use* of the API is far more ugly in the case of needing to make multiple, but we are talking about 0 real-world cases where we know this is needed. As long as there is *some* path for it, even if that path is arbitrarily complicated, I'm ok with it.

Here's what I meant by "ugly": (1) *changing* the semantics of the existing API is ugly and confusing. E.g. people don't get link errors if they call with the new parameters to an old version of the library. (2) A single parameter which is a mix of both an enum and flags is pretty ugly, especially when the existing version of the parameter isn't actually an enum but rather a channel count. It's write-only so you don't have to extract the channel count or such so it's not *terrible*, but it's ugly.

Note that the-multi-file version with N threads, you don't need N variants of it, you only need as many variants as you use simultaneously in your app. If you use both BGRA and ARGB but not vertical flip, you need 2 variants. If you use both vertical flip & non, but only RGBA, you also need 2 variants.

nguillemot:
> Sure, use of the API is far more ugly in the case of needing to make multiple, but we are talking about 0 real-world cases where we know this is needed. As long as there is some path for it, even if that path is arbitrarily complicated, I'm ok with it.

As more flags get added, there will probably be lots of cases where you want to use different options for different types of assets that you might want to load in parallel, especially given how much easier it is to parallelize graphics work with new 3D APIs coming out. At present, it's not safe to assume that you can use `stbi_load` in multiple threads at all. If you want to use stbi to do the image loading in a big existing code base and want to code defensively for the future, you have to lock the whole API under a mutex, which serializes the whole API like those old-school needlessly stateful standard C functions.

> (1) changing the semantics of the existing API is ugly and confusing. E.g. people don't get link errors if they call with the new parameters to an old version of the library.

Well, at least they'll get caught at runtime thanks to `STBI_ASSERT(req_comp >= 1 && req_comp <= 4);`

> (2) A single parameter which is a mix of both an enum and flags is pretty ugly, especially when the existing version of the parameter isn't actually an enum but rather a channel count. It's write-only so you don't have to extract the channel count or such so it's not terrible, but it's ugly.

You've already crossed that line, those enums are already there! I just added the other ones to the list. https://github.com/nothings/stb/blob/master/stb_image.h#L386

>BTW note that the-multi-file version with N threads, you don't need N variants of it, you only need as many variants as you use simultaneously in your app.

That's even worse! If you want to be safe, you need to create 2^N different variants (with N flags either on or off). If you look at the SOIL2 library, it has 13 flags. 2^13 is massive, and even just setting up a subset of those 2^13 is a lot of work for something that has no technical reason to exist, and I wouldn't be surprised if people confronted with that task just say "f*ck it" and switch to a different image loading library.

nothings:
>As more flags get added, there will probably be lots of cases where you want to use different options for different types of assets that you might want to load in parallel, especially given how much easier it is to parallelize graphics work with new 3D APIs coming out.

I've added one pair of flags in the last seven years. I've been aware of the possibility of BGR and vertical flip for basically that entire time. I don't anticipate "more flags getting added" past the set we're talking about; I have no idea what those flags would be or where they would come from, given how long the ones we're proposing have been on the table.

Heck, I could choose not to even add the BGR & vertical_flip flags after all, I mean, we've gone this long without them. They're not mandatory. So any such speculative flags I could also choose not to add. Or, if I wanted to add them, they could be added the same way, or, if they were highly likely to need to vary simultaneously, I could THEN change this API to support flags; or I could add another API function that adds flags (how I've done it for other stb libs); or I could change the state from being a global to being a thread-local.

Anticipating future API changes to avoid future back-incompatibility is a sane action, but I just don't think it applies to stb_image in the way you're talking about.

>At present, it's not safe to assume that you can use stbi_load in multiple threads at all. If you want to use stbi to do the image loading in a big existing code base and want to code defensively for the future, you have to lock the whole API under a mutex, which serializes the whole API like those old-school needlessly stateful standard C functions.

This is theoretically true but isn't true in practice. Literally nobody has every complained about the global settings that already exist, whether they be the flags that your proposal would "fix" (e.g. de-iphone), or values which your proposal doesn't fix (like global gamma settings for the gamma handling of hdr-to-ldr and ldr-to-hdr conversion).

Many STB libraries have *compile*-time configuration through macros. This is done under the assumption that in normal use, you don't need multiple versions of them. If you *do* need multiple variations, then you have to use the STATIC thing and make multiple versions. We *could* go down that exact same path for stb_image for BGR/vertical flipping; it would be consistent and *nearly* as functional. The choice of global flags instead of compile time is actually a slight increase in functionality over how similar things are handled in other stb libs.

nothings:
(And SOIL is not an image-to-pixels library; SOIL is an image-to-texture library (with a platform dependency). All 13 of the flags it has are the sort of things that arise dealing with textures, not dealing with pixels.)

nguillemot:
making the flags thread local seems like reasonable solution to me...

nothings:
Unfortunately, that's (a) a backwards-incompatible change, so it can't be applied to the existing flags lightly, (b) it requires compiler-specific support that makes porting it harder, and (c) compiler-declared thread-locals doesn't work in pre-Vista windows if the thread-local is in a DLL. http://www.nynaeve.net/?p=187

So I still consider those things to do *at some future date* if there is demand for them. In the short term, I prefer to just make them plain globals for all the above-listed reasons.

nguillemot:
Seems fair. If the need ever arises to get rid of the globals, they could be packed them up into a context struct with its pointer passed around to alternate entry points. Either way, it's a separate problem from actually implementing the features.

PR #93, opened 2015-03-27
Single channel BMP loading

Mokosha:
Make sure to set number of channels after doing post-conversion of grayscale BMP images


Mokosha:
Sorry, this was an instance of not RTFM. According to the docs:

    // ... but 'n' will always be the number that it would have been if you said 0

I'll just assume that if `req_comp != n` and the return value is non-null that the desired data is of the `req_comp` format.

PR #94, opened 2015-03-29
added stbi_vertically_flip_on_load

nguillemot:
strawman number 2, continued from #89.


nothings:
I renamed the API to stbi_set_flip_vertically_on_load for consistency with "stbi_set_unpremultiply_on_load", so you'll have to change that when you integrate the new version.

PR #102, opened 2015-04-02
Make code compile as C++ with GCC.

Sofahamster:
I've added the missing ints required to make code compile as C++ with GCC/G++.
I've also changed a few char* to const char* to make GCC quit complaining about
"deprecated conversion from string constant to 'char*'"

### DIFF
--- a/stb_voxel_render.h
+++ b/stb_voxel_render.h
@@ -464,16 +464,16 @@ STBVXDEC void stbvox_reset_buffers(stbvox_mesh_maker *mm);
 // RENDERING
 //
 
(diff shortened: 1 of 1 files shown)


nothings:
I rewrote the const-char-* handling to work differently in a way that should work for people who prefer const-correctness and for people (like me) who prefer to never use const. But since I'm not testing in a const-correct environment I might have screwed it up.

nothings:
That is, any string you pass in to an stb library should generally be const (if you're non-const, you get an implicit conversion to const on the way in), but any string the stb library returns to you should generally be non-const (if you're const, it will implicitly cast to const on the way out).

PR #104, opened 2015-04-08
Adding memory and assert defines to stb_image_write

callidus:
Adding memory and assert defines to stb_image_write to match stb_image


nothings:
Note to self: actually it has even been version-bumped. However, still need to do following:
  - tweak credits
  - remove use of memcpy and only use memmove so macro-replacement API is slightly simpler

nothings:
I changed your credit to your full name; if you'd prefer "github:callidus" I'll change it back.

nothings:
Also I renamed all the macros to be STBIW_ rather than assume they should all use the same macros as stb_image.h

callidus:
that makes sense, thank you Sean. Credit how ever you think best, im not fussy. Thanks for such a useful library

PR #105, opened 2015-04-11
Fixing various crashes when loading jpg, png and tga images.

mmozeiko:
This fixes various crashes on malformed input files.

### DIFF
--- a/stb_image.h
+++ b/stb_image.h
@@ -1144,6 +1144,10 @@ stbi_inline static int stbi__at_eof(stbi__context *s)
 
 static void stbi__skip(stbi__context *s, int n)
 {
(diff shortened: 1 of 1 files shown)


nothings:
Note to self: "done except version bump" isn't quite true, as the place(s) where asserts were turned into dynamic checks, the asserts should be removed since the question "Can this even ever happen" implied by use of assert has been (presumably) answered in the positive.

Note to submitter: I will make this change myself, doing it yourself won't get it merged any sooner. :)

mmozeiko:
Ok, that is fine. I was not sure to remove asserts or not (maybe you want them for debugging purposes).

Issue #92, opened 2015-03-24
Incorrect I/O when reading grayscale jpeg image


Mokosha:
I ran the following program on a Windows 7 machine under Visual Studio 2013. To my knowledge, SSE2 is available.

    #include <tchar.h>
    #include <cassert>

    #define STB_IMAGE_IMPLEMENTATION
    #include "stb_image.h"

    #define STB_IMAGE_WRITE_IMPLEMENTATION
    #include "stb_image_write.h"

    int _tmain(int argc, _TCHAR* argv[])
    {
    	int width, height, nChannels;
    	unsigned char *pixelData = stbi_load("test.jpg", &width, &height, &nChannels, 0);
    	assert(nChannels == 1);

    	if (!stbi_write_png("test-out.png", width, height, nChannels, pixelData, width*nChannels)) {
    		return 1;
    	}

    	return 0;
    }

Here are the source and result images:

### `test.jpg`
![test](https://cloud.githubusercontent.com/assets/885364/6812412/f40edc1e-d22b-11e4-93c8-e06d9a6b6c47.jpg)

### `test-out.png`
![test-out](https://cloud.githubusercontent.com/assets/885364/6812421/03887db2-d22c-11e4-9c74-e8cd3fdd33d3.png)

Not sure why, but the rows of pixels at the bottom of the image are not right.

Mokosha:
FWIW, the same happens on a MacBook Pro running Yosemite 10.10 under clang 3.5

nothings:
Note to self: it appears that the file terminates in the middle of the last progressive refinement, and we read all 0s and this huffman decodes to '0x01' which causes us to set every pre-quantized AC to 1, and the blocks are what you get from dcting those after dequantization. (This seems a little implausible, but that's the best I can tell.)

jpgd stuffs the empty bit-buffer with an EOS marker, but appears to not be *looking*  for those in its huffman buffer filler so that doesn't make any sense. ijg libjpeg comments claim they just stuff with 0s at end of file. Need to find a jpeg lib that is public domain and that works for this file and log what huffman values it decodes.

rygorous:
I looked into this today. The IJG code seems to have an "optimization" to short-circuit when there are no more bits left to be read. The problem is that the behavior in that case is not actually the same as the spec-mandated behavior when reading an infinite string of zero bytes.

[...]

Issue #79, opened 2015-02-05
A progressive JPEG that gives 'bad huffman code'


eloj:
I have come across a progressive jpeg that loads in GD, OS X Preview and checks out okay with jpeginfo, but returns "bad huffman code" from stbi_load.

The image in question can be downloaded from https://unsplash.com/photos/8bMPkZgazyY/download, but you'll probably have to go there with a browser and save-as as using curl straight on the URL gives a lower resolution one.

```
$ jpeginfo -ic X_canal.jpg
X_canal.jpg 5184 x 3456 24bit Exif  Progressive Huffman 3967267  [OK]

$ sha1sum X_canal.jpg
95345ef7f8d26330196149e32ec09457f58c49a8  X_canal.jpg
$ md5sum X_canal.jpg
1d68c5995aeb1bf6c346039f9d6e4d8e  X_canal.jpg
```

It entirely possible that the image is indeed corrupt, but I thought I'd offer you the chance to look at it if you want. If it's an unsupported format I feel like the error message should be made clearer, if possible.

```
$ identify -verbose X_canal.jpg
Image: X_canal.jpg
  Format: JPEG (Joint Photographic Experts Group JFIF format)
  Class: DirectClass
  Geometry: 5184x3456+0+0
  Resolution: 72x72
  Print size: 72x48
  Units: Undefined
  Type: TrueColor
  Endianess: Undefined
  Colorspace: sRGB
  Depth: 8-bit
  Channel depth:
    red: 8-bit
    green: 8-bit
    blue: 8-bit
  Channel statistics:
    Red:
      min: 0 (0)
      max: 255 (1)
      mean: 129.937 (0.509558)
      standard deviation: 97.7956 (0.383512)
      kurtosis: -1.62445
      skewness: -0.101304
    Green:
      min: 0 (0)
      max: 255 (1)
      mean: 134.391 (0.527025)
      standard deviation: 74.1044 (0.290606)
      kurtosis: -1.3512
      skewness: -0.246102
    Blue:
      min: 0 (0)
      max: 255 (1)
      mean: 92.2491 (0.361761)
      standard deviation: 95.85 (0.375882)
      kurtosis: -1.1492
      skewness: 0.596392
  Image statistics:
    Overall:
      min: 0 (0)
      max: 255 (1)
      mean: 118.859 (0.466115)
      standard deviation: 89.8938 (0.352525)
      kurtosis: -1.36635
      skewness: 0.0719763
  Rendering intent: Perceptual
[...]

PR #108, opened 2015-04-19
stb_image: Progressive AC decoding - fix ZRL code.

rygorous:
The original AC decoding logic handled ZRL (runs of 16 zeros)
incorrectly.

The problem is that the original flow set r=16 and skipped the
final coeff write when s=0. This is not actually correct. The
problem is the intervening "refinement" bits.

With the original logic, even once we decrement r to 0, we keep
reading more refinement bits for subsequent coefficients until
we find the next currently-unsent AC in the current scan. That is,
it works as if it was trying to place 17 new AC values, and only
bails at the last minute from actually setting that 17th value.

This is wrong. Once we've found the 16th previously-unsent AC, we
need to stop reading refinement bits, otherwise we get out of sync
with the bit stream (which expects us to read a huffman code next).

The easiest fix is to just do what the JPEG standard implicitly
assumes anyway: treat ZRL as a run of 15 zeros followed by an
explicit magnitude-zero AC coeff. (That is, leave s=0 and actually
write s). So this is what this fix does.

### DIFF
--- a/stb_image.h
+++ b/stb_image.h
@@ -1852,8 +1852,8 @@ static int stbi__jpeg_decode_block_prog_ac(stbi__jpeg *j, short data[64], stbi__
                   if (r)
                      j->eob_run += stbi__jpeg_get_bits(j, r);
                   r = 64; // force end of block
-               } else
-                  r = 16; // r=15 is the code for 16 0s
+               }
+               // r=15 s=0 already does the right thing (write 16 0s)
             } else {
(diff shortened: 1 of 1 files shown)


rygorous:
This fixes issue #92.

rygorous:
It appears to fix issue #79 as well; all the 4 images mentioned in the Feb 24 comment decode correctly without errors now.

nothings:
Wow, thanks for tackling this!

PR #120, opened 2015-05-10
new library : stb_gl2d_render.h

jimon:
Hi, I was working on a new library for some time, and now I need some feedback on library itself and it's integration with stb suite.

Library is a very simple 2d renderer based on OpenGL 1.1, my motivation was to create the small and humble 2d lib because I'm tired of reinventing a wheel every time when I need to render some simple 2d graphics.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,6 +15,7 @@ library    | lastest version | category | LoC | description
 **stb_rect_pack.h** | 0.06 | graphics | 560 | simple 2D rectangle packer with decent quality
 **stretchy_buffer.h** | 1.02 | utility | 210 | typesafe dynamic array for C (i.e. approximation to vector<>), doesn't compile as C++
 **stb_textedit.h** | 1.6 | UI | 1290 | guts of a text editor for games etc implementing them from scratch
+**stb_gl2d_render.h** | 0.10 | 2D&nbsp;graphics | 238 | Simplest 2D renderer based on OpenGL 1.1
 **stb_voxel_render.h** | 0.80 | 3D&nbsp;graphics | 3644 | Minecraft-esque voxel rendering "engine" with many more features
 **stb_dxt.h** | 1.04 | 3D&nbsp;graphics | 624 | Fabian "ryg" Giesen's real-time DXT compressor
 **stb_perlin.h** | 0.2 | 3D&nbsp;graphics | 175 | revised Perlin noise (3D input, 1D output)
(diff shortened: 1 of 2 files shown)


nothings:
Tricky.

I have some specific coding things I can point out, but my initial reaction is I'm not sure this is really the right thing for an stb library. (Note though "initial reaction", meaning maybe I'll change my mind if I think about it some more.)

One of the things that is generally true of the stb libraries is they're designed to be easy to drop into existing apps. But something like this is more like a framework--it's the starting point of the app (you're not normally going to drop 2d rendering into a pre-existing app), so I'm not sure the whole stb-style approach is exactly right. Another thing is that I try to keep the stb libraries very portable; stb_voxel_render is the first one that's tied to OpenGL, but even there it does support multiple OpenGL variants (and potentially other APIs), since it explicitly avoids making any API calls for you.

The only example of something I have that explicitly wraps GL is a library for my personal use called "stb_gl.h" which wraps a bunch of gl functionality that I find annoying to do, and includes extension wrangling to do deal with being OpenGL 1.1. ( https://github.com/nothings/obbg/blob/master/src/stb_gl.h ) It's not doing anything like what you're doing, but it's relevant because it wraps GL, *and* I've never released it as an official stb library because of this portability issue. For example, most people on the Mac aren't using the "legacy" OpenGL that supports 1.1 functionality, they're using the main OpenGL, which is 3.x core with no back-compatibility, so they e.g. can't even do the glMatrixMode() stuff.

Also, this seems like not very much functionality. Some of the stuff like the glClear wrapper are entirely pointless unless if it eventually got ported to multiple 3D back-ends.

So basically, my initial thought here is that to be shippable as a public stb library in the official repository, it would need to already be working on multiple platforms (at least later versions of OpenGL -- in those you have to set up shaders and do your own matrices and all that annoying stuff, so a library that took care of that would be more handy). And maybe it would be nice to have some extra functionality to make it a little more worthwhile, like some vector graphics support (which is something people have been bugging me to make for a while, and this would be a good context to do it in).

jimon:
Thanks for feedback !

My idea is to make something like NanoVG (which is drop-in vector render lib) but much much simpler, for example if somebody developing 3d game and then want to render some sprites or text, they can use this lib and maintain focus on their 3d part, instead of dealing with 2d.

My plan now is to make it portable and OpenGL version independent and add some more features : some simple font rendering with bitmap provided in source code, also soon I will work on embedding native text rendering, it can be part of this lib.

What is your policy/vision about testing something like this ?

vinjn:
If someone wants to serious 3d development, he will use shader.

And NanoVG is good enough, you will hardly beat it.

jimon:
I'll close this one because I don't have much time to work on this nowadays :(

PR #121, opened 2015-05-12
stb_image_write: write TGAs with RLE compression

fahickman:
Not a lot of extra code for a reasonable size savings. (Cuts the size of my font atlases by about 50-75%.)

### DIFF
--- a/stb_image_write.h
+++ b/stb_image_write.h
@@ -149,6 +149,14 @@ static void writefv(FILE *f, const char *fmt, va_list v)
    }
 }
 
+static void writef(FILE *f, const char *fmt, ...)
+{
+   va_list v;
+   va_start(v, fmt);
+   writefv(f, fmt, v);
+   va_end(v);
+}
+
 static void write3(FILE *f, unsigned char a, unsigned char b, unsigned char c)
 {
    unsigned char arr[3];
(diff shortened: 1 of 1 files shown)


nothings:
Cool, I'll integrate next time I have an stb-lib-fixing weekend!

r-lyeh:
can we have an option to save raw/rle tga's ? i would like to support both when saving. thx :)

fahickman:
That wouldn't be hard to add, but I guess it's up to Sean since it means an API change. Seems to me, though, if you don't want RLE, you might as well just use BMP.

r-lyeh:
but what if I want to write both? maybe by adding an additional `stbi_write_tga_rle()` that does not break API at all

nothings:
I think I'd be ok with just breaking the API and adding a flag to the function call. Using a second API seems like it'd would actually be more annoying.

nothings:
Although maybe a better approach would be to have a flag you set. Not thread-friendly, but avoids both exploding the API and allows code selection between the two modes.

r-lyeh:
and what about an optional flag mask in `comp` to enable rle like, say, 0x8000 ?

nothings:
I'm opposed to embedding flags in 'comp'. https://github.com/nothings/stb/pull/89#issuecomment-86746959

fahickman:
I added a global flag for non-RLE saving along the lines of stbi__vertically_flip_on_load. I also fixed a regression in monochrome images. Apparently I was working with an older version of the file.

(I do not know how to use git properly, so I hope I didn't mess up the pull request.)

nothings:
i've merged it and compiled it and tweaked the name of the variable and made it exposed if not STB_IMAGE_WRITE_STATIC, but I didn't test it.

PR #130, opened 2015-06-12
Fix spelling.

rohit-n:


nothings:
The misspelling is actually intentional, as it causes the j to overwrite the previous l, hence the "with bug" above it.

nothings:
But I'll add a clearer comment in the forthcoming release.

nothings:
Updated.

PR #142, opened 2015-07-02
Tweaked to enable compilation under MinGW.

Mojofreem:
Made minor changes to stb.h to compile cleanly under MinGW. There were 4 issues dealt with:

1) MinGW is _WIN32, but uses GCC assembler semantics, not Microsoft style. Added an additional preprocessor check to only compile the assembly optimized version of stb_log2_floor on Windows platforms that are not MinGW.

2) Calls to _wstat were expecting a _stat struct, not a stat. This is due to Win32 using _stat, but stb was only checking for MSC_VER. Expanded the preprocessor check to include MinGW as well.

3) mkstemp() is not available in MinGW. Added a conditional workaround using mktemp() [less secure, but functional for now] wrapped in open(). This necessitated including fcntl.h for the O_RDWR parameter to open.

4) stb_dirtree_get_dir had an invalid double slash character. Removed the extraneous slash.

### DIFF
--- a/stb.h
+++ b/stb.h
@@ -190,6 +190,9 @@ Parenthesized items have since been removed.
 #include <stdio.h>      // need FILE
 #include <string.h>     // stb_define_hash needs memcpy/memset
 #include <time.h>       // stb_dirtree
+#ifdef __MINGW32__
+   #include <fcntl.h>   // O_RDWR
+#endif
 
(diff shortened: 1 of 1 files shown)


timsjostrand:
Can confirm this fixes MinGW build issues for us as well!

Bobo1239:
Helped me as well!

PR #146, opened 2015-07-07
Added support for 16-bit PSD loading.

rmitton:
This extends the current PSD loader to add support for 16-bit images, by
quantizing them down to 8-bit upon load.

### DIFF
--- a/stb_image.h
+++ b/stb_image.h
@@ -5195,7 +5195,7 @@ static stbi_uc *stbi__psd_load(stbi__context *s, int *x, int *y, int *comp, int 
             // Read the data.
             if (bitdepth == 16) {
                for (i = 0; i < pixelCount; i++, p += 4)
-                  *p = (stbi__get16le(s) + 127) >> 8;
+                  *p = stbi__get16le(s) * 255 / 65535;
             } else {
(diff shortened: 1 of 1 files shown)


nothings:
I'm merging this and it'll be in the next version, but:

1. I changed "pixel * 255 / 65535" to "pixel / 256". There are three options: (a) create 256 buckets each of which cover 256 of the 16-bit values; (b) create 256 buckets, with the 0 and 255 buckets half-sized (this is seen for float-to-int with rounding); (c) have buckets 0-254 cover 257 16-bit values, and bucket 255 cover 1 16-bit value (this is what your code does). I have chosen (a). Please let me know if I've analyzed this wrong.

2. It doesn't support RLE-compressed 16-bit data (not sure if it's allowed, the specification is unhelpful)

3. The full 16-bit values should be exposed through the floating-point-pixel "HDR" interface

If you have any interest in doing 2 or 3, let me know, otherwise I'll open up new issues for them.

nothings:
I've added a 16-bit API on a branch, but only hooked up PNG. If you want to hook up PSD to it now, it'll get in sooner; I'll get to it enventually otherwise.

https://github.com/nothings/stb/tree/png16

PR #154, opened 2015-07-29
Fix easy_font assuming math.h is already imported

jsheard:
It uses the ceil() function but doesn't include math.h itself.


nothings:
Given the nature of this fix and the fact there's no credits section, I'm not bothering to credit you for this fix. Let me know if this bugs you.

Issue #155, opened 2015-08-01
update public-domain dedication


nothings:
update all public-domain dedications to match stb_truetype

nothings:
fixed by above patch

PR #156, opened 2015-08-01
update public-domain dedication

rwhitworth:
This is the best I can do with issue #155.  I tried to use the truetype license text as the template, but keep the formatting of the individual file headers similar to what is already present.  Merge if this is what is desired or close this request if you plan to do the work yourself @nothings.


nothings:
Thanks, I have integrated these changes locally, but they will only show up on the github master branch as I make other changes (I don't want to bump all the version numbers just for this).

PR #157, opened 2015-08-04
GIF loading improvements.

urraka:
I adapted stb_image so it lets me load gif images and animations properly so I figured I could share these changes.

There are two main things I'm changing here:

1. GIF background is made transparent where expected.
2. `stbi__gif_load_next()` was adapted so it can be used to correctly load GIF animations, with (hopefully) proper handling of disposal methods.

There was also a possible memory leak in case of loading failure: `stbi__gif_load_next()` would allocate a buffer but return NULL in some cases, and buffer never gets freed or returned to the user.

I made some test cases comparing the currently official implementation vs this modified version so these changes are more clearly visible. Basically, I loaded a bunch of gif samples and saved them as PNG using stb_image_write. Here are the results:

https://dl.dropboxusercontent.com/u/8012562/stb/index.html

### DIFF
--- a/stb_image.h
+++ b/stb_image.h
@@ -5438,8 +5438,8 @@ typedef struct
 typedef struct
 {
    int w,h;
-   stbi_uc *out;                 // output buffer (always 4 components)
-   int flags, bgindex, ratio, transparent, eflags;
+   stbi_uc *out, *old_out;             // output buffer (always 4 components)
+   int flags, bgindex, ratio, transparent, eflags, delay;
    stbi_uc  pal[256][4];
    stbi_uc lpal[256][4];
    stbi__gif_lzw codes[4096];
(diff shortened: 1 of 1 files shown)


posva:
Thanks for this! I just updated to the last version in master and was annoyed with the transparency bug

posva:
@urraka I'm struggling trying to get data from GIF with stb_image.h. Could you send me the code you used for your test or any other useful resource, please ? My email address is posva13@gmail.com

urraka:
@posva Do you mean to get all the animation frames?

posva:
@urraka yes

urraka:
@posva I used this code here https://gist.github.com/urraka/685d9a6340b26b830d49

Not great quality but basically `stbi_xload` will return a buffer with all frames one after another, and also the delay information like [image#0][delay#0][image#1][delay#1][...]. Delay is 2 bytes (uint16) little endian if I recall correctly. Also, image loading is hard coded to 4 channels RGBA format. For non animations it will return a normal image buffer. It's also skipping an y-flip thing stb can do.

posva:
@urraka Thank you very much!

PR #163, opened 2015-08-22
Fix for 32bpp bmp images that load with alpha=0 on all pixels

urraka:
I found a case where the loaded image would be incorrectly fully transparent. This seems to fix that.

### DIFF
--- a/stb_image.h
+++ b/stb_image.h
@@ -4627,8 +4627,7 @@ static stbi_uc *stbi__bmp_load(stbi__context *s, int *x, int *y, int *comp, int 
                   mg = 0xffu <<  8;
                   mb = 0xffu <<  0;
                   ma = 0xffu << 24;
-                  fake_a = 1; // @TODO: check for cases like alpha value is all 0 and switch it to 255
-                  STBI_NOTUSED(fake_a);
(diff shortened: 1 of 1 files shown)


nothings:
Fixed in next version (2.08) -- I rewrote the fix to be a bit cleaner/faster, please check if 2.08 is still fixed for you.

nothings:
Er, version 2.07, not 2.08.

urraka:
Works perfectly.

On a side note, I get a compilation error on this line:
`#if defined (STBI_NO_BMP) && (STBI_NO_TGA) && (STBI_NO_GIF)`
https://github.com/nothings/stb/blob/023ff3ffd2d62ed3e4a654001c7a6b0d081dddf2/stb_image.h#L1303

I imagine you meant:
`#if defined(STBI_NO_BMP) && defined(STBI_NO_TGA) && defined(STBI_NO_GIF)`

(this is because I'm defining `STBI_ONLY_PNG` `STBI_ONLY_JPEG` `STBI_ONLY_BMP` `STBI_ONLY_GIF`).

nothings:
Fixed that for next version (that I'll release later today, probably).

PR #172, opened 2015-09-10
Fix VS2015 warnings in stb_image.h

Reedbeta:
This PR fixes some warnings that appear in stb_image.h in VS2015 on /W4.

* Variable-shadowing warnings (C4456)
* Converting 32-bit int directly to 64-bit pointer (C4312)

### DIFF
--- a/stb_image.h
+++ b/stb_image.h
@@ -902,8 +902,8 @@ static void *stbi__malloc(size_t size)
    #define stbi__err(x,y)  stbi__err(x)
 #endif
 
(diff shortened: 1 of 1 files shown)


nothings:
Deleting the 'int i' actually introduced a bug because the outer 'i' is 'unsigned int', and it's used for a down-counted loop. But don't worry, I fixed it.

nothings:
(Although I'm a little surprised VS2015 didn't warn about comparing an unsigned int >= 0)

Reedbeta:
Aha! Good catch. Yeah, I'm surprised by that too...

PR #173, opened 2015-09-12
DMC doesn't accept struct initializer syntax

triplefox:
It expects constants. The long way compiles (but I haven't actually called this function yet)


nothings:
What is DMC?

nothings:
Yeah, that change will break VC6, so I have to split it up more.

jarikomppa:
I think it's the digital mars c compiler.

triplefox:
Yes. D language's toolchain for C bindings prefers using the Digital Mars C
compiler, at least in the Windows environment. I have an  apparently
working but untested D  binding at this point :)

PR #174, opened 2015-09-13
Fixed bug in default alpha for non RLE-compressed PSDs

nickveri:
Setting alpha channel to 1 on non RLE-compressed RGB PSDs.


nothings:
Fixed for 2.08

PR #175, opened 2015-09-13
Fix new VS2015 warnings in stb_image 2.07

Reedbeta:
I pulled master and a couple more things showed up...sorry :)

* conversion from int to stbi_uc
* unused parameter


nothings:
No worries, I've got two other fixes as well, gonna release it again later today.

PR #182, opened 2015-09-25
nobug: removed unused variable 'k'

zerhacken:


mottosso:
Just encountered this, and can confirm that simply removing `k` in seems to work. Let's merge this?

PR #187, opened 2015-10-17
Fix race condition in static crc_table

karjonas:
If calling stbi_write_png concurrently the static array crc_table would be shared between threads causing data-races.


nothings:
Have you observed this in the wild?

karjonas:
Yes. I have a threaded application that takes continuous screenshots in several threads and I will get a warning when running this with clang thread sanitizer enabled.

karjonas:
"Your solution changes from computing the CRC table once to computing it every time." Isn't the crc unique to the file? In that case I cannot see how it is _not_ computed every time?

nothings:
So, one bug is that the code is checking crc_table[1] instead of crc_table[255]; the latter will indicate that the table is fully initialized. (The loop needs to be tweaked to compute the crc_table value into a temp, then write that temp into the crc_table[i] when it's fully computed.)

This will address all practical cases (except one where e.g. the compiler does the loop in a different order, which is highly unlikely), but will probably still not pass thread sanitizer. Since we don't have cross-platform thread primitives available, the simplest solution would be:

* Add an API funciton "stbi_write_init_multithreaded()" which just calls stbiw__crc32(NULL,0).
* Call that function first before creating threads that use stbi_write.

nothings:
The crc_table is always the same, the crc is unique to the file. It's only necessary to compute the crc_table once. Indeed, the better solution might be to precompute it offline and just fill out the array in the source file.

karjonas:
I have now replaced the calculation with a pre-filled table.

PR #191, opened 2015-10-20
Fix documentation error

karjonas:


nothings:
Thanks for all your fixes, sorry it took a while to integrate.

PR #200, opened 2015-11-09
Fix the calculation of the input image shift when using subpixel regions

BSVino:
Includes some more tests which had incorrect results before, but work fine with the included fixes. The test does a subpixel region that slides from [0, 0.5] to [0.5, 1] for both upsampling and downsampling. I think that should cover all of the uses of the subpixel stuff. I'm not bothering with s < 0 or t > 1 because I don't think we ever intended to support that in the first place. At the very least, non-subpixel stuff is unaffected because s0 = t0 = 0 means in_shift = 0.

### DIFF
--- a/docs/other_libs.md
+++ b/docs/other_libs.md
@@ -64,7 +64,7 @@ this! But you might like them anyway:
 - **profiling** [Remotery](https://github.com/Celtoys/Remotery): CPU/GPU profiler Win/Mac/Linux, using web browser for viewer (Apache 2.0 license)
 - **profiling** [MicroProfile](https://bitbucket.org/jonasmeyer/microprofile): CPU (and GPU?) profiler, 1-3 header files (unlicense) *uses miniz internally*
 - **parsing** [json.h](https://github.com/sheredom/json.h): JSON parser (public domain)
-- _**parsing** [jzon.h](https://github.com/Zguy/Jzon): JSON parser, C++ (MIT license)
+- _**parsing** [jzon.h](https://github.com/Zguy/Jzon): JSON parser, C++ (MIT license)_
 - **strings** [dfa](http://bjoern.hoehrmann.de/utf-8/decoder/dfa/): fast utf8 decoder (MIT license)
 - **data structures** [klib](http://attractivechaos.github.io/klib/): many 2-file libs: hash, sort, b-tree, etc (MIT license)
 - _**data structures** [uthash](https://github.com/troydhanson/uthash): several 1-header, 1-license-file libs: generic hash, list, etc (BSD license)_
--- a/stb_image_resize.h
+++ b/stb_image_resize.h
@@ -2229,8 +2229,8 @@ static void stbir__calculate_transform(stbir__info *info, float s0, float t0, fl
         info->horizontal_scale = ((float)info->output_w / info->input_w) / (s1 - s0);
         info->vertical_scale = ((float)info->output_h / info->input_h) / (t1 - t0);
 
-        info->horizontal_shift = s0 * info->input_w / (s1 - s0);
-        info->vertical_shift = t0 * info->input_h / (t1 - t0);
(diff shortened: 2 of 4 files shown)


nothings:
Doesn't this break the 'transform' case, because you're re-interpreting in_shift as out_shift?

BSVino:
Good point. I'm not sure if it breaks the transform case or if it fixes the transform case, since it should have been in_shift in the first place. But the only client of the transform case comes from stbir_resize_subpixel, which I now realize we never actually had any test cases for, and the variable names aren't unambiguous enough to tell if they were intended to be input or output shifts. The closest line of code that indicates that is:

    info->horizontal_shift = s0 * info->input_w / (s1 - s0);

which is dubious since it's the offending line, and it was obviously never tested since here we are now.

Thinking like a person using the API, if I see the names

    float x_scale, float y_scale, float x_offset, float y_offset

then my inclination is to think that they are the x and y offsets of the input image, rather than of the output. It's more immediate to think of x_offset and y_offset as specifying the "top-left" of the input area (ie being in [0, in_w] and [0, in_h]) that I want to shrink/blow up than to think of them as transformations of the output image. But that would imply that stbir_resize_subpixel never worked at all.

I guess it's possible that someone has written code against stbir_resize_subpixel, in which case this patch will break them. If so, they must have accepted that the offsets are in output pixels without reporting that as a bug. Doesn't seem very usable to me though.

Your thoughts?

nothings:
A linear transform is 99% of the time written as
   a*x + b
so that was my assumption of what it should be. Moreover, if we expose the transform to the end user in the most generic API (I think we do), then we should preserve the old definition of it if possible.

BSVino:
Alright then, I'll do that. Give me a day or two, Mondays are crazy for me.

BSVino:
Sorry about the mess of commits, I don't know how to use git apparently. One of them is just undoing stuff from the old commit. Really the total diff in the actual code is only two lines.

There are a bunch of new test cases now making sure that upsample & downsample for stbir_resize_subpixel stayed compatible with how it was before (which makes more sense to me now that I've used it a bit) and some new ones for stbir_resize_region up/down/no scaling.

I didn't bother running the full test suite because, like before, the changes only have effect with a nonzero shift. The new tests pass just fine.

BSVino:
Actually, I thought better of it and ran the other tests and they're failing, hold off on this until I figure that out.

BSVino:
Yea, it was because I screwed up the damn merge. OK, all test cases pass now, and I even added a few more so that things that were checked with _region are checked with _subpixel. Sorry about that.

eloj:
Just git rebase and 'fixup' the old commits if you want to clean up the history before merge.

nothings:
Yes, I will do this when I merge.

BSVino:
Sorry, I guess I've made more work for you. I'll try to see if I can do it myself before I leave for the weekend.

nothings:
Don't worry about it, I want to write some more tests first.

BSVino:
OK. Note I had to dig up the test images from my email, if you can't find them I can re-send them.

BSVino:
Didn't realize I still had this PR open, but I suppose you can have this kerning stuff if you want. Did you ever get a chance to look at stbir?

nothings:
I've only had a couple sessions of merging stb fixes since you submitted, and those have prioritized the other libs, but obviously I still want the original fix here.

I actually said no just the other week to a similar kerning function: https://github.com/nothings/stb/pull/236 but I guess if multiple people want it I should reconsider.

BSVino:
My argument about N^2 yadeya was the same reason that I did it. Design wise I like my approach over his since in his patch the loop is in client code and in mine I have stdtt do the loop for you. Maybe you have some better ideas though.

nothings:
I don't even remember what those tests I wanted to write were, so I"m merging this as-is. I'm going to rewrite the kerning stuff to a middle ground between this approach and the other approach.

PR #204, opened 2015-11-14
Fix for stbi_info on TGA files

baldurk:
If you call any of the `stbi_info*` family of functions right now on a TGA file it can easily break. `stbi__hdr_info` tries to read a token to verify the header, which reads to the next newline or EOF. Even if it then rewinds on failure, the original file header may well be long gone from `s->img_buffer`.

Since there's a nice test function that will only look at the first few bytes it's interested in, using that should be safer.


DanielGibson:
I just noticed the same problem and your patch seems to work :-)

PR #208, opened 2015-12-06
stb_image.h: Improve TGA support, incl. 16bit support

DanielGibson:
16bit TGAs were always loaded as greyscale+alpha, even if the TGA image type was not 3 or 11 (the types for greyscale images). However, when not grey, it's supposed to be RGB555 (potentially with an
1-bit alpha-channel in the MSB).
Those kind of images are supported now, also when using a colormap (palette) with 16bit colors in them. It's converted to `STBI_rgb` or `STBI_rgb_alpha`.

`stbi__tga_info()` didn't correctly calculate the pixelformat (`comp`) for colormapped images - it used the TGA's "normal" bits per pixel field instead of the one for the colormap.
For some reason `stbi__tga_test()` used `stbi__get16be()`, even though TGA is little endian.
Furthermore I improved both functions to do some more sanity checks and to consistently rewind the context.

I tested the code with TGA test images from https://github.com/ftrvxmtrx/tga/tree/master/testdata (which seem to mostly be original testdata from Targa) and https://github.com/mike-lischke/GraphicEx/tree/master/TestSuite%20original/TGA
It seemed correct in all  tests; the only thing I'm not sure about is `ctc32.tga` from the Targa testdata, which is completely translucent (but so it is in imagemagick's display and in the Gimp).
This is because it's a 32bit TGA (that even reports an 8bit alpha chan) and all the alpha values are set to 0 - so maybe it's intended to be invisible or it's for some TGA interpreter (or old Targa graphics chip) that interprets the "Attribute channel bits" not as alpha but as "overlay bits", whatever that means. Whatever, other 32bit TGAs work as expected.

### DIFF
--- a/stb_image.h
+++ b/stb_image.h
@@ -4811,6 +4811,26 @@ static stbi_uc *stbi__bmp_load(stbi__context *s, int *x, int *y, int *comp, int 
 // Targa Truevision - TGA
 // by Jonathan Dummer
 #ifndef STBI_NO_TGA
+// returns STBI_rgb or whatever, 0 on error
+static int stbi__tga_get_comp(int bits_per_pixel, int alpha_bits, int is_grey, int* is_rgb16)
+{
+   // only RGB or RGBA (incl. 16bit) or grey allowed
+   if(is_rgb16) *is_rgb16 = 0;
+   switch(bits_per_pixel) {
+      case 8:  return STBI_grey;
+      case 15: if(is_rgb16) *is_rgb16 = 1; return STBI_rgb;
+      case 16:
+         if(is_grey) return STBI_grey_alpha;
+         else {
+            if(is_rgb16) *is_rgb16 = 1;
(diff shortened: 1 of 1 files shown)


DanielGibson:
Ook, forget the "alpha channel in 16bit RGB TGAs" thing, it doesn't seem to work if done "correctly"[*], where a bit set means opaque and no bit set means translucent - doing that makes most 16bit TGAs I could find (all I assumed having an alpha-channel due to having an alpha-bit set) completely translucent (invisible), I very much doubt that's intended.
So now all non-greyscale 16bit TGAs are converted to RGB with no alpha.

[*] "correctly" according to http://www.paulbourke.net/dataformats/tga/ http://www.gamers.org/dEngine/quake3/TGA.txt or http://www.ryanjuckett.com/programming/parsing-colors-in-a-tga-file/

PR #210, opened 2015-12-12
Initial 16-bit PNG support

socks-the-fox:
I went with more or less the most simplistic route possible to add this, taking advantage of the fact that the PNG data filters are applied per-byte instead of per-pixel-channel and just decoding every other byte for 16-bit PNGs.

Current known issue: It doesn't handle the 16-bit version of the `tRNS` chunk at all. Adding support for that would require a lot of rewriting due to the way the code currently cheats (completely skipping the low-order bytes that would be needed for a proper comparison; since the `tRNS` chunk is handled *after* the data is discarded doing the same may result in false-positives, for example: `Img: 0xFFF0 tRNS: 0xFFFF` -> `Img: 0xFF tRNS: 0xFF`).

### DIFF
--- a/stb_image.h
+++ b/stb_image.h
@@ -4003,6 +4003,7 @@ static int stbi__create_png_image_raw(stbi__png *a, stbi_uc *raw, stbi__uint32 r
       stbi_uc *prior = cur - stride;
       int filter = *raw++;
       int filter_bytes = img_n;
+      int raw_step = (depth == 16? 2 : 1);
       int width = x;
       if (filter > 4)
          return stbi__err("invalid filter","Corrupt PNG");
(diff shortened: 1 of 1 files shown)


nothings:
I don't think the cheaty approach makes sense, since the float path needs the full 16-bit value anyway. So I think we want to decode the full 16-bit image, then convert to 8-bit (allowing 16-bit trans) or convert to float. This will be inefficient in the 16-bit-file, load as 8-bit path, but people who want speed don't need to use that path.

Also, adding a "*raw_step" to the inner loop is probably a bad idea. (It would be fine for the 16-bit path, but you're also paying that price on the 8-bit-image path.) Whereas a fully-decoded 16-bit image has to just decode all the bytes so won't need a 'raw_step' kind of thing (although if it turns out to be a pain, it's fine if we just have to replicate the filter inner loops for 8 and 16 bit).

nothings:
It may also just be cleaner to basically have a totally separate 16-bit path for the pixel decode path if the 16-bit support is going to slow down the 8-bit path. If that's ~50 lines I'm fine with it.

socks-the-fox:
After reviewing the code some more, I think I can add 'proper' 16-bit PNG support. I'll issue a new pull request when that one is ready.

PR #211, opened 2015-12-13
Avoiding conditional directives that break statements

RomeroMalaquias:


nothings:
why?

RomeroMalaquias:
Hi,
this change is only aesthetical, it's a suggestion based on code style guidelines from the Linux Kernel. 
[Linux kernel coding style](https://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/tree/Documentation/CodingStyle#n892)

nothings:
This file isn't part of the linux kernel. I'm not going to randomly apply some rules from random style guides to random stb libraries.

Also, the linked style guide doesn't actually recommend this change.

RomeroMalaquias:
Ok, thanks for your time

PR #212, opened 2015-12-14
Proper png 16bpc

socks-the-fox:
A proper implementation of the 16-bit-per-channel PNG format, with `tRNS` support. Float interface takes advantage of greater depth, byte interface just takes the most significant byte and uses that for speed.

### DIFF
--- a/stb_image.h
+++ b/stb_image.h
@@ -21,7 +21,7 @@
           avoid problematic images and only need the trivial interface
 
       JPEG baseline & progressive (12 bpc/arithmetic not supported, same as stock IJG lib)
-      PNG 1/2/4/8-bit-per-channel (16 bpc not supported)
+      PNG 1/2/4/8/16-bit-per-channel
 
       TGA (not sure what subset, if a subset)
       BMP non-1bpp, non-RLE
@@ -146,6 +146,7 @@
 
 
    Latest revision history:
+      2.xx  (2015-12-14) add 16-bit PNG support
       2.08  (2015-09-13) fix to 2.07 cleanup, reading RGB PSD as RGBA
       2.07  (2015-09-13) partial animated GIF support
                          limited 16-bit PSD support
@@ -195,7 +196,7 @@
     Omar Cornut (1/2/4-bit PNG)                  Ken Hamada
     Nicolas Guillemot (vertical flip)            Cort Stratton
     Richard Mitton (16-bit PSD)                  Blazej Dariusz Roszkowski
-                                                 Thibault Reuille
+    socks-the-fox@github (16-bit PNG)            Thibault Reuille
                                                  Paul Du Bois
                                                  Guillaume George
                                                  Jerry Jansson
@@ -228,7 +229,6 @@ distribute, and modify this file as you see fit.
 // DOCUMENTATION
 //
 // Limitations:
-//    - no 16-bit-per-channel PNG
 //    - no 12-bit-per-channel JPEG
 //    - no JPEGs with arithmetic coding
 //    - no 1-bit BMP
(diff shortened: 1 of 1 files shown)


nothings:
Sorry for the delay reviewing this, as I had xmas break and then I was sick.

It mostly looks reasonable, but I have the following issues:

1. "ldr16_to_hdr" is maybe (or maybe not) a mistake. A 16-bit channel isn't really "low dynamic range".

 Indeed, for all I know 16-bit PNGs may not be gamma-corrected; need to investigate what is common for this. Currently stb_image ignores the gamma stored in the PNG because I doubt for 8-bit it's trustworthy, but maybe at least for 16-bit PNG should respect it? Or maybe there should be a separate set of gamma scaling values used for 16-bit-channel files, or a flag for whether to respect it or not or to assume they're gamma corrected or not. What a mess.

2. I'm not sure if stbi__convert_format_float is a good idea. Need to keep in mind that we may eventually supply a non-float interface to 16-bit-per-channel files, so this conversion step should probably happen in 16-bit, and *then* do conversion to hdr/float? (Also, regardless, the conversion from RGB to Y probably needs to be different in gamma- and non-gamma spaces. I don't know the details.)

3. For performance reasons, the loop which sets the second byte of alpha to 255 after the main decode loop should be moved inside the for(j) loop (and only set one scanline of data each time); this will let it update the alpha channel while it's still in cache.

4. "if (stbi__parse_png_file(p, STBI__SCAN_load, req_comp) && stbi__reduce_png(p))"

 I don't do this in stb style (multiple side-effects in an if; generally in stb if it has a side-effect, it has to be bare without any other booleans); change it to nested ifs.

5. Unless I'm missing something, you're not handling endianness conversion in the float path. Rather than have an #ifdef, probably the best way to handle this is to have the first function called after the raw decode (currently that's stbi__ldr16_to_hdr, but as noted above there should probably be a convert_format16 first instead) take uint8 and output uint16 by manually combining bytes, rather than casting the input to uint16; I believe the stbi__reduce_png path already handles this correctly.

nothings:
Also don't bother trying to update version number, contributor list, etc., as it turns out that just makes it harder for me when I'm merging multiple fixes.

socks-the-fox:
As far as gamma goes, I'd say consistency between 8-bit and 16-bit paths should be most important. The specs make no distinction between 8bpc and 16bpc regarding gAMA, cHRM, sRGB and iCCP, so IMO we shouldn't either. If you decide applying gamma and other color space corrections is important, I'd recommend making them their own separate functions so other formats that have that info can take advantage as well.

I will admit most of what I did for the float path was kind of tacked on as an afterthought, as I only needed 8-bit color data. I'll focus on improving that.

nothings:
Consistency is not important; what works for users is what's important.

In the existing 8-bit code, in an imaginary ideal world, we would use the gAMA chunk. In practice, I doubt it's trustworthy, and we just rely on a user-set gamma.

The fact that the *spec* doesn't distinguish between gamma for 8-bit and 16-bit is because the *spec* assumes you *are* applying gAMA. For example, suppose hypothetically we live in a world where all 8-bit images are sRGB, and all 16-bit images are linear. Then in this hypothetical world, the reason the specification makes no distinction is because the specification allows for the 8-bit images get a gAMA chunk (or rather an sRGB chunk), and 16bpc has whatever chunk is necessary to indicate it's linear.

But we live in a world where we're *ignoring* the gAMA/sRGB chunks entirely. In the existing code we rely on users to know what gamma their images are (and for e.g. games they do, they're going to be authored with a consistent gamma). We can continue in the 16-bit case to rely on users knowing what gamma their images are, but we should handle the case where their 16-bit images are linear and their 8-bit images are sRGB, and your proposal ignores that.

On the other hand we could drop the float support entirely; I took a look and the existing PSD code does support 16-bpc but only by loading as 8-bit, and it likewise ignores the gamma issue and just discards the low byte. So PNG that did the same would be consistent with that, at least.

socks-the-fox:
I would say, drop 16bpc PNG float support initially, and in future releases work towards adding functions that allow *all* image formats that support these features to take advantage of higher bit depths, gamma, color space, etc.

I'd have to spend a lot more time getting familiar with the other format's parsing code, but I'd be willing to get started on that once we've decided on this.

nothings:
That's fine with me.

socks-the-fox:
Alright. I'm going to go ahead and close this pull request and start from a clean slate, as I kind of littered little things all over to implement the 16-bit -> float part. It'll also give me a chance to clean a few things up a bit more and work with the latest version.

PR #216, opened 2016-01-02
Added LUrlParser

corporateshark:
Lightweight URL & URI parser (RFC 1738, RFC 3986)


nothings:
This is marked as MIT license in your submission, but it looks like a modified BSD license?

Also, the modified clause ("Redistributions in binary form must display the name 'Sergey Kosarevsky' __in the credits of the application__, if such credits exist. The __authors of this work must be notified via email__ ([email address omitted] in this case of redistribution.)" seems pretty onerous, to the point that I'm not really comfortable recommending it to people. (It may not even be GPL-compatible, I don't actually know.)

corporateshark:
Indeed. License issues fixed, canonical MIT license is applied from now on.

nothings:
The C and H file still have the old license. :)

corporateshark:
Fixed.

nothings:
Thanks for the changes and for submitting here!

PR #219, opened 2016-01-07
fix compile warning

qnnnnez:


nothings:
Fixed this a different way to avoid dealing with platform-dependent warnings of unused expressions etc.

PR #220, opened 2016-01-11
ttf: Correct direct use of fabs()

sglass68:
Add a macro to allow fabs() to be provided by the user, as with the other
maths functions.

Signed-off-by: Simon Glass <sjg@chromium.org>


nothings:
Uh, this diff defines STBTT_fabs() but never calls it?

sglass68:
Oops. I've sent a new one.

PR #221, opened 2016-01-14
Added support for comments in PNM files (as generated by Gimp at least)

romigrou:
As the commit implies, here's support for loading PNM files with comments. It's been tested against Gimp-generated files plus a few other hand-generated ones.


nothings:
Sorry, someone submitted a fix for this that I've had in since November, I just never released it.

romigrou:
My bad, I've had this sitting on my drive since October :)

nothings:
I'll close this but let me know if it's doesn't handle your cases.

PR #222, opened 2016-01-14
Added the old size as argument to STBI_REALLOC and STBIW_REALLOC

romigrou:
I often find myself in situations where allocators don't have a realloc feature and querying the size of an allocated block is not possible either. As a workaround I added the previously allocated size as an argument to the STBxx_REALLOC() macros.

Unfortunately, this change breaks backward compatibility for those who redefined the macros. But then, the fix is trivial.


nothings:
I will probably change this so it doesn't break backwards compatibility by having their be two different macro names and you can define either one.

romigrou:
I should have thought about that solution. Thanks for the merge.

PR #225, opened 2016-01-18
Initial 16 bit per channel PNG support

socks-the-fox:
Initial 16 bit per channel PNG support. No float interface yet. This one cleans up a few little things I noticed and fixes a couple of mistakes I missed in the last pull request. Third time's a charm?

### DIFF
--- a/stb_image.h
+++ b/stb_image.h
@@ -3938,6 +3938,7 @@ typedef struct
 {
    stbi__context *s;
    stbi_uc *idata, *expanded, *out;
+   int depth;
 } stbi__png;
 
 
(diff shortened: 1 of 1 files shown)


nothings:
Transparency still isn't endian independent. The tc16 values are explicitly decoded from big-endian to native-endian, but the raw data is just casted from uint8 to uint16. Not sure what the 'both bytes are the same' bit means.

socks-the-fox:
Right, I keep forgetting the buffer isn't platform native at that point.

Here's what I'm thinking: Since we're going to need to convert the buffer to platform native once the 16bpc interface is implemented anyway, I'll just implement the conversion now, so that future functions that need to work with the full 16bpc data get it in the correct format right away. Then I won't need to change anything in the transparency function.

The "both bytes are the same" refers to the fact that the alpha value for all pixels at that point is `0xFFFF` which is represented the same big-endian or little-endian. Once I force the data into the platform native format it won't matter and I can probably just remove the comment.

nothings:
Sure, sounds reasonable.

socks-the-fox:
Fixed up the endianness issues and squished it. Not happy with having to do the conversion outside the main loop but my rudimentary attempts to do it within weren't working.

nothings:
Yeah, it's not optimal, but eh, reading 16 bit images as 8-bit not being optimal ain't the end of the world.

socks-the-fox:
Yup. Better than bringing in the *entire* libPNG codebase just to read 16-bit PNGs like I was doing in my project though, and as things get improved it'll eventually get fixed.

socks-the-fox:
Hey, just wanted to follow up and see if there is anything more you want me to do with this!

nothings:
Nope, it's just waiting for the next time i update stb_image.

PR #226, opened 2016-01-21
Silence unused variable warning in STBI_REALLOC_SIZED macro

elvman:
A dirty hack to silence the "unused variable" warning. Compiler will optimise this and remove the (0*oldsz) part with any optimisation enabled.

### DIFF
--- a/stb_image.h
+++ b/stb_image.h
@@ -637,7 +637,7 @@ typedef unsigned char validate_uint32[sizeof(stbi__uint32)==4 ? 1 : -1];
 #endif
 
 #ifndef STBI_REALLOC_SIZED
-#define STBI_REALLOC_SIZED(p,oldsz,newsz) STBI_REALLOC(p,newsz)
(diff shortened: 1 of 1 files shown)


eugeneopalev:
Also you can define this macro like this:
```c
#define STBI_REALLOC_SIZED(p,oldsz,newsz) STBI_REALLOC(p,newsz);oldsz
```
This simple reference to oldsz will be optimized even in debug build without any optimization enabled.

elvman:
That is a really bad idea. Your code will not compile if used in "if" statement. Also happy debugging the this: bool isThree = STBI_REALLOC_SIZED(p, 2, 3) == 3;

nothings:
Can you just list what the actual warnings are, and I can mark up the variables as unused, instead of trying to hack it in the macro?

eugeneopalev:
Oh, that is really bad idea. My bad.
BTW: why does your STBI_REALLOC_SIZED return the block size?

nothings:
I've released an attempted fix for this (using NOTUSED to suppress what I assume the warnings are). Let me know if it didn't work.

PR #230, opened 2016-01-23
Add atomic-ring to other_libs

djcapelis:
I saw your awesome list of single file libraries and thought I'd propose a project I wrote as part of it. :)  It's a lock-free queue I use to pass tasks between threads in most of my multithreaded code.  No locks.  It does unfortunately require C11, but there's no dependencies beyond that, not even POSIX.  Should work on every architecture you can find a C11 compiler for.  (The tests uses POSIX, but the library is all pure standardized cross platform C.)

Cheers!  And thanks for putting this list together, it's quirky and I like it.

### DIFF
--- a/docs/other_libs.md
+++ b/docs/other_libs.md
@@ -54,6 +54,7 @@ images | [jpeg-compressor](https://github.com/richgel999/jpeg-compressor) | **pu
 math | [mm_vec.h](https://github.com/vurtun/mmx) | BSD | **1** | SIMD vector math
 math | [ShaderFastLibs](https://github.com/michaldrobot/ShaderFastLibs) | MIT | **1** | approximate transcendental functions optimized for shaders (esp. GCN)
 multithreading | [mm_sched.h](https://github.com/vurtun/mmx) | zlib | **1** | cross-platform multithreaded task scheduler
+multithreading | [atomic-ring](https://github.com/djcapelis/atomic-ring) | **public&nbsp;domain** | 2 | Single Producer Single Consumer (SPSC) lock-free queue
 network | **[zed_net](https://github.com/ZedZull/zed_net)**|**public&nbsp;domain**|**1**|cross-platform socket wrapper
 network | [mm_web.h](https://github.com/vurtun/mmx) | BSD | **1** | lightweight webserver, fork of webby
 network | [par_easycurl.h](https://github.com/prideout/par) | MIT | **1** | curl wrapper


nothings:
Can't you implement an SPSC queue without atomics? The writer writes head and the reader writes tail (or vice versa, depending on your preferred queue terminology), while each only reads the other one.

I guess you need a barrier between writing an element and updating the head, though.

djcapelis:
The head and tail pointers are owned by the individual threads, atomics are needed for the one piece of shared state: an atomic int that tells you how many items are in the queue.  I don't think you can get away without that being atomic, but if you figure out a way, be sure to publish the paper, because a whole field of computer science would like to hear about it. :)

nothings:
What does "atomic" mean in the C11 context? I'm used to it being used to refer to the ordinary special 'atomic operations' like compare-and-swap or atomic addition operations, which actually involve extra , more sophisticated instructions at the processor level; these are the tools people usually use for creating lock-free data structures, but they're unnecessary for SPSC circular buffer.

If it just means a variable that gets read/written atomically, well ok; I wasn't aware that this needed explicitly calling out (for naturally aligned variables), but ok. (But people have been deploying SPSC circular buffers for decades w/o such things, and that's where my knowledge comes from.)

nothings:
Most google results seem to suggest that this is forgotten knowledge, but here's an example, from the Linux kernel:

````
  By using memory barriers in conjunction with circular buffers, you can avoid
  the need to:

   (1) use a single lock to govern access to both ends of the buffer, thus
       allowing the buffer to be filled and emptied at the same time; and

   (2) use atomic counter operations.

  There are two sides to this: the producer that fills the buffer, and the
  consumer that empties it.  Only one thing should be filling a buffer at any one
  time, and only one thing should be emptying a buffer at any one time, but the
  two sides can operate simultaneously.
````

https://www.kernel.org/doc/Documentation/circular-buffers.txt

To answer your question explicitly, the item count is 'head-tail' (or vice versa, depending on convention); such a count can be stale in exactly the same ways item_count can be stale, which never matters for correctness.

djcapelis:
The kernel code you reference uses smp_acquire and smp_release atomic functions within the kernel.  There are actually more of these operations in that code than there are in atomic_ring, as it uses atomicity on both the head and tail pointers to gain thread-safety.

nothings:
"torn?" I assume your "++item_count" is an atomic increment, at least.

The code from that link only uses those operations to turn the SPSC queue into an MPMC queue. There is no synchronization at all between the producer and the consumer except memory barriers. As the big block of text says.

djcapelis:
I see, is it atomic_fetch_add_explicit(&aring->items, 1, memory_order_release); that's freaking you out?  On x86 it's the same as a memory barrier, I'm mostly using atomics for the memory barriers.  Paul's code is clever (as his code tends to be) and atomic-ring could possibly be a bit faster on some architectures if I danced pointers instead of used an item count, but uhm, are you really quibbling about an operation like this that happens on a local core's cache?  The main performance overhead comes from the operation that's done on cache lines in the first shared level cache, which has to be done either way whether you use memory barriers or not.  And frankly I'm not sure Paul's code is necessarily faster if the pointers aren't on the same cache line.  I'll have to benchmark the difference sometime.  Maybe it'll get below the 140 cycles it takes now.

(The other alternative for perf improvement here is batching, which is what the workqueue for Apple's GCD does.)

But I mean, whatever makes you happy.  If this change doesn't, don't accept the pull. :)

nothings:
It's not about performance (I haven't even looked at the code), it's that using C11 atomics means it's only usable from C11 and later. That seems weird when we've done without for decades, but of course we've done so non-portably, I'm sure. But it effectively makes the code harder for other people to drop into their projects (if they're not already using C11; I don't have stats). I'm actually more interested in *practically* portable libraries, like ones that have #ifdefs for the five most common platforms and really make it work in practice. (But this may also be my bias from being on Windows where even if you *do* use a modern compiler you can't even rely on C99.)

If I cared about performance I'd explore the consequences of storing head/tail/item_count in the same cache line. It's not really "false sharing" exactly since they tend to get updated at the same time, but I suspect head & tail in separate cache lines (and no item_count) would have better ability to get updated by both threads at once, since they both don't need to take ownership of the same cache line. (As long as you have item_count, they do both need to own that cache line.) But that's probably not really that important at the level of threading granularity these things tend to be used at, and regardless, it's a level of detail I don't generally scrutinize libs on this list to!

nothings:
Also I just reformatted the list source massively, but don't worry about updating the PR to match it, if we do resolve this I'll update it myself.

djcapelis:
I don't care what you do, but I have to get work done now, so this has to be my last comment on this.  Sorry. :)

I agree the reliance on C11 is a total bummer, but I didn't want to rely on inline assembly that was specific to a particular set of processors or platform specific functions, and those were pretty much the remaining options.

C11 atomics just provide the first standardized way of doing the same tricks in C everyone has been doing for all those years.  You literally couldn't do it without inline asm (either in your own code, or in a function you called to do the dirty work) before C11.

Wild, right?

nothings:
"You literally couldn't do it without inline asm": I don't know about gcc, but this is *literally* not true. All that is required is memory barriers; on x86, given its memory model, I think you only need compiler memory barriers (which on non-gcc platforms were [instrinsics not asm](https://en.wikipedia.org/wiki/Memory_ordering#Compile-time_memory_barrier_implementation)), and regardless of x86 memory model, you go from [instrinsics](https://en.wikipedia.org/wiki/Memory_ordering#Compiler_support_for_hardware_memory_barriers) to the extreme of e.g. MSVC 2005, where I think all that was required was 'volatile'. ("With Visual Studio 2003, volatile to volatile references are ordered; the compiler will not re-order volatile variable access. With Visual Studio 2005, the compiler also uses acquire semantics for read operations on volatile variables and release semantics for write operations on volatile variables (when supported by the CPU)." )

Anyway, I understand that this discussion is a waste of time from your point of view. But since you're wrong about the historical angle and the minimum requirements for SPSC circular-buffer it's hard for me to judge the validity of your other claims.

djcapelis:
Intrinsics explicitly are not part of C and explicitly aren't portable.  The defined behavior of volatile isn't what MSVC says it is.  The minimum requirements for an SPSC workqueue are memory barriers, which is *exactly* what I use C11 atomics to get.

Closing.  And please the next time you talk with someone about C11 atomics, actually bother to know what they are first.

PR #232, opened 2016-01-23
Added minctest and tinyexpr.

codeplea:
Hi. I love your code!

I was wondering if you might consider adding my small C libraries to your `other_libs` list: [TinyExpr](https://github.com/codeplea/tinyexpr) and [minctest](https://github.com/codeplea/minctest).


nothings:
You might consider adding ``#ifdef __cplusplus / extern "C"`` to make your libs usable from C++ as well.

codeplea:
Thanks for the merge!

I've been purposely avoiding C++ for a few years now, but I'll add the `#ifdef __cuplusplus` guards as you suggested.

By the way, you're one of my role models. I remember discovering `stb_truetype.h` as being like finding an oasis in the desert.

nothings:
Thanks!

PR #236, opened 2016-01-26
Added functions for iterating over the kerning values

JCash:


JCash:
Since I'm (currently) using this in an offline pipeline tool, it's beneficial to retrieve these values for later usage in game.

I tried to keep the style of the file as much as possible, but ofc it might differ a bit.

nothings:
Wouldn't being able to iterate over the codepoints defined by the file be the first order of business? Or is there already a way to that I'm forgetting?

JCash:
Personally, I'm used to having the set of wanted code points as input to the tool, and I'm going under the assumption that the font I use contain the set (or warn if it doesn't (stbtt_FindGlyphIndex returns 0)). So in the past I haven't needed to iterate over the code points in the font.

But I think it might be a good idea, and I'll be happy to add that to this pull request if you'd like?

nothings:
No, you don't have to, I'm just trying to understand. If you know you're using a certain N code points, you *could* just query every N*N pair of those with the current API, right? Since this is an offline tool, is that actually significantly slower?

JCash:
Yes, that would be a way to do it. And as you suggest, it would take no time in practice. However, I think it's more about the "neatness" of it all. The number of actual pair kernings is quite small compared to N*N, so for me it just seems "wasteful" to do a such a search when there is a shorter path to take. I prefer an access methodology over a query methodology. (My number of pair kernings is ~1-2% of the possible pair combinations, english font)

That said, I have to add that although my tool is currently offline, I _am_ moving towards a realtime solution where the stbtt_GetGlyphKernAdvance() function would work well, and would maybe make these new functions obsolete (for me). I merely found these helper functions useful, and thought that others might find them useful too.

In conclusion, if no one else has mentioned this before, it might not be a good fit for the api. :)

nothings:
Yeah, there's absolutely nothing wrong with adding them in the abstract, I'm just a little careful since every function added to the API makes it a little harder to understand.

PR #242, opened 2016-02-10
Slightly modify the public domain license

oon3m0oo:
Slightly modify the public domain license to keep it in the public domain, but make it clear that even when dedications might not be recognized that the code is still usable.  Given that this isn't dual-licensing under a different license, I'm hoping this will be acceptable, and considered within the spirit of stb.  This is a wonderful library, and this change makes it more generally usable.

### DIFF
--- a/docs/stb_howto.txt
+++ b/docs/stb_howto.txt
@@ -165,9 +165,9 @@ public domain declarations aren't necessary recognized
 in the USA and some other locations. For that reason,
 I recommend a declaration along these lines:
 
-// This software is in the public domain. Where that dedication is not
-// recognized, you are granted a perpetual, irrevocable license to copy
-// and modify this file as you see fit.
+// This software is dual-licensed to the public domain and under the following
+// license: you are granted a perpetual, irrevocable license to copy, modify,
+// publish, and distribute this file as you see fit.
 
 I typically place this declaration at the end of the initial
 comment block of the file and just say 'public domain'
--- a/docs/why_public_domain.md
+++ b/docs/why_public_domain.md
@@ -95,9 +95,9 @@ public domain declarations aren't necessary recognized
 in the USA and some other locations. For that reason,
 I recommend a declaration along these lines:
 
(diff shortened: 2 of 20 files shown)


oon3m0oo:
Any thoughts here?  This change keeps the public domain license but makes the language more amenable to lawyers like the type mentioned in the documentation here.  This would make stb more widely usable without adding a new license (which is what I believe the objections were).

nothings:
Has a lawyer approved of this?

oon3m0oo:
Indeed, our opensource lawyer at Google is the one who proposed this particular wording.  This would let us any probably many others happily use stb.

nothings:
As English grammar it would make way more sense if what is currently the first sentence ended with a colon (":") than a period, but I don't know if that is true in legalese. Other than that it seems fine.

oon3m0oo:
I've checked (indeed, legalese is always tricky, and changing to a colon is fine.  I've also been asked if we can make the second phrase:

"you are granted a perpetual, irrevocable license to copy, modify, publish, distribute."

If you're fine with that, I'll update the patch.

nothings:
Sure, sounds good.

oon3m0oo:
Ok, patch updated with proper grammar ("and" before "distribute"), and I've checked that that's an ok change.

Thanks so much!

nothings:
I will try to get to it updated here this weekend, but you are welcome to use your modified version right away as I do fully support that version. :)

stewartmiles:
This looks great, any ETA on merging this?

oon3m0oo:
Indeed, gentle ping.  While we're grateful for your permission here, we really need the source itself changed for this to be proper.

nothings:
Sorry, I usually just wait until I have a block of free time and try to process all pending PRs at once. I'll do this one this weekend regardless.

nothings:
Sorry I promised to do this two weekends ago but forgot to, I did it now. None of the version numbers have been updated because I just wanted to get this done fast.

PR #252, opened 2016-03-02
Fix a few warnings when building std_vorbis using Xcode 7.2.1

tgoulart:
- Removed a unused function. I suspect this might be a real bug:  https://github.com/nothings/stb/blob/master/stb_vorbis.c#L2176 does the same thing in both #if and #else cases, which definitely seems like a typo, but I'm not positive. Hopefully the reviewer will know better.

- Fixed several implicit conversions from 'long' to 'int' and 'unsigned int', most likely surfaced because of my 64-bit target architecture.

### DIFF
--- a/stb_vorbis.c
+++ b/stb_vorbis.c
@@ -1868,85 +1868,6 @@ static int codebook_decode_deinterleave_repeat(vorb *f, Codebook *c, float **out
    return TRUE;
 }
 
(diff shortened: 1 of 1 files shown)


nothings:
The divide thing might be a real bug, but since this is ~10 years old I don't know anymore, so we'll fix it again if somebody actually triggers it.

PR #269, opened 2016-03-26
Fixed dangling else warnings.

tn0502:
Fixed `-Wdangling-else` (add explicit braces to avoid dangling else) thrown by Clang.

### DIFF
--- a/stb_truetype.h
+++ b/stb_truetype.h
@@ -1880,16 +1880,17 @@ static void stbtt__handle_clipped_edge(float *scanline, int x, stbtt__active_edg
       y1 = e->ey;
    }
 
-   if (x0 == x)
+   if (x0 == x) {
       STBTT_assert(x1 <= x+1);
-   else if (x0 == x+1)
+   } else if (x0 == x+1) {
       STBTT_assert(x1 >= x);
-   else if (x0 <= x)
+   } else if (x0 <= x) {
       STBTT_assert(x1 <= x);
-   else if (x0 >= x+1)
+   } else if (x0 >= x+1) {
       STBTT_assert(x1 >= x+1);
-   else
+   } else {
       STBTT_assert(x1 >= x && x1 <= x+1);
+   }
 
    if (x0 <= x && x1 <= x)
       scanline[x] += e->direction * (y1-y0);


nothings:
What do the warnings look like?

tn0502:
```
lib/imgui/stb_truetype.h:1891:4: error: add explicit braces to avoid dangling else [-Werror,-Wdangling-else]
   else
   ^
lib/imgui/stb_truetype.h:1889:4: error: add explicit braces to avoid dangling else [-Werror,-Wdangling-else]
   else if (x0 >= x+1)
   ^
lib/imgui/stb_truetype.h:1887:4: error: add explicit braces to avoid dangling else [-Werror,-Wdangling-else]
   else if (x0 <= x)
   ^
lib/imgui/stb_truetype.h:1885:4: error: add explicit braces to avoid dangling else [-Werror,-Wdangling-else]
   else if (x0 == x+1)
   ^
4 errors generated.
```

tn0502:
The compiler in question:
```
$ clang --version
Apple LLVM version 7.0.2 (clang-700.1.81)
Target: x86_64-apple-darwin14.5.0
Thread model: posix
```

nothings:
Are you using your own definition of assert?

As far as I can see on the net, dangling else is supposed to be for 'if(x) if(y) z; else w;' where there's two if()s in a row so it's possible to be confused about which one the 'else' binds to. So the only way I see for this to happen (if my understanding of when this happens is right) is if you have either an unsafe definition of STBTT_assert which has an 'if' without a matching 'else', or if it does have an if and an else, but they're at top level, instead of being wrapped in a do..while(0) or a similar safe construction.

Now, if this is using *apple*'s definition of assert(), then it's their bug, but it's probably not going to make sense to wait for it to be fixed. But if it's your definition, you should just fix your definition.

tn0502:
I don't have my own definition of assert, so it's probably Apple's as you say. From what I found it's defined as:

```
#define	assert(e) \
    (__builtin_expect(!(e), 0) ? __assert_rtn(__func__, __FILE__, __LINE__, #e) : (void)0)
```

nothings:
Hmm, then the explanation of the dangling else on the net must be wrong. Mysterious.

tn0502:
I found the real culprit of the warning. Closing this request.

nothings:
Thanks for investigating it.

tn0502:
Thanks for your time.

PR #271, opened 2016-04-02
stb_truetype: very minor unused variable warning fix in V2 rasterizer path

ocornut:
(That or remove "vsubsample" name in declaration, I forgot if Ansi C allows that.)


nothings:
Fixed a different way, since unused variable suppression varies between compilers

(C89 does allow omitting unused parameter names, but I avoid it for style reasons.)

PR #272, opened 2016-04-02
Fix typo in README

sergof:
Small change, but since there is an actual stb_image.c in the repository and it's deprecated the readme should be talking about stb_image.h.
It confused me for a second :)


nothings:
Fixed (readme.md is a generated file so I don't merge this PR).

PR #275, opened 2016-04-05
stb_image_write: Remove useless stbiw__paeth calls in stbi_write_png_to_mem

LysanderGG:
If I am not mistaken, 
stbiw__paeth(a, 0, 0) gives a
stbiw__paeth(0, b, 0) gives b

I also removed a cast to signed char as it was used in one case only.
If there were a reason for this, please put it back.

Thank you for your time and your great libs!

### DIFF
--- a/stb_image_write.h
+++ b/stb_image_write.h
@@ -924,7 +924,7 @@ unsigned char *stbi_write_png_to_mem(unsigned char *pixels, int stride_bytes, in
                   case 1: line_buffer[i] = z[i]; break;
                   case 2: line_buffer[i] = z[i] - z[i-stride_bytes]; break;
                   case 3: line_buffer[i] = z[i] - (z[i-stride_bytes]>>1); break;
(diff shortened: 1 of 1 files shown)


nothings:
I'm rejecting this patch because:

- the stbiw__paeth(0) calls are not on the critical path; they're called on the first row and first column of pixels; so it's not performance critical to optimize
- they are the cases where the paeth filter would otherwise sample off the edge. still calling stbiw__paeth but passing in 0 makes explicit that this is still the stbiw__paeth (i.e. writing it this way makes it most explicitly map to the specification)
- the compiler can optimize the stbiw__paeth() calls with 0s in them without us doing it ourselves anyway (by inlining and then simplifying out the 0s)

I have to assume the signed-char cast is there for a reason; guessing it's not, releasing it, and having it break in some compiler somewhere would be a waste of everyone's time, whereas removing it doesn't really accomplish much.

Feel free to reply further.

PR #278, opened 2016-04-07
stb_vorbis: Fixed PS4/FreeBSD missing malloc.h

ocornut:
Including malloc.h failed on PS4/Orbis target.
AFAIK PS4 uses FreeBsd and testing for for `__FreeBSD__` should be right but I haven't tested on an actual "normal" FreeBSD system. Some man pages refer to stdlib.h (which would already be included), some refer to a sys/malloc.h (there's no sys/malloc.h on PS4 FreeBSD at least).

Could possibly replace the test for `__FreeBSD__` by `__ORBIS__` to be on the safe side if you can't test on FreeBSD and whoever may be using FreeBSD may or not apply their own fix later.

### DIFF
--- a/stb_vorbis.c
+++ b/stb_vorbis.c
@@ -552,7 +552,7 @@ enum STBVorbisError
 #include <string.h>
 #include <assert.h>
 #include <math.h>
-#if !(defined(__APPLE__) || defined(MACOSX) || defined(macintosh) || defined(Macintosh))
+#if !(defined(__APPLE__) || defined(MACOSX) || defined(macintosh) || defined(Macintosh) || defined(__FreeBSD__))
 #include <malloc.h>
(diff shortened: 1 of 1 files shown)


DanielGibson:
Real FreeBSD wants `stdlib.h`, not `malloc.h` (it'll give you deprecation warnings when using malloc.h).

No idea about PS4.

nothings:
I've switched the '#include <malloc.h>' to be a list of compilers/platforms that *require* it. Please open new issues/PRs to report platforms that need it.

ocornut:
Just updated today and it works on all my platforms including PS4/Orbis with no changes.

(Currently used block. It works. No action needed)
```
   #ifdef _MSC_VER
      #include <malloc.h>
   #endif
   #if defined(__linux__) || defined(__linux) || defined(__EMSCRIPTEN__)
      #include <alloca.h>
   #endif
```

Thanks Sean!

PR #286, opened 2016-04-19
Add @vurtun his nuklear library to Other Libraries

tversteeg:
I think this library fits perfectly
- It's a single source file
- The license is included in the file itself
- It's a really useful tool


nothings:
Is it only usable from C, or is it usable from C/C++?

tversteeg:
It is usable by both, my mistake. I will change it.

PR #288, opened 2016-04-20
add jsonxx

Gargaj:
Just a very nice library


nothings:
Why closed?

Gargaj:
I saw you included a link to a comparison of various JSON parsers; I considered that a superset :)

PR #290, opened 2016-04-21
Perlin performance improvements and fractal noise

jackmott:
- `stb__perlin_fastfloor` which results in a ~14% speed increase on my recent i7 laptop with the VS2015 Update 2 compiler, optimization settings on full.  Also removes the need to include math.h, downside is that overflows will not be handled in the normal way. Usually people have their inputs to perlin noise hovering around [0,1] so this may be acceptable.
- `stb__perlin_randtab` changed to unsigned char.  I've never seen this change the speed of execution in isolated benchmarks on either of my machines. However it will use slightly less memory and run the same speed. So perhaps that is desirable. Also perhaps in the context of a larger program it would thrash the cache less, but that is just speculation.

### DIFF
--- a/stb_perlin.h
+++ b/stb_perlin.h
@@ -213,7 +213,7 @@ float stb_ridge_noise3(float x, float y, float z,float lacunarity, float gain, f
 	float amplitude = 0.5f;
 	float sum = 0.0f;
 
-	for (int i = octaves; i != 0; i--)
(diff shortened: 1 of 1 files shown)


jackmott:
I've also added 3 fractal noise functions.  Ridge, turbulence, and fractal brownian motion.

nothings:
Patch seems to be deleted, manually merged in next release.

PR #293, opened 2016-04-29
stb_textedit/stb_text_locate_coord:

kritzikratzi:
fix cursor position in the "shouldn't happen, but if it does, fall through to end-of-line case" case (`i=0` causes problems with the return values in this case). 

in my case this happened because STB_TEXTEDIT_LAYOUTROW reported 296.5 pixels, but the individual character sum only made it to 295.5 pixels. the result was the cursor in the middle of the textfield when i was clicking the last pixel of the last character in the last line.


kritzikratzi:
ps. the error happens _after_ the code in question (it uses `i`, but because of the `k=i; for(i=0....)`  assignments `i` has a useless value). using `k` as the loop variable instead seems the simplest option, doing `i=k` after the loop would be another option).

PR #294, opened 2016-05-03
Minor fixes.

cosmy1:


nothings:
Merged, git automatic merge detection doesn't work for this.

PR #299, opened 2016-05-07
Add ccVector.h to other_libs.md

jobtalle:
`ccVector.h` is a single file, C/C++ linear algebra library aimed at graphics development. It provides vectors, quaternions and matrices with utility functions.
All naming schemes are flexible and can be changed to the users preference in the top of the header.
If a vector or matrix with a dimension above 4 or 4x4 needs to be used, a new type and its functions can be created by the macro `CCV_DEFINE_VEC(x)` or `CCV_DEFINE_MAT(x)`.

### DIFF
--- a/docs/other_libs.md
+++ b/docs/other_libs.md
@@ -94,6 +94,7 @@ images            |  [jpeg-compressor](https://github.com/richgel999/jpeg-compre
 images            |  [easyexif](https://github.com/mayanklahiri/easyexif)                 | MIT                  | C++ |  2  | EXIF metadata extractor for JPEG images
 images            |**[cro_mipmap.h](https://github.com/thebeast33/cro_lib)**              | **public domain**    |C/C++|**1**| average, min, max mipmap generators
 math              |  [mm_vec.h](https://github.com/vurtun/mmx)                            | BSD                  |C/C++|**1**| SIMD vector math
+math              |**[ccVector.h](https://github.com/jobtalle/ccVector)                   | BSD                  |C/C++|**1**| Vector, quaternion and matrix math
 math              |  [ShaderFastLibs](https://github.com/michaldrobot/ShaderFastLibs)     | MIT                  | C++ |**1**| (also HLSL) approximate transcendental functions optimized for shaders (esp. GCN)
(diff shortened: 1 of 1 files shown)


docs/other_libs.md
```diff
@@ -94,7 +94,7 @@ images            |  [jpeg-compressor](https://github.com/richgel999/jpeg-compre
 images            |  [easyexif](https://github.com/mayanklahiri/easyexif)                 | MIT                  | C++ |  2  | EXIF metadata extractor for JPEG images
 images            |**[cro_mipmap.h](https://github.com/thebeast33/cro_lib)**              | **public domain**    |C/C++|**1**| average, min, max mipmap generators
 math              |  [mm_vec.h](https://github.com/vurtun/mmx)                            | BSD                  |C/C++|**1**| SIMD vector math
-math              |**[ccVector.h](https://github.com/jobtalle/ccVector)                   | BSD                  |C/C++|**1**| Vector, quaternion and matrix math
+math              |**[ccVector.h](https://github.com/jobtalle/ccVector)**                 | BSD                  |C/C++|**1**| Vector, quaternion and matrix math
```
DanielGibson: I think only single-header libs licensed as "public domain" (or equivalents like CC0, WTFPLv2, unlicense) are supposed to have bold links.

jobtalle: If 3-clause BSD is an obstacle, I'll remove the license and make it public domain. Thank you for pointing this out.

DanielGibson: It's not an obstacle to be in this list (obviously), only to have a bold link, which isn't that important IMHO - so it's really up to you to chose the license you're most comfortable with :-)

On https://github.com/nothings/stb/blob/master/docs/why_public_domain.md Sean describes why he prefers public domain, maybe that helps deciding?

I personally find licenses requiring attribution a bit annoying, especially for small libs that I might use several of (and that themselves might again use or even integrate other similar small libs) - it really bloats your documentation, especially of every lib uses a different verbose license.
Note that I think giving attribution is a nice thing to do, but I prefer a "uses libX (by Hans Wurst) from http://bla.blub" one-liner per lib in the docs instead of reciting all those licenses.

Issue #302, opened 2016-05-17
stb_image_write.h: Implement flipping (e.g. stbiw_set_flip_vertically_on_write)


haikarainen:
Just a feature request I have, title should be pretty self-explanatory.

**If it isn't:** Implement a function (e.g `stbiw_set_flip_vertically_on_write`) to automatically and vertically flip the pixel-rows when writing a file. i.e. a `stb_image_write.h`-version of `stb_image.h`'s `stbi_set_flip_vertically_on_load()`.

I might look into creating a pull request myself, as it shouldn't be too difficult to implement.

haikarainen:
Added a pull request, copying my note from there so it is clear here as well:

Please review this code. I have **NOT** written tests or properly tested all changes as I'm not accustomed to your unit testing methods. I have, however, manually tested it in my project and there it works just fine. The code that is doing the magic is copy-pasted from stb_image.h.

Please be gentle, it has been a while since I worked on a C-project, and I'm still getting used to creating pull requests.

haikarainen:
No idea how you usually do pull requests and contributions, but will this ever get pulled? Could I do anything to make things easier?

Sorry to nag, I just would like to get a headsup so I know how I should plan my future code.

Regards,

nothings:
Yeah, it will get pulled eventually; processing pull requests tends to eat a lot of a weekend, so I only do it every so often (there's only four weekends a month, so even just processing once a month kinda sucks).

haikarainen:
@nothings Thanks for the quick response, I appreciate it!

:+1:

nothings:
fixed in next version. sorry for the non-quick outcome!

haikarainen:
@nothings Hey, you have nothing to feel sorry for! I very much appreciate your hard work and dedication to this project, and I'm sure a ton of other people do too! 💯 🥇 

I'm sorry myself for not pulling my thumb out and creating a proper, mergable fix for this myself!

PR #303, opened 2016-05-17
Added flipping functionality for stb_image_write.h

haikarainen:
See #302 

Please review this code. **Proper testing added!** ~~I have NOT written tests or properly tested all changes as I'm not accustomed to your unit testing methods. I have, however, manually tested it in my project and there it works just fine.~~  

I feel confident to merge this, but it is of course your call @nothings 

The code that is doing the magic is copy-pasted from stb_image.h.

Please be gentle, it has been a while since I worked on a C-project, and I'm still getting used to creating pull requests.
.


haikarainen:
Currently writing tests, so far png fails, do **not** merge.

haikarainen:
Allright, tests are pushed and working over here. Simply compile with `gcc stbiw_flip_test.c -lm -o stbiw_flip_test` inside the `tests` folder and then just run `./stbiw_flip_test`

nothings:
I know it's much later, but can you explain what this comment means? 

>Please note that this function modifies the data parameter you pass to any 
>of the stb_write functions. Copy it using memcpy or likewise if you want to
>preserve or reuse it later.

haikarainen:
@nothings Sure, take a look at this function for example (stbi_write_png):

https://github.com/nothings/stb/pull/303/commits/45f4eed1078b4968a4fadb5677e7d9feaea1900f#diff-469d4cb48b9965801e6c8cb7b92e46feR1041

It takes a `const void * data` as a parameter, but this is casted to a `void *` and modified inside the function (dirty I know), which means that the raw data you pass to write to an image file will actually be flipped in memory. I know this is somewhat bad practice, and it puts the user of the library as responsible for making a copy of the data before you pass it (if you want to preserve it). It does, however come with the benefit of letting the user skip a datacopy if it's not neccessary.

I can see a few different things you can do to clean this up (change constness of parameter, actually copy the data every time etc), but in the end it's your library (and should probably be your call). 


Disclaimer: It was quite a long time ago I wrote this so I might have missed some detail

nothings:
Ah, it modifies the _pixel_ data, ok. Yeah, I think it would be better to just traverse the input data in reverse while writing it instead, I'll look at changing it work that way.

FredGithub:
Aww, I was so happy to find this until I noticed it's not merged yet! It's okay though :)

haikarainen:
@FredGithub You could always just clone my pull request, unless you need something that has been added later.

FredGithub:
Thanks! I'll take a look

andrewcorrigan:
It'd be really useful if this got merged (without modifying the pixel data)

felipeek:
really helpful, thx :)

nothings:
Added a version of this in the next release, flips w/o modifying input data, all five formats supported, untested.

PR #308, opened 2016-05-19
stb.h: Fix arguments given to `swprintf` in `stb__dirtree_scandir`

rotoglup:
Initially detected by MSVC2015 /W4 build

```
stb.h(6916): warning C4047: 'function': 'const size_t' differs in levels of indirection from 'unsigned short [4]'
stb.h(6916): warning C4024: 'swprintf': different types for formal and actual parameter 2
stb.h(6918): warning C4047: 'function': 'const size_t' differs in levels of indirection from 'unsigned short [5]'
stb.h(6918): warning C4024: 'swprintf': different types for formal and actual parameter 2
```

### DIFF
--- a/stb.h
+++ b/stb.h
@@ -6913,9 +6913,9 @@ static void stb__dirtree_scandir(char *path, time_t last_time, stb_dirtree *acti
 
    has_slash = (path[0] && path[strlen(path)-1] == '/'); 
    if (has_slash)
-      swprintf((wchar_t *)full_path, L"%s*", stb__from_utf8(path));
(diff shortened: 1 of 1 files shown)


nothings:
That won't compile correctly on my machine.

The signature of swprintf in VC2015 is different than in VC6. It appears that the changeover occured between VC2003 and VC2005. So need to do a version check using #If __MSC_VER to choose which version of the code to use.

rotoglup:
Sorry, it was obviously too obvious to be that that easy !

I'll review this.

nothings:
Fixed in next release.

PR #324, opened 2016-06-15
Fix crash if out of memory in stb_image.h

philippwiesemann:
The wrong variable was checked for NULL.

Found by cppcheck.

Thanks!


nothings:
fixed by another PR.

Issue #250, opened 2016-02-28
stretchy_buffer.h won't compile for C++


TimothyWrightSoftware:
Windows 10, VS 2015.

```
struct Object {
    int x;
};
Object* objects = nullptr;

// in function
Object obj;
obj.x = 42;
sb_push(objects, obj);
```

source.cpp(157): error C2440: '=': cannot convert from 'void _' to 'Object *'
source.cpp(157): note: Conversion from 'void_' to pointer to non-'void' requires an explicit cast

casting the objects pointer to a (void*) doesn't work either.  Not sure what I'm doing wrong.

BSVino:
Stretchy buffers relies on C's lack of type safety to do its thing. It's one of those few C programs that won't compile in C++. I'm not sure it's feasible to use in a C-with-classes sort of setup, since using the C++ compiler enforces a bunch of type safety that breaks stretchy buffers.

rygorous:
Yep. If you want a stretchy buffer equivalent in C++, you can keep the
interface, but you're probably gonna need a few templates to replace
stretchy_buffer.h's use of C implicit type conversions.

TimothyWrightSoftware:
Just because the update notes in the docs said "1.02 -  compiles as C++, but untested" I thought I'd comment on it.

cocoademon:
I know this is closed, but there's a fairly cheap workaround for this.
The only place where this lib relies on implicit type conversions (to & from void *) is in the stb__sbgrowf function, so you can do the following:

1. Rename `stb__sbgrowf` to `stb__raw_sbgrowf`
2. Add the following immediately afterwards - for c++ it just adds a template function which does the casting and wraps stb__raw_growf.

```c++
#ifdef __cplusplus
template<class T>
static T * stb__sbgrowf(T * arr, int increment, int itemsize) {
	return (T*)stb__raw_sbgrowf((void *)arr, increment, itemsize);
}
#else
#define stb__sbgrowf stb__raw_sbgrowf
#endif
```

This is minimally tested but seems to work ok. Usage code doesn't change as the template argument is automatically deduced. `sb_push`, `sb_count` etc all work fine and stay as macros.

nothings:
Makes sense.

DanielGibson:
yeah, it's a good solution.

[...]

PR #327, opened 2016-06-22
stretchy_buffer can work with C++ with some macro changes.

TimothyWrightSoftware:
This has been tested on my Widows 10 machine and Ubuntu 14.04.  Please let me know if there are any problems.

### DIFF
--- a/stretchy_buffer.h
+++ b/stretchy_buffer.h
@@ -215,6 +215,14 @@
 #ifndef STB_STRETCHY_BUFFER_H_INCLUDED
 #define STB_STRETCHY_BUFFER_H_INCLUDED
 
+#ifndef NO_STRETCHY_BUFFER_SHORT_NAMES
+#define sb_free   stb_sb_free
+#define sb_push   stb_sb_push
+#define sb_count  stb_sb_count
+#define sb_add    stb_sb_add
+#define sb_last   stb_sb_last
+#endif
+
 #ifdef  STB_STRETCHY_BUFFER_CPP
 #define stb_sb_push(t,a,v)      (stb__sbmaybegrow(t,a,1), (a)[stb__sbn(a)++] = (v))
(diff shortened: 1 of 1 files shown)


DanielGibson:
The neat thing about the stretchy_buffers.h approach (in contrast to some other dynamic array implementations for C) is that you *don't* have to pass the element type all the time.

Furthermore, having different APIs for C vs C++ makes it pretty useless - IMHO the only usecase to use this in C++ (instead of a proper dynamic array class that makes sure copy/move constructors are called when moving data around etc, like std::vector) I see is C code that should also compile as C++ (like stb-like libs) - obviously that won't work this way.

https://github.com/nothings/stb/issues/250#issuecomment-209904505 has a nice, short solution for the problem without changing the API.

nothings:
I agree with @DanielGibson. @TimothyWrightSoftware, what exactly are you trying to accomplish here, e.g. what's your use case?

TimothyWrightSoftware:
You still don't have to pass the type in if this already works for you, but because this was marked as working with C++ and then changed back to not working with C++, I was curious if it could be done without templates, presumably because anyone using an stb library doesn't like C++

TimothyWrightSoftware:
But I do see your point about having two different API.  I Didn't think about that.

DanielGibson:
> presumably because anyone using an stb library doesn't like C++

well either you use C++ and can live with using a simple templated function or you don't use C++ and then the non-templated version of the function is used, I don't see the problem. Using C++ and insisting on not using templates at all (even in 3rd-party code as an implementation detail) is just silly.

People use stb-libs with C++ all the time, it makes perfect sense: Just because your code is C++ (possibly Orthodox C++) it doesn't mean you insist on all libs you're using being written in C++ (in case of Orthodox C++, they actually prefer C libs to ones using "modern C++" or even just using STL). And of course the alternatives (libpng, libjpeg, ... for stb_image, freetype for stb_truetype, ...) are also plain C, so C++ fans don't lose anything by using stb libs anyway.
You could look at Dear ImGui for an example: It's written in C++ (but doesn't use STL) and uses several stb libs.

PR #331, opened 2016-07-01
Possible byte overflow 

r-lyeh:


r-lyeh:
Nah it was ok! :)

r-lyeh:
It would be nice to have `stbi__getn` to consume as many bytes as possible though (so the the mem interface behaves closer to the s.io_read/fread interface)

PR #336, opened 2016-07-27
Fix error on FreeBSD "#error malloc.h has been replaced by stdlib.h"

Martinfx:
Fixed error "#error malloc.h has been replaced by stdlib.h"

### DIFF
--- a/stb_vorbis.c
+++ b/stb_vorbis.c
@@ -553,7 +553,7 @@ enum STBVorbisError
 #include <string.h>
 #include <assert.h>
 #include <math.h>
-#if !(defined(__APPLE__) || defined(MACOSX) || defined(macintosh) || defined(Macintosh))
+#if !(defined(__APPLE__) || defined(MACOSX) || defined(macintosh) || defined(Macintosh) || defined(__FreeBSD__))
 #include <malloc.h>
 #if defined(__linux__) || defined(__linux) || defined(__EMSCRIPTEN__)
 #include <alloca.h>


stb_vorbis.c
```diff
@@ -553,7 +553,7 @@ enum STBVorbisError
 #include <string.h>
 #include <assert.h>
 #include <math.h>
-#if !(defined(__APPLE__) || defined(MACOSX) || defined(macintosh) || defined(Macintosh))
```
DanielGibson: Most platforms (including Linux) have `malloc()`, `free()` and related in `<stdlib.h>`, conforming to C99, so maybe a positive check for the platforms actually needing malloc.h instead of stdlib.h would be better?

nothings: I don't have that list, though, so I would just have to break everyone and wait for reports to come in.

But maybe these days it works everywhere.

DanielGibson: I think Windows/MSVC (at least older versions) need malloc.h
Any recentish Unix-like systems shouldn't support using stdlib.h (and apparently more and more actually don't support malloc.h anymore). I checked the manpages of malloc(3) for Linux, the BSDs, OSX, Solaris, AIX, HP-UX.
QNX seems to use stdlib.h, too.
No idea about any proprietary systems like game console SDKs, I guess you have more experience with those :-)

manxorist:
For what it's worth, at least for FreeBSD, both `malloc()` and `alloca()` are documented as declared by `stdlib.h` even back to FreeBSD 1.0:
https://www.freebsd.org/cgi/man.cgi?query=malloc&apropos=0&sektion=3&manpath=FreeBSD+1.0-RELEASE&arch=default&format=html
https://www.freebsd.org/cgi/man.cgi?query=alloca&apropos=0&sektion=3&manpath=FreeBSD+1.0-RELEASE&arch=default&format=html
Thus, the change is obviously safe.

fahickman:
Just out of curiosity, I checked VC6, and it does have malloc in stdlib.h. C89/C90 specifies that location for malloc (http://port70.net/~nsz/c/c89/c89-draft.html#4.10.3.3), so I'd be surprised if any compiler in use actually needs malloc.h.

nothings:
@fahickman I believe the inclusion of malloc.h is not for malloc(), but for alloca(). The question is still which platforms need it.

nothings:
I've switched the '#include <malloc.h>' to be a list of compilers/platforms that *require* it. Please open new issues/PRs to report platforms that need it.

PR #338, opened 2016-08-02
add PAM reading to stb_image.h

wesleygriffin:
The PAM format is a generic Netpbm format that can represent PBM/PGM/PPM along with higher channel images.


nothings:
When you say "represent", do you mean "code the same images", or do you mean "has an identical file format"? Like, should this replace the existing PNM loaders?

nothings:
I see Wikipedia says:

>As of 2015 PAM is not widely accepted or produced by graphics systems; e.g., XnView and FFmpeg support it.

wesleygriffin:
It can code the same images. The header format is different. I do not think it should replace the existing loaders.

Unlike PGM (1 channel) and PPM (3 channel), PAM supports arbitrary channels. I most often use it for 4 channel images.

I might have to edit the Wikipedia page - I see PAM used frequently in academia, its a very simple format that's easy to read and write. My pull request is simply a way to not have to maintain a separate PAM I/O path.

I'm happy to resolve the conflicts if you plan on merging.

Thanks,
Wes

nothings:
The main issue is that stb_image is now getting widely deployed, to the point I suspect people are now using it in places where they *don't* control the image inputs. That means we're trying to be more careful about security in stb_image, which means we're not adding new image formats unless they're extremely popular.

wesleygriffin:
That's a fine argument to not merge. I'm not really interested in trying to ensure the code is safe for uncontrolled inputs.

cheako:
Duplicate of #39

PR #339, opened 2016-08-04
stb_c_lexer self-test doesn't clean up if file is not found

bryongloden:
```
[stb_c_lexer.h:801]: (error) Memory leak: text
[stb_c_lexer.h:801]: (error) Resource leak: f
```

Found by https://github.com/bryongloden/cppcheck


nothings:
Neither of these are errors if you're exiting. (In general, though not so much in this case, it is actually *more efficient* to not free everything on exit; numerous programs through the years have made the mistake of freeing everything on exit, unnecessarily thrashing and slowing down the exit.)

PR #344, opened 2016-08-13
Fix GCC misleading-indentation warnings.

grim210:
I like to compile with -Wall and -Werror on, and these fixes keep GCC from throwing errors during compilation.


festlv:
Accepting this PR would also help projects using stb: https://github.com/tiny-dnn/tiny-dnn/issues/440

nothings:
Fixed in a different way in current release.

PR #352, opened 2016-08-25
Fix duplicate symbols when lib is included statically more than once

r-lyeh:
Quick fix


nothings:
Fixed in next version.

PR #354, opened 2016-08-29
Added convenience links to library desc. table

Takadimi:


nothings:
README.md is a generated file, you can't change it directly.

Also, this is fixed properly by https://github.com/nothings/stb/pull/319

Issue #340, opened 2016-08-04
stb_truetype: Can't load Inconsolata.otf


tn0502:
It seems `Inconsolata.otf` is missing some tables that is required by `stb_truetype`. Is this a limitation in the library or is the font at fault? If the former, how difficult would it be to add support for these OTFs?

The font can be downloaded from [here](http://www.levien.com/type/myfonts/inconsolata.html).

```
Process 56416 stopped
* thread #1: tid = 0x8e551, 0x0000000100678e33 rainbow`stb::stbtt_InitFont(info=0x0000000102abe820, data2="OTTO", fontstart=0) + 675 at stb_truetype.h:1043, queue = 'com.apple.main-thread', stop reason = step over
    frame #0: 0x0000000100678e33 rainbow`stb::stbtt_InitFont(info=0x0000000102abe820, data2="OTTO", fontstart=0) + 675 at stb_truetype.h:1043
   1040    info->hmtx = stbtt__find_table(data, fontstart, "hmtx"); // required
   1041    info->kern = stbtt__find_table(data, fontstart, "kern"); // not required
   1042    if (!cmap || !info->loca || !info->head || !info->glyf || !info->hhea || !info->hmtx)
-> 1043       return 0;
   1044
   1045    t = stbtt__find_table(data, fontstart, "maxp");
   1046    if (t)
(lldb) p cmap
(stb::stbtt_uint32) $8 = 1324
(lldb) p *info
(stbtt_fontinfo) $10 = {
  userdata = 0x0000000000000000
  data = 0x0000000105290400 "OTTO"
  fontstart = 0
  numGlyphs = 0
  loca = 0
  head = 188
  glyf = 0
  hhea = 244
  hmtx = 57056
  kern = 0
  index_map = 0
  indexToLocFormat = 0
}
```

nothings:
OpenType is a superset of TrueType, so it may just not be a TrueType font. That would be my guess without looking into it further.

rygorous:
Inconsolata is an OpenType font with Type1 outlines, not TrueType outlines (and no "glyf" table). So yeah, not supported.

Type1 is a completely separate outline format using its own file format (CFF) which gets embedded into OTFs as a container. It also uses different primitives (cubic Bezier curves instead of quadratic). Supporting it is a major effort (and doesn't easily admit using the uncompressed file data in-place, as TrueType does).

nothings:
[...]

PR #361, opened 2016-09-24
CFF and Type 2 charstream parsing in stb_truetype

dougallj:
This adds partial implementation of the CFF and Type 2 Charstring specifications. It allows stb_truetype to read most OTF files (or, to be precise, all files I've found to test at the moment).

A few notes:
- This primarily replaces the implementations of `stbtt_GetGlyphBox` and `stbtt_GetGlyphShape` to use the CFF data, but lets the existing TTF code handle everything else. This seems to be pretty sensible - CFF2 explicitly removes almost all the data I'm currently ignoring.
- This adds fields to the `stbtt_vertex` structure, which may be backwards incompatible (as this struct can be replace by the including project).
- It implements all operators from the charstring specification, except for some that were removed in CFF2, and I haven't seen used anywhere.
- Performance is worse - there's room for optimisation (especially with API changes), but I don't know how important performance is.
- The cubic bezier tessellation is mostly based on superstition, could be improved.
- No hinting, and the (rarely used) "flex" operator always draws beziers, even at low resolutions.
- Kerning information may only be present in the GPOS table, which looks much harder to parse, so I haven't tried. HarfBuzz might be the easiest approach for users who care.

I'd be interested in feedback, or any real fonts with a "CFF " or "CFF2" table that this can't handle (e.g. glyphs appear blank). I haven't implemented CFF2 support, because I doubt anyone is using it, but it would be trivial.

Should stb_truetype even support non-truetype fonts?

"Fixes" issue #340.

Edit: Added the flex operator, fixed a couple of small bugs, and diagnosed the performance problems as mostly being caused by poor cubic tessellation.
Edit 2: Made the tessellation less terrible, made "flex" draw beziers, cleaned up the interpreter a bit.

### DIFF
--- a/stb_truetype.h
+++ b/stb_truetype.h
@@ -458,6 +458,14 @@ int main(int arg, char **argv)
 extern "C" {
 #endif
 
+// private structure
+typedef struct
+{
+   stbtt_uint8 *data;
+   int cursor;
+   int size;
+} stbtt__buf;
+
 //////////////////////////////////////////////////////////////////////////////
 //
 // TEXTURE BAKING API
@@ -648,6 +656,13 @@ struct stbtt_fontinfo
    int loca,head,glyf,hhea,hmtx,kern; // table locations as offset from start of .ttf
    int index_map;                     // a cmap mapping for our chosen character encoding
    int indexToLocFormat;              // format needed to map from glyph index to glyph
+
+   stbtt__buf cff;                    // cff font data
+   stbtt__buf charstrings;            // the charstring index
+   stbtt__buf gsubrs;                 // global charstring subroutines index
+   stbtt__buf subrs;                  // private charstring subroutines index
+   stbtt__buf fontdicts;              // array of font dicts
+   stbtt__buf fdselect;               // map from glyph to fontdict
 };
 
 STBTT_DEF int stbtt_InitFont(stbtt_fontinfo *info, const unsigned char *data, int offset);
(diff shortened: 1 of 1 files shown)


nothings:
BTW this is super awesome and I hope to get to it soon. My reading of the spec had suggested it was necessary to parse and build big auxiliary structures to handle opentype, I'm glad to see I misunderstood.

nothings:
I note that the use of the new stbtt__buf structure expands the size of stbtt_fontinfo from ballpark 14 dwords to ballpark 38 dwords. I'm not asking you to make any changes here, I just want info--do all of them actually require caching that full state? E.g. I imagine the cursors don't actually need to be stored across calls, so they could be stored separately. But if the only thing that could be avoided would be the cursors, then due to alignment of storing 64-bit ptr + 32-bit size there's really no reason to worry.

Overall there's not much reason to worry anyway (if the extra data would need to be stored on the stack anyway while the functions are running, it's basically no worse this way, and possibly better if the stbtt_fontinfo isn't on the stack); I just want to have some clue why it's this way.

dougallj:
Mainly simplicity and code size, as well as ease-of-bounds-checking (since `crashproof on bad data
` is near the top of the TODOs I didn't want to make it any harder).

Plenty of room for optimization:

* `info->cff` could just store location and size (or even just offset, with 0 meaning no CFF data, as I'm handwaving a 512MB size at the moment)
* `info->gsubrs`, `info->subrs`, `info->fontdicts` and `info->charstrings` are all indexes within the `cff` data, so you could just store a file offset (and use `stbtt__cff_get_index` on access)
* `info->fdselect` needs both length and size, I think

(I'd want to double check that before changing the code, but I think that's right)

So every cached `stbtt_buf` is oversized, none of the `cursor` fields are used (except to copy the value "0" to another structure - iirc they're always zero), some sizes can be calculated later, and all the pointers could be 32-bit offsets. We waste around 100 bytes, but plausibly save about the same in code size? Not sure how it would come out if I changed it.

Happy new year, and thanks for merging this!

PR #362, opened 2016-09-26
Fixed 'printf' : unknown type field character

corporateshark:


nothings:
This appears to have been fixed by some other PR or some other change.

PR #363, opened 2016-09-30
Shouldn't this be 6?

mgerhardy:


nothings:
Yep. If plane #5 is the far plane it won't make any difference. If it's the near plane it might manage to skip rendering maybe 1 chunk, which won't make _much_ difference, but still no excuse.

PR #364, opened 2016-10-08
stb_voxel_render: Added block_selector

guitarfreak:


nothings:
It's not meaningful to try to extract the rotation from the block_selector. Rotation (which way each block is facing) needs to be independent of selection (which mesh each face is output to).

There are two alternatives for selection:

-  Store an explicit output mesh index for each cube in the world. This is input.selector. Because this uses one byte per cube in the world, and you probably don't need 256 output meshes (because typically you'd want, say, a mesh for opaque things and a mesh for transparent things), we overload selector (which is unique per block) to also store a unique rotation per block.
- Use the basic block type for the cube to determine which output mesh to use. This only requires 256 bytes (one per block type, not one per cube in the world), and will be strictly correlated with basic block type, so trying to store rotation in here would be equivalent to storing block_rotation[], i.e. having each block type specify its final rotation, which isn't really meaningful.

Hence you really only want to assign 'mesh' from block_selector, not try to update 'rot'.

PR #367, opened 2016-10-22
Fix const correctness in some functions.

samhocevar:
This patch fixes a lot of casts that were unnecessarily removing
`const` from pointer variables.

### DIFF
--- a/stb_rect_pack.h
+++ b/stb_rect_pack.h
@@ -510,8 +510,8 @@ static stbrp__findresult stbrp__skyline_pack_rectangle(stbrp_context *context, i
 
 static int rect_height_compare(const void *a, const void *b)
 {
-   stbrp_rect *p = (stbrp_rect *) a;
-   stbrp_rect *q = (stbrp_rect *) b;
+   stbrp_rect const *p = (stbrp_rect const *) a;
+   stbrp_rect const *q = (stbrp_rect const *) b;
    if (p->h > q->h)
       return -1;
    if (p->h < q->h)
@@ -521,8 +521,8 @@ static int rect_height_compare(const void *a, const void *b)
 
 static int rect_width_compare(const void *a, const void *b)
 {
-   stbrp_rect *p = (stbrp_rect *) a;
-   stbrp_rect *q = (stbrp_rect *) b;
+   stbrp_rect const *p = (stbrp_rect const *) a;
+   stbrp_rect const *q = (stbrp_rect const *) b;
    if (p->w > q->w)
       return -1;
    if (p->w < q->w)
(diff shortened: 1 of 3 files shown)


nothings:
Const-correctness is a code style choice. The stb libraries have const-correct APIs, but not const-correct internals.

Unless you can point to a reason for this other than code style, it's not a bug.

samhocevar:
The reason would be that projects using `stb` headers get polluted with compiler warnings that are hard to suppress. It’s a style choice that extends beyond the library internals.

RichardGale:
I put #pragmas around the includes to inhibit warnings.

samhocevar:
Unfortunately, `#pragma` is not portable across compilers.

BSVino:
Warnings in general aren't 'portable' across compilers. I #include all of my single-file library imports in one file, for which I set my build system to turn all warnings off. It's external code so I don't care about warnings from it.

nothings:
Sure, compiler warnings are a real problem, but enforcing a style rule may not be the appropriate solution to it.

Compilers *have* to let you compile away const at least some of the time, so we just need to find out what is best practice for that and follow that. So, what compiler is producing warnings, and what exactly is the warning?

nothings:
Ok, I looked at the dear imgui report, I see it's gcc's -Wcast-qual.

As far as I can find on the internet, there's no good way to work around this except to explicitly disable the warning:

   http://bytes.com/topic/c/answers/706939-casting-away-constness
   http://stackoverflow.com/questions/13249756/explicit-ignore-warning-from-wcast-qual-cast-discards-attribute-const

Doing that w/o disabling it elsewhere requires "#pragma GCC diagnostic push", which requires a minimum of GCC 4.6. (Really? GCC has no way of dealing with this properly pre-4.6???)

Probably the best solution is to take all of the problematic API functions and rename them to be internal only functions, then wrap them with new one-line public functions that just cast-away and call the internal functions. Then can use a single properly-#ifdef'd #pragma wrapping around that whole block of functions.

samhocevar:
I am compiling [`imgui`](https://github.com/ocornut/imgui) using GCC 6.2 (though clang 3.8.1 also qualifies). The relevant compilation flag is `-Wcast-qual` and the emitted warnings related to `stb` can be found here: http://paste.debian.net/887369/

nothings:
@samhocevar If my description above makes sense, I can take a shot at doing it and emailing you the files for you to test. Email me if that seems worthwhile.

samhocevar:
Thanks, I’ll happily test any changes (my address is sam@hocevar.net).

But I was under the impression that the “best” solution (in terms of effort) would be to just change the few places where a `foo const *` is cast to a `foo *`; is there a real problem with that approach that I am overlooking?

nothings:
You've only tackled three libraries. I assume in many other libraries one would have to propagate the const many levels down, and it would be a significant change there. That is, the real problem is that it's imposing a style change on the libraries--you're proposing that henceforth they be written everywhere const-correct.

Doing what I describe makes it easier for future maintainence work to stay const-incorrect, and makes it easy to add new functions that have const-correct API.

However, your fixes to the qsort-comparison functions though are correct as-is, since they're not part of the public API, they're part of interacting with the C stdlib API.

nothings:
Revised way to fix this was validated by email, checked-in now.

PR #370, opened 2016-11-26
stb_dxt: 3Dc support

snake5:
As I wanted to compress some tangent space normal maps, I read some articles about the topic and found out that the 3Dc format is very good, and it is actually two DXT5 alpha blocks added together.

The commit shows how easy it would be to support 3Dc here. And that's really about it, since I wasn't sure what to do with the unused 'mode' variable as well as the padding bytes that 'src' requires (they didn't matter for me but they might for someone else).

Any chance you're interested in supporting 3Dc compression explicitly in this library?

### DIFF
--- a/stb_dxt.h
+++ b/stb_dxt.h
@@ -32,6 +32,7 @@
 #define STB_DXT_HIGHQUAL  2   // high quality mode, does two refinement steps instead of 1. ~30-40% slower.
 
 void stb_compress_dxt_block(unsigned char *dest, const unsigned char *src, int alpha, int mode);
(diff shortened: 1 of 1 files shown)


nothings:
merged in next version, I got rid of the padding bytes and mode and renamed to BC5 rather than 3DC; however, it's untested so please test!

snake5:
Two things:

* function declaration did not match the definition
* there were a few more changes needed to stb__CompressAlphaBlock

Patch: https://gist.github.com/snake5/2baf14237c7c20f5ad8a1e3d91106f03

nothings:
fix will go up in a minute, thanks!

PR #376, opened 2016-12-05
Number of fonts in a TrueType file

IntellectualKitty:
This PR adds the function stbtt_GetNumberOfFonts to determine the number of fonts in a .ttc or .ttf file.  It uses the same basic code as asstbtt_GetFontOffsetForIndex, as well as following the structure of encapsulating internal methods from the API (stbtt_GetNumberOfFonts calls stbtt_GetNumberOfFonts_internal).

Note that the code to check for Apple TrueType fonts from PR #375 is NOT included in this PR.

### DIFF
--- a/stb_truetype.h
+++ b/stb_truetype.h
@@ -95,7 +95,8 @@
 //
 //   "Load" a font file from a memory buffer (you have to keep the buffer loaded)
 //           stbtt_InitFont()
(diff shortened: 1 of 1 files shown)


nothings:
FWIW I never added this originally because you can just repeatedly call stbtt_GetFontOffsetForIndex until it fails, but I'm going ahead and adding it anyway.

PR #379, opened 2016-12-05
The Si prefixes for mega, giga and tera are upper case

jarikomppa:
(lower case m is not mega but milli)

More proper fix would probably make these "Kibibytes" etc, as they're 1024 and not 1000 divisors, but this is pretty harmless change.

### DIFF
--- a/stb_sprintf.h
+++ b/stb_sprintf.h
@@ -555,7 +555,7 @@ STBSP__PUBLICDEF int STB_SPRINTF_DECORATE( vsprintfcb )( STBSP_SPRINTFCB * callb
         pr = 0;
         
         // handle k,m,g,t
-        if (fl&STBSP__METRIC_SUFFIX) { tail[0]=1; tail[1]=' '; { if (fl>>24) { tail[2]="_kmgt"[fl>>24]; tail[0]=2; } } };
+        if (fl&STBSP__METRIC_SUFFIX) { tail[0]=1; tail[1]=' '; { if (fl>>24) { tail[2]="_kMGT"[fl>>24]; tail[0]=2; } } };
 
         flt_lead:
         // get the length that we copied
--- a/stb_sprintf.h
+++ b/stb_sprintf.h
@@ -133,6 +133,9 @@ PERFORMANCE vs MSVC 2008 32-/64-bit (GCC is even slower than MSVC):
 
 #include <stdarg.h>  // for va_list()
 
+// Uncomment for old-school prefixes (KB, MB instead of KiB, MiB) with the $ format
+//#define STBSP__JEDEC_PREFIX
+
 #ifndef STB_SPRINTF_MIN
 #define STB_SPRINTF_MIN 512 // how many characters per callback
 #endif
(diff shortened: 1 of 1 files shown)


jarikomppa:
Okay, added the Ki stuff to the patch.

jarikomppa:
Sure, but the power-of-1000 would be a different prefix.

nothings:
Ok, I didn't really explain my vague thoughts, but the point is it presumably shouldn't be on a #define. I guess there are three cases (1000-as-K, 1024-as-Ki, 1024-as-K) overall, so the question is whether it makes sense to lock the last two to a #define or not.

jarikomppa:
Yeah, better have it configurable on the fly; the change I did was minimal
change that should be safe to integrate. I'll look into how the prefixes
are parsed to see if there's a simple way to add more configurability (and
if not, how to add more prefixes).

jarikomppa:
Would this work:
$ - SI 1000->k
$$ - SI 1024->Ki
$$$ - JEDEC 1024->K

I'd also like the space between numbers and the SI prefix to be optional,
but that's a different dimension (and having more than three $'s in a
format would be overly messy)

nothings:
Is $ a modifier or is it a terminal character of the %-sequence? If the latter, it would be bad to have $/$$/$$$, since then there's no way to print a bare '$' after the sequence if the type is $ or $$.

jarikomppa:
A bit more involved change. I tried to follow the code style somewhat, but probably failed miserably. Anyway, the results look like this:

    stbsp_sprintf(temp, "%10d %$d %$$d %$$$d  %_$d %_$_$d %$$$_d",v,v,v,v,v,v,v);

    2147483647 2.1 G 2.0 Gi 2.0 G  2.1G 2.0Gi 2.0G
    1073741823 1.1 G 1024.0 Mi 1024.0 M  1.1G 1024.0Mi 1024.0M
     536870911 536.9 M 512.0 Mi 512.0 M  536.9M 512.0Mi 512.0M
     268435455 268.4 M 256.0 Mi 256.0 M  268.4M 256.0Mi 256.0M
     134217727 134.2 M 128.0 Mi 128.0 M  134.2M 128.0Mi 128.0M
      67108863 67.1 M 64.0 Mi 64.0 M  67.1M 64.0Mi 64.0M
      33554431 33.6 M 32.0 Mi 32.0 M  33.6M 32.0Mi 32.0M
      16777215 16.8 M 16.0 Mi 16.0 M  16.8M 16.0Mi 16.0M
       8388607 8.4 M 8.0 Mi 8.0 M  8.4M 8.0Mi 8.0M
       4194303 4.2 M 4.0 Mi 4.0 M  4.2M 4.0Mi 4.0M
       2097151 2.1 M 2.0 Mi 2.0 M  2.1M 2.0Mi 2.0M
       1048575 1.0 M 1024.0 Ki 1024.0 K  1.0M 1024.0Ki 1024.0K
        524287 524.3 k 512.0 Ki 512.0 K  524.3k 512.0Ki 512.0K
        262143 262.1 k 256.0 Ki 256.0 K  262.1k 256.0Ki 256.0K
        131071 131.1 k 128.0 Ki 128.0 K  131.1k 128.0Ki 128.0K
         65535 65.5 k 64.0 Ki 64.0 K  65.5k 64.0Ki 64.0K
         32767 32.8 k 32.0 Ki 32.0 K  32.8k 32.0Ki 32.0K
         16383 16.4 k 16.0 Ki 16.0 K  16.4k 16.0Ki 16.0K
          8191 8.2 k 8.0 Ki 8.0 K  8.2k 8.0Ki 8.0K
          4095 4.1 k 4.0 Ki 4.0 K  4.1k 4.0Ki 4.0K
          2047 2.0 k 2.0 Ki 2.0 K  2.0k 2.0Ki 2.0K
          1023 1 k 1023  1023   1k 1023 1023
           511 511  511  511   511 511 511
           255 255  255  255   255 255 255

I did notice while doing this that the $-formatted stuff doesn't work with the padding parameters (so doing %5$d, for example, doesn't work).

jarikomppa:
It's a modifier.

jarikomppa:
The $ character has to be a modifier because otherwise printf wouldn't know what format the source data is in. I also added the _ modifier to get rid of the space between number and the postfix.

PR #380, opened 2016-12-05
Fix comments where rr* functions hadn't been renamed yet.

Kelimion:
Just a bit of janitorial work. The original functions seem to have been called rr* and not all comments had been updated to reflect the new names yet.


nothings:
This was fixed independently.

PR #381, opened 2016-12-05
Fix the check for a percent character

d26435:
It was matching everything less than 0x26, so could cause a minor performance loss. Also made the second if statement superfluous.


jeffatrad:
Cool - nice catch!!

PR #391, opened 2016-12-27
stb_vorbis: Rename Point to stbv__point

jlnr:
`Point` clashes with the struct of the same name in MacTypes.h (included by CoreFoundation.h on macOS/iOS), which breaks stb_vorbis.c when a precompiled header includes these files. (Admittedly, this is only an issue because of my oddball project setup...)

The stb_image libraries use the `stbi__` prefix for internal structs/functions, so I've prefixed this struct with `stbv__`.

I'm not sure that "point" is a descriptive name to begin with. The struct is only used in one place, and the comment states that it's not really a point, but repurposed as something else. Is there a better name for what it is? An "X with index"? Should I rather rename the struct instead of prefixing it?

### DIFF
--- a/stb_vorbis.c
+++ b/stb_vorbis.c
@@ -1270,12 +1270,12 @@ static void neighbors(uint16 *x, int n, int *plow, int *phigh)
 typedef struct
 {
    uint16 x,y;
-} Point;
+} stbv__point;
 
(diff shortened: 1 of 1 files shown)


nothings:
I renamed this further as you suggested.

PR #415, opened 2017-02-22
Add C++ linkage to stb_dxt

Mokosha:
Just adds the extern "C" to the header to match the other files so that we can use it with C++ compilers, too.


nothings:
added for next version, doesn't actually use the pull request for random reasons

Mokosha:
OK, thanks!

PR #424, opened 2017-03-03
stb_image: JPEG: Provide failure reason for unknown marker

jeremysawicki:
Provide a failure reason when an unknown marker is encountered in a JPEG file.

This is the only case of a missing failure reason that I encountered in the ImageNet data set, though others exist in the code (e.g. "return L==0;" in the same function).

### DIFF
--- a/stb_image.h
+++ b/stb_image.h
@@ -2917,7 +2917,7 @@ static int stbi__process_marker(stbi__jpeg *z, int m)
       stbi__skip(z->s, stbi__get16be(z->s)-2);
       return 1;
    }
-   return 0;
+   return stbi__err("unknown marker","Corrupt JPEG");
 }
 
 // after we see SOS


nothings:
Thank you for all this work!

nothings:
Do all the images from ImageNet now work, or is there still more to do?

jeremysawicki:
I still have a couple more changes to send your way:

* Support for 4-component CMYK/YCCK JPEGs.  That is the last set of ImageNet files that cannot be decoded.  The changes may or may not be suitable for inclusion though.  Proper CMYK->RGB conversion requires dealing with color profiles, etc., which gets fairly complex.  I am just merging the black and treating CMY as 1-RGB, which results in colors that look somewhat reasonable but are not exactly correct.  I don't know if you'll fall on the side of "something is better than nothing" or "nothing is better than wrong".

* A better test for RGB JPEGs that detects some more cases.  I just worked out how to do it without false positives.

PR #425, opened 2017-03-03
Fix a pair of warnings in stb_image.h

GrayHatter:
We're using stbi in https://github.com/uTox/uTox and are working towards `-Werror`, we're willing to disable the indentation warnings, but can't disable these two.

Also, the bit shift was UB, and according to the internet some compilers will and DO optimize out that line.


GrayHatter:
Ignoring the white space changes makes this easier to parse https://github.com/nothings/stb/pull/425/files?w=1 

One line of tab replacement, the rest are spaces at EOL

rygorous:
As-is, the "(void) ri;" breaks C98 compilation.

Can you split whitespace changes into a separate commit?

Maybe change the magnitude update to be just "-= (1<<magbits) - 1" and
sidestep both signed left shift/twos complement issues entirely.

nothings:
FWIW, I'm fine with the PR as is.

PR #427, opened 2017-03-04
stb_image: JPEG: Support 4-channel CMYK/YCCK images

jeremysawicki:
The ImageNet dataset contains 320 4-channel JPEGs:

* 247 with an Adobe APP14 segment that specifies color transform 0x02, meaning YCCK.
* 71 with an Adobe APP14 segment that specifies color transform 0x00, meaning CMYK.  These also use component IDs 'C', 'M', 'Y', 'K'.
* 2 with no Adobe APP14 segment.

This fix uses the APP14 segment to distinguish the first two cases, ignoring the component IDs.

The final two images contain only 0xff bytes in the fourth channel, so it is hard to tell whether that channel represents black or alpha or something else.  One of the images needs to be decoded like a normal YCbCr JPEG to look correct.  The other is an abstract geometric shape, so it is hard to tell what the correct colors are.  My best guess is that they are some form of YCbCr + alpha.  This fix decodes them as YCbCr and ignores the fourth channel.

This fix does not do proper CMYK->RGB conversion with color profiles, etc., so while the resulting colors look somewhat reasonable they are not exactly correct.

Some changes overlap with RGB detection (#426) -- sorry for any conflicts.

### DIFF
--- a/stb_image.h
+++ b/stb_image.h
@@ -1691,6 +1691,7 @@ typedef struct
    int            succ_high;
    int            succ_low;
    int            eob_run;
+   int            app14;
    int            rgb;
 
    int scan_n, order[4];
@@ -2625,7 +2626,7 @@ static void stbi__jpeg_reset(stbi__jpeg *j)
    j->code_bits = 0;
    j->code_buffer = 0;
    j->nomore = 0;
-   j->img_comp[0].dc_pred = j->img_comp[1].dc_pred = j->img_comp[2].dc_pred = 0;
+   j->img_comp[0].dc_pred = j->img_comp[1].dc_pred = j->img_comp[2].dc_pred = j->img_comp[3].dc_pred = 0;
    j->marker = STBI__MARKER_none;
    j->todo = j->restart_interval ? j->restart_interval : 0x7fffffff;
    j->eob_run = 0;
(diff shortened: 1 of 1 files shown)


jeremysawicki:
This is the last change to handle all images in the ImageNet dataset.

(Actually, the dataset includes 26 gzip-compressed JPEGs for some reason, but that seems outside the scope of stb_image.)

nothings:
FWIW, I replaced the (a*b+128)/255 with this code (from an algorithm by Jim Blinn):
```
// fast 0..255 * 0..255 => 0..255 rounded multiplication
static stbi_uc stbi__blinn_8x8(stbi_uc x, stbi_uc y)
{
   unsigned int t = x*y + 128;
   return (stbi_uc) ((t + (t >>8)) >> 8);
}
```

nothings:
Also, I did have to manually fix a bunch of merge errors between this and rgb, so you might want to retest to make sure I didn't inadvertently break anything.

nothings:
merged

jeremysawicki:
I verified that all ImageNet images still decode, and the 4-channel and RGB images still look right.  Thanks for incorporating all these changes.

Also, nice improvement for the division by 255.

nothings:
Thanks for testing and for your fixes!

PR #432, opened 2017-03-11
stb_image_write.h: JPEG writing support based on jo_jpeg.cpp

DanielGibson:
I ported Jon Olick's awesome public domain jo_jpeg.cpp to plain C and to use the stbiw__write_context abstraction instead of plain stdio.
Tested on 64bit Linux with GCC and clang and WinXP with MSVC6.

I noticed that some of the stbi_write_* function definitions were missing STBIWDEF, so I added that.

### DIFF
--- a/stb_image_write.h
+++ b/stb_image_write.h
@@ -1283,7 +1283,8 @@ static int stbi_write_jpg_core(stbi__write_context *s, int width, int height, in
    {
       static const unsigned char head0[] = { 0xFF,0xD8,0xFF,0xE0,0,0x10,'J','F','I','F',0,1,1,0,0,1,0,1,0,0,0xFF,0xDB,0,0x84,0 };
       static const unsigned char head2[] = { 0xFF,0xDA,0,0xC,3,1,0,2,0x11,3,0x11,0,0x3F,0 };
-      const unsigned char head1[] = { 0xFF,0xC0,0,0x11,8,height>>8,height&0xFF,width>>8,width&0xFF,3,1,0x11,0,2,0x11,1,3,0x11,1,0xFF,0xC4,0x01,0xA2,0 };
+      const unsigned char head1[] = { 0xFF,0xC0,0,0x11,8,(unsigned char)(height>>8),STBIW_UCHAR(height),(unsigned char)(width>>8),STBIW_UCHAR(width),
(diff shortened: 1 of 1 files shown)


nothings:
I generally prefer not to turn other people's single-file public domain libraries into stb libraries. I realize that C++ => C is useful; maybe you should just make a repo for a 'jo_jpeg.c' (or 'jo_jpeg.h')?

DanielGibson:
C++ => C is one useful thing, but I think supporting stbi_write_func is pretty neat.
And of course having one more supported file type in stb_image_write.
I don't see how (from the original authors point of view) me providing jp_jpeg.h is that different from integrating it into stb_image_write.h

Maybe I should have first asked Jon what he thinks about it..

stb_image_write.h
```diff
@@ -1013,6 +1029,350 @@ STBIWDEF int stbi_write_png_to_func(stbi_write_func *func, void *context, int x,
...
+         fdtbl_Y[k]  = 1 / (YTable [stbiw__jpg_ZigZag[k]] * aasf[row] * aasf[col]);
+         fdtbl_UV[k] = 1 / (UVTable[stbiw__jpg_ZigZag[k]] * aasf[row] * aasf[col]);
+      }
+   }
+
+   // Write Headers
+   {
+      static const unsigned char head0[] = { 0xFF,0xD8,0xFF,0xE0,0,0x10,'J','F','I','F',0,1,1,0,0,1,0,1,0,0,0xFF,0xDB,0,0x84,0 };
+      static const unsigned char head2[] = { 0xFF,0xDA,0,0xC,3,1,0,2,0x11,3,0x11,0,0x3F,0 };
+      const unsigned char head1[] = { 0xFF,0xC0,0,0x11,8,height>>8,height&0xFF,width>>8,width&0xFF,3,1,0x11,0,2,0x11,1,3,0x11,1,0xFF,0xC4,0x01,0xA2,0 };
```
kieselnb: This line causes errors when compiling with clang and -Wc++11-narrowing. Adding casts makes it happy, e.g.
`(unsigned char)(height>>8)`

nothings:
Given Olick's retweet, yeah, I will integrate this, just need to "get around to it™". Thanks for the fixes.

DanielGibson:
Thanks for pointing this out, @kieselnb, just pushed a fix for this (works for me, tested with clang++ 3.8 on Linux amd64)

Sounds good @nothings, thanks! :)

Issue #417, opened 2017-02-24
A strange compilation warning in stb_image_resize.h


Flix01:
Hi.
Usually I don't care about warnings at all... but when I compile using the emscripten compiler, it seems to emphasize on some of them.
What I'd just like to know is only if this is the wanted behaviour or not:

```
stb_image_resize.h:1819:186: warning: 
      integer literal is too large to be represented in type 'long' and is
      subject to undefined behavior under C++98, interpreting as
      'unsigned long'; this literal will have type 'long long' in C++11 onwards
      [-Wc++11-compat]
  ...+ alpha_channel])) * 4294967295);
                                ^
```

nothings:
It appears it's immediately converted to double anyway, so it should probably just be a floating point constant. Does that sound right @BSVino ?

BSVino:
Think so, should be 4294967295.0. Want me to make a patch?

nothings:
Sure, thanks.

Flix01:
I can confirm that by replacing 4294967295 with 4294967295.0 the warning disappears.
Thanks!

BSVino:
I'm about to leave for GDC, I'll do the patch when I get back. Sorry about that.

PR #436, opened 2017-03-15
Fix some STBIR warnings

BSVino:
Apple clang -Wall and -Wextra warnings are all fixed, and all of the MSVC /W4 warnings worth fixing are fixed. Ran the test cases, everything OK.


BSVino:
Resolves https://github.com/nothings/stb/issues/417

PR #437, opened 2017-03-18
Added zeroing of the out after it is allocated in the function stbi__create_png_image_raw

rds1983:
As that `out` contents are used in the encoding.
I.e. in that line:
`case STBI__F_up         : cur[k] = STBI__BYTECAST(raw[k] + prior[k]); break;`
In that code `prior` points to `out`.

**p.s.** Thanks for the awesome library. I work on porting it to C#: https://github.com/rds1983/StbSharp

### DIFF
--- a/stb_image.h
+++ b/stb_image.h
@@ -4190,6 +4190,7 @@ static int stbi__create_png_image_raw(stbi__png *a, stbi_uc *raw, stbi__uint32 r
    STBI_ASSERT(out_n == s->img_n || out_n == s->img_n+1);
    a->out = (stbi_uc *) stbi__malloc_mad3(x, y, output_bytes, 0); // extra bytes to write off the end into
    if (!a->out) return stbi__err("outofmem", "Out of memory");
+   memset(a->out, 0, x * y * output_bytes);  
 
    img_width_bytes = (((img_n * x * depth) + 7) >> 3);
    img_len = (img_width_bytes + 1) * y;


nothings:
Have you seen this go wrong, or are you just guessing it is wrong based on trying to port it?

Because this shouldn't affect the code for two reasons:

1. 'prior' points to out 'higher up' in the image, which has always been decoded already, or it points *outside* the image (when cur is on the first line) which is random memory even after adding the memset.

2. When cur is on the first line the code doesn't actually even use 'prior' because of the use of first_row_filter[] means none of the paths that sample prior are used.

rds1983:
I've noticed that it produces different result depending on what 'out' is filled with.
I am attaching a small project that confirms it. It loads the same image(stored at byte array) twice: first time when 'out' is filled with 255s, second - with zeroes. Then it compares amount of different bytes. It gives 1756 different bytes on my end.
[StbTest.zip](https://github.com/nothings/stb/files/852616/StbTest.zip)

nothings:
That file isn't decoded correctly with your fix (and even if it were, memsetting the entire image to 0 is an unnecessary slowdown on very large images).

The correct fix is in stbi__create_png_image_raw, it is necessary inside the j<y loop to recompute 'prior = cur - stride' in the 'depth < 8' branch after the code updates 'cur'.

This will be fixed in the next release of stb_image. Thanks for reporting the bug!

nothings:
Also, just because you're going to run into this eventually--you're using the API wrong, if you request 4 channels per pixel (STBI_rgb_alpha), the image that comes back to you has 4 channels per pixel, but 'comp' returns you NOT how many channels were returned to you, but how many were in the file, so your test code looking for differences is only looping over x*y*3 instead of x*y*4 of the bytes. (This is documented but not well enough, a lot of people make this mistake. The API has this unintuitive behavior for historical reasons.)

rds1983:
Thanks for the help. I've tried that fix(recomputing of 'prior' inside 'depth<8') and now there are no difference in results with different filler. Even after changing 'comp' to 4.

Issue #393, opened 2016-12-28
stb_printf.h triggers complaints from clang's Address Sanitiser


tim-rex:
$ cat test.c
```
#include <stdio.h>

#define STB_SPRINTF_IMPLEMENTATION
#include "../STB/stb/stb_sprintf.h"

int main (int argc, char *argv[])
{
    const unsigned int LEN = 1024;
    char str[LEN];

    printf ("Hello\n");
    stbsp_snprintf (str, LEN, "World?");
    printf ("%s\n", str);
}
```

$ clang test.c -fsanitize=address
$ ./a.out
```
Hello
=================================================================
==28032==ERROR: AddressSanitizer: global-buffer-overflow on address 0x000102ed3584 at pc 0x000102ec10d0 bp 0x7fff5cd3dbd0 sp 0x7fff5cd3dbc8
READ of size 4 at 0x000102ed3584 thread T0
    #0 0x102ec10cf in stbsp_vsprintfcb (a.out+0x1000010cf)
    #1 0x102ecec66 in stbsp_vsnprintf (a.out+0x10000ec66)
    #2 0x102ecf50a in stbsp_snprintf (a.out+0x10000f50a)
    #3 0x102ecf709 in main (a.out+0x10000f709)
    #4 0x7fffe5f9e254 in start (libdyld.dylib+0x5254)

0x000102ed3584 is located 60 bytes to the left of global variable '<string literal>' defined in 'test.c:14:13' (0x102ed35c0) of size 4
  '<string literal>' is ascii string '%s
'
0x000102ed3587 is located 0 bytes to the right of global variable '<string literal>' defined in 'test.c:13:31' (0x102ed3580) of size 7
  '<string literal>' is ascii string 'World?'
SUMMARY: AddressSanitizer: global-buffer-overflow (a.out+0x1000010cf) in stbsp_vsprintfcb
Shadow bytes around the buggy address:
  0x1000205da660: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x1000205da670: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x1000205da680: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x1000205da690: 00 00 00 00 00 00 00 00 05 f9 f9 f9 f9 f9 f9 f9
  0x1000205da6a0: 06 f9 f9 f9 f9 f9 f9 f9 07 f9 f9 f9 f9 f9 f9 f9
[...]

PR #442, opened 2017-04-18
stb_sprintf: Improve ASan workaround

h-s-c:
Adds support for more versions of clang and gcc.

### DIFF
--- a/stb_sprintf.h
+++ b/stb_sprintf.h
@@ -9,6 +9,7 @@
 //    github:d26435
 //    github:trex78
 //    Jari Komppa (SI suffixes)
+//    Kevin Schmidt
 //
 // LICENSE:
 //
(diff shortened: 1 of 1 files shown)


rygorous:
This seems super-dodgy to me.

I think we should *fix* ASAN-reported issues, not add ever more boilerplate to suppress them.

h-s-c:
According to the original ASan issue thread this should be ok.
[https://github.com/nothings/stb/issues/393](url)

> Multiple places in the code will trigger this. It's because it relies on assumptions on how literal and heap allocated strings are aligned in memory and read them in 4-byte blocks. It's almost always safe to do so but it's technically undefined behaviour to read past the string.

nothings:
Had to merge this manually because I didn't want the Makefile or test_sprintf.c changes. Hope I didn't screw anything up!

PR #451, opened 2017-05-04
STBIDEF STBIRDEF STBIWDEF inline

beru:
Hi, this PR staticinlines `stb_image`, `stb_image_resize`, `stb_image_write`.

### DIFF
--- a/stb_image.h
+++ b/stb_image.h
@@ -325,7 +325,9 @@ typedef unsigned short stbi_us;
 extern "C" {
 #endif
 
-#ifdef STB_IMAGE_STATIC
+#if defined(STB_IMAGE_STATIC) && defined(STB_IMAGE_INLINE)
+#define STBIDEF static inline
+#elif defined(STB_IMAGE_STATIC)
 #define STBIDEF static
 #else
 #define STBIDEF extern
--- a/stb_image.h
+++ b/stb_image.h
@@ -325,10 +325,16 @@ typedef unsigned short stbi_us;
 extern "C" {
 #endif
 
-#if defined(STB_IMAGE_STATIC) && defined(STB_IMAGE_INLINE)
-#define STBIDEF static inline
-#elif defined(STB_IMAGE_STATIC)
+#ifdef STB_IMAGE_STATIC
 #define STBIDEF static
+#elif defined(STB_IMAGE_INLINE)
+#ifdef __cplusplus
+#define STBIDEF inline
+#elif defined(__STDC_VERSION__) && __STDC_VERSION__ >= 199901L
+#define STBIDEF inline
+#else
+#error "inline not supported."
+#endif
 #else
 #define STBIDEF extern
 #endif
--- a/stb_image_resize.h
+++ b/stb_image_resize.h
@@ -191,16 +191,22 @@ typedef uint16_t stbir_uint16;
 typedef uint32_t stbir_uint32;
 #endif
 
-#if defined(STB_IMAGE_RESIZE_STATIC) && defined(STB_IMAGE_RESIZE_INLINE)
-#define STBIRDEF static inline
-#elif defined(STB_IMAGE_RESIZE_STATIC)
-#define STBIRDEF static
-#else
 #ifdef __cplusplus
-#define STBIRDEF extern "C"
+extern "C" {
+#endif
+
+#ifdef STB_IMAGE_RESIZE_STATIC
+#define STBIRDEF static
+#elif defined(STB_IMAGE_RESIZE_INLINE)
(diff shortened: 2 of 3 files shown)


nothings:
1. Inline isn't standard in C before C99, so you can't just turn it on for STB_IMAGE_STATIC etc. (You could version test to see if it's C99 though.)

2. What is the benefit to doing it? For example, in stb_image, the only functions annoted with STBIDEF are the *API* functions, which generally aren't doing a lot, but might be called from multiple places in user code, needlessly expanding the generated code without seeing any performance benefit.

beru:
> 1. Inline isn't standard in C before C99, so you can't just turn it on for STB_IMAGE_STATIC etc. (You could version test to see if it's C99 though.)

Please check the code again, I tweaked preprocessor macros based on your comment.
Maybe a bit ugly....

Some old compilers support inlining with `__inline` or `__inline__`, but those are not supported with this PR. I wish 20th century ghosts won't haunt me...

> 2. What is the benefit to doing it? For example, in stb_image, the only functions annoted with STBIDEF are the API functions, which generally aren't doing a lot, but might be called from multiple places in user code, needlessly expanding the generated code without seeing any performance benefit.

By using `inline` keywords, I can remove `defined but not used [-Wunused-function]` warnings when I use stb library from header-only C++ library.

https://github.com/tiny-dnn/tiny-dnn/pull/685 is related.

nothings:
>By using inline keywords, I can remove defined but not used [-Wunused-function] warnings when I use >stb library from header-only C++ library.

Makes sense. But what about all the functions that aren't STBIDEF? They're not warned about being unused since they're referenced by a static inline function? This feels like a hack that may not work for all time, I suppose there's no better way to turn off the warning locally to that block of code?

beru:
I don't know much about gcc but according to this page.
https://gcc.gnu.org/onlinedocs/gcc-3.4.3/gcc/Warning-Options.html
> -Wunused-function
Warn whenever a static function is declared but not defined or a non\-inline static function is unused. 

So using inline keyword might be the only way to take a hike?
Or stop using `-Wall` or `-Wunused-function` option is less-distressful but people are spartan to hunt bugs.

beru:
> Makes sense. But what about all the functions that aren't STBIDEF? They're not warned about being unused since they're referenced by a static inline function? This feels like a hack that may not work for all time, I suppose there's no better way to turn off the warning locally to that block of code?

Actually, I haven't deeply tried to find other ways to turn off the warnings.

```cpp
#define STBIDEF inline
#undef STB_IMAGE_STATIC
#include "stb/stb_image.h"
```
This might work...

Kerndog73:
Would declaring every function as `STBI_DEF` and then letting the client set `STBI_DEF` completely break everything?

I'm using stb_image in a header-only C++ library. I had to manually declare every function as inline to get it to work but I would rather the preprocessor do that for me.

    // define functions as inline
    #define STBIDEF inline
    // define globals as static
    #define STB_IMAGE_STATIC
    #include "stb_image.h"

nothings:
Letting the client define a STBIDEFPRIVATE or some such might be the best way around this, yeah.

nothings:
In the next release, STB_IMAGE_WRITE supports setting STBIWDEF before you include it. Let me know if this is a workable solution and I'll do it to the others.

Kerndog73:
It looks like defining `STBIWDEF` actually isn't necessary. It would seem that I have forgotten the meaning of `static`. And I call myself a programmer... 

The reason I was getting linker errors the first time around was because **some** of the private functions in image_write aren't `static`. So I just thought, "oh, duplicate symbol, EVERYTHING MUST BE INLINE". When in fact, all I had to do was define `STB_IMAGE_WRITE_STATIC`.

These private functions in stb_image_write are not `static`

- `stbi_zlib_compress`
- `stbi_write_png_to_mem`
- `stbiw__write_run_data`
- `stbiw__write_dump_data`
- `stbiw__write_hdr_scanline`
- `stbiw__linear_to_rgb`

There could be more in the other libraries. It's an easy mistake to make. All you have to do is put this

    #define STB_XXX_STATIC
    #define STB_XXX_IMPLEMENTATION
    #include "stb_xxx.h"

in two separate C files and then link them together and you'll find the rest.

I also just downloaded the latest version of stb_image and found an occurrence of `STBI_REALLOC` being used instead of `STBI_REALLOC_SIZED`. Luckily, there's only two and they're right next to each other. Inside `stbi__load_gif_main` at lines 6526 and 6528.

nothings:
added a compile test and fixed all the non-static functions in stb_image, stb_image_write, and stb_truetype. for next release

PR #453, opened 2017-05-04
Fix scope for stbi_load(), stbi_load_16() & stbi_loadf() when STBI_NO_STDIO

poppolopoppo:
The 'filename' variant of these 3 functions were declared but not defined.


rygorous:
Duplicate of #440. I've merged both #440 and your earlier #452 into the dev branch; closing.

PR #456, opened 2017-05-22
Add I/O callback support to stb_vorbis

nicebyte:
This change adds support for generic I/O callbacks to stb_vorbis.

This serves the same purpose as the pushdata api - to be able to feed bytes from a custom source into the library. However, I found that pushdata was less suitable for my particular use case and felt that a file-like abstraction was a better approach.

Note: I deliberately did not reimplement stdio-based functionality in terms of the new API. My assumption is that users should be able to use stdio-based functionality without relying on the callbacks API whatsoever. If you think otherwise, please let me know.

Testing: I tested decoding a short OGG file, with the following combinations of #defines:
  - (no additional defines)
  - STB_VORBIS_NO_STDIO only
  - STB_VORBIS_NO_IOCALLBACKS only

### DIFF
--- a/stb_vorbis.c
+++ b/stb_vorbis.c
@@ -192,7 +192,7 @@ typedef struct
   // occurred.
   int error;
 } stb_vorbis_iocallbacks;
-#endif
+#endif // STB_VORBIS_NO_IOCALLBACKS
 
 // get general information about the file
 extern stb_vorbis_info stb_vorbis_get_info(stb_vorbis *f);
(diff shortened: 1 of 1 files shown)


rygorous:
This is effectively a feature request (I added a new issue #478 for it), but I'm vetoing this particular implementation.

My specific concerns are this:
* The API surface for the installable IO callbacks is way too large. Having both `read` and a more specific `getcharacter` is redundant and means extra work for everyone who wants to use the IO callbacks. Similarly, configurable values for `EOF`, `SEEK_SET` etc. add, not remove, complexity on both sides.
* Having a `getcharacter` callback is a really bad idea. `getc` and `fgetc` are traditionally implemented as macros that directly manipulate the internal `FILE` buffer pointers, and stb_vorbis implicitly relies on `get8` being inlined; actually doing a callback for every single byte read is really expensive and to be avoided.
* Adding yet more #ifdefs (and also ifs at run-time when the installable IO callback support isn't disabled) makes what is already code that's messier than it should be even worse.

If we want all three of "regular stdio", "decode from memory" and "IO callbacks", the best way to do that is to map all three to a built-in buffered IO layer - I think that's the direction such a patch should go, and I'll look into it.

rpvelloso:
@rygorous, using callbacks the user can do the caching as he/she sees fit. IMHO caching is not a decoder responsibility. Anyway, caching is just another layer on top of this PR, I think this is a step further in the right direction.

PR #459, opened 2017-05-30
Fixed 'printf' : unknown type field character

corporateshark:


rygorous:
Can you be more precise about what issue is this "fixing"?

As-is, this breaks the code on every Visual C++ version before VS2015 (because they don't support %zd), so it would need a platform #ifdef at the very least.

corporateshark:
It gives a warning on VS 2017.

rygorous:
Alright, fix that should work with other VC++ vers is checked in with commit 501812f307cd4b7609a1c6226f26bb7780cf08b8, will make it to master whenever we next cut a release.

PR #460, opened 2017-05-30
Disabled Clang pedantic warning (equality comparison with extraneous …

corporateshark:


rygorous:
I'm totally OK with fixing the code that *causes* warning,s but stb_image (and other stb libraries) really shouldn't get in the business of locally disabling compiler warnings via #pragmas. This code gets used on a bunch of exotic compiler/platform combinations and disabling warnings, especially at the more marginal (-Wextra / -pedantic etc.) levels, quickly gets out of hand.

What platform and Clang version are you getting warnings with?

corporateshark:
```
Android clang version 5.0.300080  (based on LLVM 5.0.300080)
Target: x86_64-unknown-linux
```

The code that causes the warning is in the same file actually.

nothings:
Sorry, I should have said earlier, I need to know which line(s) are actually producing warnings so they can be fixed manually, not using #pragmas.

nothings:
Closing this since it's been a year without a response.

PR #462, opened 2017-06-07
stb_image optimise vertical flip

yonni12:
A simple optimisation to have the current "vertically_flip_on_load" work by using memcpy for each row, rather than each channel value individually for each pixel.


nothings:
1. stb_image doesn't allow declare-anywhere in interests of working on pre-C99 compilers (you can see the failure from the Travis CI in the thread)

2. How about avoiding malloc some of the time? E.g. declare a 2KB or 4KB buffer on the stack, and then malloc if that size isn't large enough. (This is why it wasn't written with memcpy in the first place and was a @TODO.)

yonni12:
Made those changes

PR #463, opened 2017-06-07
adds OS/2 table metrics to stb_truetype

ornitorrincos:
Adds what in theory are only microsoft specific OS/2 table and gets typoascend/descend/linegap, which are different form the ones found on the hhea table

### DIFF
--- a/stb_truetype.h
+++ b/stb_truetype.h
@@ -110,6 +110,7 @@
 //   Character advance/positioning
 //           stbtt_GetCodepointHMetrics()
 //           stbtt_GetFontVMetrics()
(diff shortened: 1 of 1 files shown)


nothings:
Does the updated commit mean you hadn't actually tested the first version?

ornitorrincos:
I had a bug that I didn't actually catch, second version is tested and being used, sorry

rygorous:
I implemented the equivalent in commit 30c7c6b5832fca322319b04a55eb7a6181683f5f (but with less code), renamed the function to `GetFontVMetricsOS2`, and added you to the authors list. And yes, I tested it. :)

Issue #113, opened 2015-04-21
stb_image and stb_image_write: allow to use zlib for png compression


guillaumechereau:
I noticed that the zlib compression in stb_image_write was rather poor compared to what the actual zlib library would do.  For example saving an rgba image filled with the gray color (128,128,128,255) prodices a png of size 42KB.  When I use zlib to compress the data instead of stb algorithm, I get 6.4KB.

Many projects already use zlib, so I think having the option to use it would be nice, if that is not too much work.

nothings:
This is documented. Use libpng?

nothings:
Although, note to self: worth investigating why that doesn't compress smaller.

guillaumechereau:
libpng works, but I still prefer stb_image_write interface, plus the fact that it is a single file.
Isn't the compression problem due to the fact that you use a fixed huffman tree?

nothings:
If you hook it up to zlib it's no longer a single file.

I guess my stance is that if you want something to ship to customers, you might as well use libpng; stb_image_write is intended for quick-and-dirty stuff where you can always run pngcrush etc if you really care.

The fixed tree definitely limits the effective compression, but a basically-constant file should probably compress better anyway, so I wonder if the match processing isn't right.

r-lyeh:
@guillaumechereau, lodepng is somewhere between stb_image_write and libpng. you might give it a run as well.

jarikomppa:
miniz appears to be a small, unlicensed zlib implementation - https://code.google.com/p/miniz/ - could possibly be used as an alternative?

guillaumechereau:
On the platforms I am mostly interested in (android), zlib is always present, so it is basically free to use it.  I thought adding a test macro like STB_HAS_ZLIB to the code could have been a solution.  But I agree with @nothings that if the stb implementation has a problem, it would be better to fix it first.

jarikomppa:
I meant, if a better zlib implementation is wanted in stb_image_write and the desire is to keep it in one file, that might be one shortcut..

DanielGibson:
miniz actually contains a png compression function that is easy to use, see https://code.google.com/p/miniz/source/browse/trunk/miniz.c#822  
Maybe just use that instead of stb_image_write?

(Of course it would be great if stb_image_write's png compression was closer to libpng, but then the optimization should be done in the lib itself and not by using other dependencies - especially if the dependency can compress png images itself anyway)

nothings:
[...]

PR #470, opened 2017-07-04
stb_image_write.h: PNG: allow setting zlib compr. level and custom compress function

DanielGibson:
* I think it's useful if the compression level can be configured, and it was an easy change, as the builtin compression function already has a "quality" argument.
  I just added a global variable (`int stbi_write_png_level`, analogous to `stbi_write_tga_with_rle`) for this. 
* As it has been pointed out in #113, letting stbi_write_png() use an optimized deflate implementation (like from zlib) creates smaller PNGs, closer to libpng. My second commit allows doing that.
  I think this is useful, even if it adds an (optional!) dependency, because many projects use zlib, miniz or similar anyway, but might prefer stb_image(_write) over libpng because it's easier to integrate and has a nicer API.

### DIFF
--- a/stb_image_write.h
+++ b/stb_image_write.h
@@ -73,6 +73,9 @@ USAGE:
    writer, both because it is in BGR order and because it may have padding
    at the end of the line.)
 
+   PNG allows you to set the deflate compression level by setting the global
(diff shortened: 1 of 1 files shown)


bvgastel:
This is very useful, thanks! Please merge this change.

PR #473, opened 2017-07-17
stb_image_write: Support pgm support

gasparitiago:
Initial implementation for writing 8-bit grayscale images on pgm format.


nothings:
I generally do not plan on adding new file formats to stb_image_write. Is there any particular reason you think pgm is needed?

gasparitiago:
I was working with some processing image libraries that uses this format (related to kinect data streams) and the lacking of writing the image on PGM format was an issue.

I know there is another libraries that saves in this format, like OpenCV, but for simple projects it's much more easy to include only the headers from STB than the OpenCV.

But, like you said that you don't plan to add new file formats, please close the pull request.

PR #449, opened 2017-04-28
stb_sprintf: avoid GCC 6 misleading-indentation warnings

wojdyr:
with minimal changes - two line breaks only

(I didn't add myself to the list of contributors as it's only whitespace)

FTR the warnings were:
```
stb_sprintf.h: In function ‘int stbsp_vsprintfcb(char* (*)(char*, void*, int), void*, char*, const char*, __va_list_tag*)’:
stb_sprintf.h:507:9: warning: this ‘if’ clause does not guard... [-Wmisleading-indentation]
         if ( l > (stbsp__uint32)pr ) l = pr; while ((l>1)&&(pr)&&(sn[l-1]=='0')) { --pr; --l; }
         ^~
stb_sprintf.h:507:46: note: ...this statement, but the latter is misleadingly indented as if it is guarded by the ‘if’
         if ( l > (stbsp__uint32)pr ) l = pr; while ((l>1)&&(pr)&&(sn[l-1]=='0')) { --pr; --l; }
                                              ^~~~~
stb_sprintf.h:588:11: warning: this ‘if’ clause does not guard... [-Wmisleading-indentation]
           if ((stbsp__int32)(l+n)>pr) l=pr-n; i=l; while(i) { *s++=*sn++; --i; }
           ^~
stb_sprintf.h:588:47: note: ...this statement, but the latter is misleadingly indented as if it is guarded by the ‘if’
           if ((stbsp__int32)(l+n)>pr) l=pr-n; i=l; while(i) { *s++=*sn++; --i; }
```

PR #474, opened 2017-07-22
Sprintf reformat

rygorous:
This is an attempt to make stb_sprintf at least a bit more readable.

Start with clang-format configured to produce a reasonable semblance of "normal" stb style, then manually fix up a few dodgy pieces of formatting in the result, and finally remove some *really* gratuitous gotos.

### DIFF
--- a/stb_sprintf.h
+++ b/stb_sprintf.h
@@ -559,13 +559,9 @@ STBSP__PUBLICDEF int STB_SPRINTF_DECORATE(vsprintfcb)(STBSP_SPRINTFCB *callback,
          cs = 0;
          goto scopy;
 #else
-      case 'A': // float
-         h = hexu;
-         goto hexfloat;
-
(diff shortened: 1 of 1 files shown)


nothings:
This is great but I think we should merge the other stb_sprintf pull requests first because redoing them after the fact will be harder than fixing up this PR. Seem reasonable?

rygorous:
Sure!

rygorous:
I think #449 is obsoleted by this, though.

rygorous:
Yeah, there are two outstanding pull requests; one is just formatting changes and subsumed by this one, and the other is just #defines at the top to add yet more ways to turn off Address Sanitizer (on different compilers).

The latter will be just as easy to merge after, but I think we should instead fix the code so ASan has nothing to complain about. So I'm merging this one now. :)

PR #485, opened 2017-07-28
stb_sprintf.h: Don't compare uninitialized value when using zero.

rohit-n:
Haven't tested extensively, but this silences about 800 valgrind errors when starting my game up, with no other issues.


nothings:
What's then "when using zero" mean in this context?

rohit-n:
Using zero in the additional arguments to replace %i. Example:

```
char buffer[32];
stbsp_sprintf(buffer, "test%i", 0);
```

PR #488, opened 2017-08-07
stb_image_write: allow forcing PNG filter and compression

akx:
This adds a new `stbi_write_png_to_mem_ex` function with two new parameters:

* `force_filter` being < 0 means the original behavior (i.e. figure out the best-performing filter per scanline); any other values 0 <= x <= 4 correspond to PNG filters (0 = none, 1 = sub, 2 = up, 3 = average, 4 = Paeth).
* `compression_level` being < 0 equals `compression_level` 8 (the previous value). The higher this is, the better the compression should be (though it will use more memory).

These new parameters are not (yet) exposed for the higher-level API functions, and the documented API does not change as of this commit.

For what it's worth, encoding a 1920x1080 image  with `force_filter=0` is about 45% faster than `force_filter=-1` on my particular machine.

### DIFF
--- a/stb_image_write.h
+++ b/stb_image_write.h
@@ -910,62 +910,90 @@ static unsigned char stbiw__paeth(int a, int b, int c)
    return STBIW_UCHAR(c);
 }
 
-// @OPTIMIZE: provide an option that always forces left-predict or paeth predict
-unsigned char *stbi_write_png_to_mem(unsigned char *pixels, int stride_bytes, int x, int y, int n, int *out_len)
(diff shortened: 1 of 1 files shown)


akx:
Oh, and there's one TODO in the code:

> TODO: What's the upper limit for compression_level?

I couldn't divine the semantics of the `quality` parameter for `stbi_zlib_compress`...

nothings:
I changed the API to this so it's through global variables so there's no new functions.

PR #495, opened 2017-08-24
Add basic GPOS kerning

danielmaciel:
@nothings  @dougallj I have implemented a very simple 'stbtt_GetGlyphGPOSAdvance' function which retrieves GPOS-based kerning advances. It is far from complete (Horizontal latin stuff mostly), but it works for all the GPOS-based fonts I'm currently using plus a few others.

I have added some asserts which people might hit in case their fonts require a specific unimplemented format. Let me know if you hit one of those.

I have also not implemented any binary searching, so it is a little bit slow. I'm not sure how important it is to implement binary search there, I might experiment with it at some point maybe on this or next week?

Best regards,
Daniel

### DIFF
--- a/stb_truetype.h
+++ b/stb_truetype.h
@@ -22,6 +22,7 @@
 //   Mikko Mononen: compound shape support, more cmap formats
 //   Tor Andersson: kerning, subpixel rendering
 //   Dougall Johnson: OpenType / Type 2 font handling
+//   Daniel Ribeiro Maciel: basic GPOS-based kerning
 //
 //   Misc other:
 //       Ryan Gordon
@@ -695,7 +696,7 @@ struct stbtt_fontinfo
 
    int numGlyphs;                     // number of glyphs, needed for range checking
 
-   int loca,head,glyf,hhea,hmtx,kern; // table locations as offset from start of .ttf
+   int loca,head,glyf,hhea,hmtx,kern,gpos; // table locations as offset from start of .ttf
    int index_map;                     // a cmap mapping for our chosen character encoding
    int indexToLocFormat;              // format needed to map from glyph index to glyph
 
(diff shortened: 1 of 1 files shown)


nothings:
Is GPOS kerning different from the user's point of view from the existing kerning, i.e. should this just be part of the existing function? (Regardless, it presumably shouldn't be couldn't GPOS because that implies people knowing about the file format.)

danielmaciel:
'GPOS' kerning is pretty much the same thing as 'kern' kerning. I assumed the 'KernAdvance' function related to the 'kern' info, so I created a 'GPOS' version of it. Want me to pack them together?

nothings:
Yeah, "KernAdvance" maybe should have been "KerningAdvance" to avoid that ambiguity, but too late to change now. So they should definitely be integrated so people using KernAdvance just automatically get support for it. If GPOS can also support vertical,  then at some point we'll probably need to add a new function that can support both, but there's no need to do it now if your patch doesn't do it.

danielmaciel:
Okay! Working on it.

danielmaciel:
@nothings I'm done

About the vertical stuff: I have not implemented it. Might do someday if needed.
I might also implement some GSUB (ligature and contextual) things at some point - will send you a pull request if I do.

danielmaciel:
I have also made the asserts on 'not implemented yet' formats disabled by default since it might annoy people.

One can still enable them by changing the define for STBTT_GPOS_TODO_assert(x) if wanted.

nothings:
Also, yeah, binary search is probably worth doing, since pretty much every other table accessed multiple times is binary searched. Otherwise if there's a large table, querying all the data in the table takes O(N^2) time.

danielmaciel:
Okay! Gonna work on binary searches later today, will ping when done.

nothings:
There's no hurry, it'll probably be a while before I integrate it.

danielmaciel:
Done with them binary searches. You were right - it is about 2x faster now on my scenarios.

danielmaciel:
Just fixed a bug on a specific format (posFormat 1)

danielmaciel:
I have noticed an off-by-one error on my binsearches. Sorry about that. Unfortunately I have no access to a dev machine atm, can't fix them before Monday. Will ping you when I do.

danielmaciel:
Sorry, forgot to ping. Fixed the off-by-one stuff yesterday. Cheers.

danielmaciel:
Rebased to your stbtt current.

PR #505, opened 2017-09-19
Add a function to create a copy of a stretchy buffer

crcunningham:
I needed a way to copy the buffer and didn't want to make assumptions in my code that uses the buffer about the size of the header.


nothings:
Probably this just be part of full revamp discussed in https://github.com/nothings/stb/pull/35

PR #506, opened 2017-09-21
Fix implicit-fallthrough warning

alculquicondor:
This comment matches the required regex from GCC when `-Wimplicit-fallthrough=3`, which is the default with `-Wextra`.


nothings:
fixed this by hand rather than deal with the merge conflict i'd've gotten from the name in the credits list. in next release.

alculquicondor:
Thanks. No worries.

PR #524, opened 2017-11-26
Fix warning on unused stbi__sse2_available when skipping JPEG support.

stolk:
If you use stb_image.h without the JPEG support, then the compiler will warn:

```
/home/bram/src/stb/stb_image.h:651:12: warning: unused function 'stbi__sse2_available'
      [-Wunused-function]
static int stbi__sse2_available(void)
```

This function is only needed if you use the JPEG support.

### DIFF
--- a/stb_image.h
+++ b/stb_image.h
@@ -612,20 +612,20 @@ typedef unsigned char validate_uint32[sizeof(stbi__uint32)==4 ? 1 : -1];
 #endif
 
 #if !defined(STBI_NO_SIMD) && (defined(STBI__X86_TARGET) || defined(STBI__X64_TARGET))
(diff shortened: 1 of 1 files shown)


nothings:
Check the 'files changed' tab here--most of the changes are whitespace insertion between '#' and 'define' and similar. Can you please remove those changes so I can see the meaningful changes more easily?

technosaurus:
It just wraps `stbi__sse2_available` functions in `#ifndef STBI_NO_JPEG`, but this happened when sse2 was removed for gcc, so it is a stopgap.

nothings:
Looks to me like this bug was fixed independently.

PR #528, opened 2017-12-13
Add test/CMakeLists.txt

zmatsuo:
I use Visual Stduio 2017. I try to read/convert dsp/dsw file, but I can't.
So I create CMakeLists.txt based on original dsp/dsw files for create solution and project files.
Some errors occur at compile time.
I will investigate errors by VS2017 or not.


zmatsuo:
marged #529

PR #529, opened 2017-12-13
Add functions with wchar_t to stb_image.h and stb_image_write.h

zmatsuo:
wfopen (UNICODE,wchar_t) version will use UNICODE version (CreateW) API.

- CJK support
In CJK environment wchar_t support is important.
ex. (CP932)
"C:\Documents and Settings\佐藤\デスクトップ\可能性の表.txt"
hex dump
```
00000000: 433a 5c44 6f63 756d 656e 7473 2061 6e64  C:\Documents and
00000010: 2053 6574 7469 6e67 735c 8db2 93a1 5c83   Settings\....\.
00000020: 6683 5883 4e83 6783 6283 765c 89c2 945c  f.X.N.g.b.v\...\
00000030: 90ab 82cc 955c 2e74 7874                 .....\.txt
```
filename includes path separator charactor(`'\'`). it's dangerous.

- UNC
"\\server\share\.."
"\\server@SSL\path\.."

- MAX_PATH limit
MAX_PATH -> 32767 

- see
https://msdn.microsoft.com/ja-jp/library/windows/desktop/aa365247(v=vs.85).aspx

### DIFF
--- a/stb_image.h
+++ b/stb_image.h
@@ -1115,22 +1115,21 @@ static void stbi__float_postprocess(float *result, int *x, int *y, int *comp, in
 
 #ifndef STBI_NO_STDIO
 
-static FILE *stbi__fopen(char const *filename, char const *mode)
+static FILE *stbi__fopen_rb(char const *filename)
 {
    FILE *f;
 #if defined(_MSC_VER) && _MSC_VER >= 1400
-   if (0 != fopen_s(&f, filename, mode))
+   if (0 != fopen_s(&f, filename, "rb"))
       f=0;
 #else
-   f = fopen(filename, mode);
+   f = fopen(filename, "rb");
 #endif
    return f;
 }
 
-
 STBIDEF stbi_uc *stbi_load(char const *filename, int *x, int *y, int *comp, int req_comp)
 {
-   FILE *f = stbi__fopen(filename, "rb");
+   FILE *f = stbi__fopen_rb(filename);
    unsigned char *result;
    if (!f) return stbi__errpuc("can't fopen", "Unable to open file");
    result = stbi_load_from_file(f,x,y,comp,req_comp);
@@ -1166,14 +1165,13 @@ STBIDEF stbi__uint16 *stbi_load_from_file_16(FILE *f, int *x, int *y, int *comp,
 
 STBIDEF stbi_us *stbi_load_16(char const *filename, int *x, int *y, int *comp, int req_comp)
 {
-   FILE *f = stbi__fopen(filename, "rb");
+   FILE *f = stbi__fopen_rb(filename);
    stbi__uint16 *result;
    if (!f) return (stbi_us *) stbi__errpuc("can't fopen", "Unable to open file");
    result = stbi_load_from_file_16(f,x,y,comp,req_comp);
    fclose(f);
    return result;
 }
-
 
 #endif //!STBI_NO_STDIO
 
@@ -1241,8 +1239,8 @@ STBIDEF float *stbi_loadf_from_callbacks(stbi_io_callbacks const *clbk, void *us
 #ifndef STBI_NO_STDIO
 STBIDEF float *stbi_loadf(char const *filename, int *x, int *y, int *comp, int req_comp)
 {
+   FILE *f = stbi__fopen_rb(filename);
    float *result;
-   FILE *f = stbi__fopen(filename, "rb");
    if (!f) return stbi__errpf("can't fopen", "Unable to open file");
    result = stbi_loadf_from_file(f,x,y,comp,req_comp);
    fclose(f);
@@ -1277,9 +1275,9 @@ STBIDEF int stbi_is_hdr_from_memory(stbi_uc const *buffer, int len)
 }
 
 #ifndef STBI_NO_STDIO
-STBIDEF int      stbi_is_hdr          (char const *filename)
+STBIDEF int stbi_is_hdr(char const *filename)
 {
-   FILE *f = stbi__fopen(filename, "rb");
+   FILE *f = stbi__fopen_rb(filename);
    int result=0;
    if (f) {
       result = stbi_is_hdr_from_file(f);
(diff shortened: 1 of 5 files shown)


rygorous:
Unicode support in filenames is a valid request, but I'm quite strongly opposed to stacking yet another 2x factor on top of the combinatorial explosion of functions we have. There's already too many of them.

At this point, on pretty much every other platform except Windows, regular fopen just takes UTF-8 file names. I would much prefer for us to do that on Windows as well, have the functions to keep their existing signatures and add a STBI_UTF8 (or similar name) #define that, on Windows, makes the fopen calls go through a wrapper that converts the passed-in file name from UTF8 to UTF16 and calls _wfopen. (We want to keep the existing path that keeps using code pages as the default for backwards compatibility.)

nothings:
Concur.

zmatsuo:
I remove wchar_t filename fnctions from stb_image.h and stb_image_write.h.
I modify stbi_write_XXX_to_file() in stb_image_write.h .
stbi_write_XXX_to_file() works similarly to stbi_XXXX_from_file().
We can use fopen variations(utf8,acp,utf16 filename..) through these function.

And I add test/image_test_wchar.c.
This file is test code for stbi_write_XXX_to_file() and stbi_XXXX_from_file(),
And This file is fopen variation sample.

nothings:
What is the purpose of the stbi__fopen_rb change in stb_image.h?

nothings:
Oh, I bet that was leftover because the wchar_t version of fopen takes wchar mode strings. So you presumably should just get rid of the stb_image.h changes entirely.

nothings:
The _to_file approach isn't unreasonable, but I think the idea outlined by rygorous above is more straightforward. If we include a UTF8->UTF16 that works internally on windows, and also add a UTF16->UTF8 helper function so people on windows who already have a utf16 filename can use that helper, I think that would be pretty ideal.

Does that sound reasonable?

nothings:
The _to_file approach isn't unreasonable, but I think the idea outlined by rygorous above is better since we avoid exploding the API functions. If we include a UTF8->UTF16 that works internally on windows, and also add a UTF16->UTF8 helper function so people on windows who already have a utf16 filename can use that helper, I think that would be pretty ideal.

Does that sound reasonable?

DanielGibson:
See https://github.com/DanielGibson/Snippets/blob/filehandling/DG_filehandling.h#L445-L458 for an example for such an fopen-wrapper (yes, it's public domain code)

For UTF16->UTF8 conversion something like `WideCharToMultiByte(CP_UTF8, 0, wCharString, -1, utf8outBuffer, utf8outBufferSize, NULL, NULL)` can be used

`WideCharToMultiByte()` and `MultiByteToWideChar()` are in kernel32.dll, i.e. it should always be available and doesn't create a dependency on a CRT

zmatsuo:
```
  FILE *f = stbi__fopen(filename, "rb");  // 1
  FILE *f = stbi__fopen_rb(filename);     // 2
```
I think that compiler will create smaller code than 1.
All stbi_fopen() in stb_image.h has "rb".
It is a very very small improvement :-)

zmatsuo:
```
UTF-16(wchar_t string)
|
| [helper function]
|
v
UTF-8 string
|
v
stbi_write_XXX()  (API require UTF-8 strings)
```

Now, stbi_write_XXX() APIs use fopen(). fopen() use ANSI codepage string on Windows (see https://en.wikipedia.org/wiki/Windows_code_page#ANSI_code_page and https://www.google.co.jp/search?q=getacp+windows ).

APIs which require ANSI codepage string filename parameter are necessary.
If APIs require UTF-8 string..

```
ANSI code page string(char string)
|
| [helper function]
|
v
UTF-8 string
|
v
stbi_write_XXX()  (API require UTF-8 strings)
```

It is not reasonable.

If we add APIs which require UTF-8 strings, thay are unnecessary in environment using UTF-8 string (linux etc).

When we add APIs which use wchar_t strings,
string conversion is unnecessary.
I think it is very reasonable.


stbi_write_XXX_to_file() API can be used like stbi_load_from_file() in stb_image.h. This api has no filename parameter. The user of the header can decide encoding of the file name and pass FILE *.

And we can also use to_file() and from_file() APIs to serialize multiple images (and data) into one data file.

ex. (pseudo code)
```
  FILE *fp = _wfopen(L"data.bin", L"wb");
  fwrite(..., fp);
  stbi_write_png_to_file(fp, ...);
  fwrite(..., fp);
  stbi_write_png_to_file(fp, ...);
  fwrite(..., fp);
  stbi_write_png_to_file(fp, ...);
  fclose(fp);

  FILE *fp = fopen("data.bin", "rb");
  fread(..., fp);
  stbi_load_from_file(fp, ...);
  fread(..., fp);
  stbi_load_from_file(fp, ...);
  fread(..., fp);
  stbi_load_from_file(fp, ...);
  fclose(fp);
```

If we can add the wchar_t string filename APIs, we can easily use it from Visual Studio. In Visual Studio, wchar_t is commonly used as a character string encoded by Unicode (UTF-16LE), and it can easily be written as L "filename.txt" in the source. There is no compatibility problem with both exeutable file and OS setting on Windows (https://www.google.co.jp/search?q=windows+control+panel+Region+language+non+unicode).

I agree that the number of APIs increases.
And I think that wchar_t string (and char string) filename APIs will become a small wrapper of stbi_write_XXX_to_fileto().

DanielGibson:
If I understood correctly, what @rygorous  and @nothings are suggesting is:
* The API remains as it is
* **On Windows**, the `const char*` arguments should be handled as UTF-8, and converted to `wchar_t` which is then passed to `_wfopen()`.
  * no change on other platforms, they should all support UTF-8 directly
  * even on Windows, this changed behavior should only be used only `#ifdef STBI_WIN32_UTF8_FILENAMES` (or similar name), otherwise (for backwards compatibility reasons) `const char*` is still treated as ANSI and regular `fopen()` is used
  * it's the users problem how they get a UTF-8 filename/path
  * to help with that, maybe add a `wchar_t*` -> UTF-8 string conversion function they can use

nothings:
@zmatsuo

1. I'm not concerned with small optimizations like fopen_rb. Not worth it.

2. Maybe your issue is that the thing @rygorous proposed and I concurred with will result in double-conversion on Windows if you start with wchar_t filenames -- you'll convert them to utf8, and then they will be converted back to wchar_t. *I don't care about the inefficiency of this double conversion.* I prefer the reduced complexity of not doubling the number of filename API functions to worrying about performance conversion of filenames (especially when decoding images is going to take much longer anyway).

zmatsuo:
I think that it is useful if we can use UTF-8 string file name.
(However, it is useful for me to have a function of wchar_t version, but I understand that you do not like increasing functions :-)
I will call the APIs with file name converted to UTF-8 from wchar_t string.

In code page 1252 (ISO 8859), in effect, conversion of UTF-8 string may not be necessary.
The CJK environment is inconvenient.

Please close this PR.

PR #534, opened 2018-01-02
Using secure CRT calls to avoid MSVC compile errors

### LINKED ISSUE #533: Secure CRT Errors with Visual Studio
https://github.com/nothings/stb/blob/9d9f75eb682dd98b34de08bb5c489c6c561c9fa6/stb_image_write.h#L233

If I compile stb_image_write.h in Visual Studio 2017 (likely previous versions as well) I get the following sort of errors, "This function or variable may be unsafe", for functions such as fopen, sprintf, etc. Is there any anticipation for fixing these? I understand they used to be warnings (they're errors now), but it would be nice to avoid having to turn off CRT deprecation errors just for this header.

I can probably whip up a solution if need-be. I can't believe I haven't see anything here about this yet.

xeekworx:
Fixes #533 for CRT calls that the Microsoft C/C++ compiler considers deprecated and errors when used. The proposed changes are to use the secure versions of these calls if a Microsoft compiler is being used. For this fix, namely, fopen and sprintf.

If you're using an extremely old version of a Microsoft compiler this might be a problem without being more detailed about the _MSC_VER macro.

### DIFF
--- a/stb_image_write.h
+++ b/stb_image_write.h
@@ -10,8 +10,8 @@
 
    Will probably not work correctly with strict-aliasing optimizations.
 
-   If using a modern Microsoft Compiler non-safe versions of CRT calls may cause 
-   compilation warnings or even errors. To avoid this, aldo before #including,
+   If using a modern Microsoft Compiler, non-safe versions of CRT calls may cause 
+   compilation warnings or even errors. To avoid this, also before #including,
 
        #define STBI_MSC_SECURE_CRT
 
--- a/stb_image_write.h
+++ b/stb_image_write.h
@@ -10,6 +10,11 @@
 
    Will probably not work correctly with strict-aliasing optimizations.
 
+   If using a modern Microsoft Compiler non-safe versions of CRT calls may cause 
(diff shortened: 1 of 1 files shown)


nothings:
Cool. I'll update this for old MSVC compilers, you don't have to worry about it.

nothings:
Actually, since you made it ifdef'd on an explicit setting, I'll leave it as is since on old compilers you just won't turn it on.

stb_image_write.h
```diff
@@ -230,7 +235,12 @@ static void stbi__stdio_write(void *context, void *data, int size)
 
 static int stbi__start_write_file(stbi__write_context *s, const char *filename)
 {
-   FILE *f = fopen(filename, "wb");
+   FILE *f;
+#ifdef STBI_MSC_SECURE_CRT
+   fopen_s(&f, filename, "wb");
```
DanielGibson: I'm not sure this code works as is - are you sure `fopen_s()` sets f to NULL if it fails?

nothings:
I'll fix it on the assumption that it doesn't, thanks @DanielGibson

PR #536, opened 2018-01-10
Fix #430

### LINKED ISSUE #430: stb_sprintf: snprintf fails on size-only queries
stb's snprintf behaves differently from C99 snprintf when only the size is requested, so this query fails:

```
size_t size1 = stbsp_snprintf( nullptr, 0, fmt, ... ); // always returns 0
size_t size2 = snprintf( nullptr, 0, fmt,  ... ); // returns the length of the computed string
assert(size1 == size2); // fails
```

for now I solve the problem by providing my own callback, but this behavior is quite unexpected.

update: possibly a duplicate of #408, also referenced in ocornut/imgui#1038

r-lyeh-archived:


nothings:
So, github shows the mergeinstructions for this as a patch instead of a full PR. Did you create it in a different than normal way?

The problem is that the merge instructions are:
    git checkout -b r-lyeh-archived-master master
    curl -L https://github.com/nothings/stb/pull/536.patch | git am -3

But if I run the curl command I get "Sorry, this diff is unavailable".

nothings:
Attention @r-lyeh

nothings:
I guess I can just merge it by c&p from the web since it's only two blocks.

nothings:
Yeah, ok, I did that. Fixed in next release.

nothings:
Oh right, this was paniq's fix from issue 430 anyway. (Because the 430 was only in the title, no link to it was automatically generated.)

r-lyeh:
It was a regular PR but the account I have archived that github account meanwhile, so I guess the PR is not available for some reason :)

Anyways, good that you merged it somehow!

PR #538, opened 2018-01-29
Perlin, added seed, fixed fractal wrapping, performance increase

Auburns:
Added seed argument to `stb_perlin_noise3` this allowed me to fix the fractal artifact, but it is also a great option to have for noise generation. Since the seed is an unsigned char it can be added to the first permutation lookup without any risk of out of bounds array lookups, this also means it has virtually zero performance impact.

Changed the final permutation table lookup to use a version of `stb__perlin_randtab` with all values mod 12. This removes the need for the indicies table in `stb__perlin_grad` since the value is already within the possible gradient indies. It also does a better job of reducing the gradient bias. Removing this extra array lookup gave me a 3% performance increase, MSVC17.

Made each octave of the fractals use a different seed, this prevents [this artifact](https://i.imgur.com/kER5Gk6.jpg) from happening at the origin. This happens because each octave is using very similar positions to the previous octave, having a different seed per octave guarantees a different noise output despite the similar position.

Removed wrap arguments from the fractal functions as they cannot function correctly in a fractal due to each octave wrapping at a different offset.

### DIFF
--- a/stb_perlin.h
+++ b/stb_perlin.h
@@ -1,4 +1,4 @@
-// stb_perlin.h - v0.3 - perlin noise
+// stb_perlin.h - v0.4 - perlin noise
 // public domain single-file C implementation by Sean Barrett
 //
 // LICENSE
@@ -18,7 +18,8 @@
 //                           float z,
 //                           int   x_wrap=0,
 //                           int   y_wrap=0,
-//                           int   z_wrap=0)
+//                           int   z_wrap=0,
+//                           unsigned char seed=0)
 //
 // This function computes a random value at the coordinate (x,y,z).
 // Adjacent random values are continuous but the noise fluctuates
@@ -32,6 +33,9 @@
 // details of the implementation, even if you ask for larger or no
 // wrapping.)
 //
+// Using a different "seed" will provide a unique variation
+// of the noise output.
+//
 // Fractal Noise:
 //
 // Three common fractal noise functions are included, which produce 
@@ -41,15 +45,15 @@
 //
 // float stb_perlin_ridge_noise3(float x, float y, float z,
 //                               float lacunarity, float gain, float offset, int octaves,
-//                               int x_wrap, int y_wrap, int z_wrap);
+//                               unsigned char seed=0)
 //
 // float stb_perlin_fbm_noise3(float x, float y, float z,
 //                             float lacunarity, float gain, int octaves,
-//                             int x_wrap, int y_wrap, int z_wrap);
+//                             unsigned char seed=0);
 //
 // float stb_perlin_turbulence_noise3(float x, float y, float z,
 //                                    float lacunarity, float gain,int octaves,
-//                                    int x_wrap, int y_wrap, int z_wrap);
+//                                    unsigned char seed=0)
 //
 // Typical values to start playing with:
 //     octaves    =   6     -- number of "octaves" of noise3() to sum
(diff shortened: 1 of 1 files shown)


Auburns:
As a note, this change will cause a reseed of all noise produced compared to before, due to a new method for selecting gradients.

nothings:
1. I assume the argument for removing the wrapping is not because they have different *offsets* (which shouldn't affect wrapping at all), but that they have non-power-of-two *lacunarity*, so they don't actually wrap. (The whole purpose of the wrap parameters in noise3 is that you *can* making wrapping multi-octave noise, but only if you use power-of-two scales and corresponding wrapping.)

2. The intent has always been that the third coordinate of 3D noise can be used as a seed parameter when you want 2D noise. Is there actual demand for seeded 3D noise?

3. I don't think a 3% speedup is worth breaking back-compatibility of noise3. (I'm less concerned about breaking back-compatibility of ridge/fbm/turbulent noise.) But I'm not clear why noise3's behavior changed, since the table changes you described don't seem like they would change gradient selection.

OvermindDL1:
> Is there actual demand for seeded 3D noise?

Yes, definitely.  Especially in voxel setups.  In a few cases I've even used up to 6 dimensional perlin noise (yay C++ templates for that horror, abstracted it out into N dimensions long ago).

Auburns:
1. Yes sorry I should have been more clear. The wrapping would still
basically work if using lacunarity=power-of-two, although the higher
octaves would be repeating unnecessarily. If wrapping was to be implemented
for fractals it would have to enforce that lacunarity=power-of-two and it
should correctly be adjusted for higher octaves so they also wrap at the
same location as the base octave.

2. The seed for perlin was mainly to fix the artifact issue for fractals.
And it adds the option to seed 3D noise, since this has negligible
performance impact I feel it is a good addition.

3. Since I replaced the indicies table with the mod12 table it will give a
different hash resulting in a different gradient being selected. I could
remove this change without affecting any of the other changes, you could
then produce the same output as before using a seed of 0.

On 29 Jan 2018 11:34 pm, "Sean Barrett" <notifications@github.com> wrote:


   1.

   I assume the argument for removing the wrapping is not because they have
   different *offsets* (which shouldn't affect wrapping at all), but that
   they have non-power-of-two *lacunarity*, so they don't actually wrap.
   (The whole purpose of the wrap parameters in noise3 is that you *can*
   making wrapping multi-octave noise, but only if you use power-of-two scales
   and corresponding wrapping.)
   2.

   The intent has always been that the third coordinate of 3D noise can be
   used as a seed parameter when you want 2D noise. Is there actual demand for
   seeded 3D noise?
   3.

   I don't think a 3% speedup is worth breaking back-compatibility of
   noise3. (I'm less concerned about breaking back-compatibility of
   ridge/fbm/turbulent noise.) But I'm not clear why noise3's behavior
   changed, since the table changes you described don't seem like they would
   change gradient selection.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/nothings/stb/pull/538#issuecomment-361424240>, or mute
the thread

nothings:
Re: 3: my point is instead of setting new_table[i] to old_table[i]%12, you could instead set new_table to indices[old_table[i]&63], thus eliminating the lookup without changing the old behavior.

2. If the main point is to fix the artifact issues for fractals, I'd prefer it to not change the API to noise3. (Alternatively, if noise3's public API is going to have a seed, it should be more than 8-bits. But that would require a different implementation.)

So assuming that's not worth tackling (since it would cause a slowdown), change your current stb_perlin_noise3 to stb_perline_noise_internal, add an stb_perlin_noise3 that just calls it with 0 as the seed, and have the other functions call perline_noise3_internal with their seed.

To be clear about this, noise3 is what I implemented, and the other stuff is something that someone else added that I don't care about particularly. So changing those to be better behaved is fine, but I don't want to change noise3 unnecessarily. Here, that shows up as avoiding changing the API (the seed) and avoiding changing the behavior back-compatibility-wise.

nothings:
Looks good, will integrate soon.

nothings:
"soon", sigh.

In next release.

PR #547, opened 2018-02-01
Replace raw math calls with STBTT_* decorated variants

pmttavara:
Also, adds STBTT_fmod(x,y) to the list of define-able math functions.
This lets the lib be able to build without the CRT, as (I assume) was intended.
This is not hard to fix user-side, but I thought it might as well work out of the box.
I put my name in the contributors because the rules said so, but this is a pretty trivial change.


nothings:
Fix already made for next version, thanks though!

PR #550, opened 2018-02-03
fix stb_vorbis.c

XProger:
fix FPU exception on PlayStation Portable (cmp NAN with ZERO) at the first "INVERSE COUPLING" stage


nothings:
I manually added the final memset from your PR, which is I assume the part relevant to this bug, and left out the STB_VORBIS_FAST_INIT stuff. In next release.

Issue #557, opened 2018-02-06
stb_image unicode support


nothings:
It would be nice to support Unicode filenames on windows. We can do this by making the char* be interpreted as UTF-8, converting to UTF-16, and using Windows UTF-16-base file operations. For back compatibility, the existing code-page behavior should be available on a define (or vice versa, old behavior is default, can ask for unicode explicitly--or even check for Windows UNICODE define). Also add a helper for windows that converts wchar_t to UTF-8 so if user has a wchar_t filename, they can use the utf-8 mode successfully. See discussion in #529

jrsmith17:
I provided a pull request for this issue [here](https://github.com/nothings/stb/pull/559). Unicode needs to be explicitly defined or it defaults to the old behavior. I also provided a helper function for converting wchar_t to UTF-8. Let me know if this meets your needs. Thanks!

nothings:
Fixed and released a couple weeks ago.

PR #559, opened 2018-02-12
Support for loading files with unicode names

jrsmith17:
Referencing issue [#557](https://github.com/nothings/stb/issues/557)

Coincidentally, I was already working on this issue separate from the discussion that happened on [PR529](https://github.com/nothings/stb/pull/529).

Notes:
- I didn't make any formal test cases, but I did test reading two texture files called "test-dotsひらがな.tga" and "test-dotß.tga". The files loaded with my modifications passed through to my OpenGL code as expected. I also tested writing the same filenames in a pet ray tracer (and the image contents matched between unicode and non unicode filenames)
- I saw discussion between using #ifdef UNICODE or STBI_UTF8. I opted for UNICODE because I figured someone who wanted to load a unicode named file probably already has that defined. 
- If I understood the consensus correctly, you wanted the user to provide a correct char* filename as input. I used [this library](https://github.com/nemtrif/utfcpp) for my personal test cases, but also provided the requested wchar_t to utf-8 utility function.

### DIFF
--- a/stb_image.h
+++ b/stb_image.h
@@ -1147,11 +1147,15 @@ static void stbi__float_postprocess(float *result, int *x, int *y, int *comp, in
 #ifndef STBI_NO_STDIO
 
 char* stbi_convert_wchar_to_utf8(wchar_t* input) {
+#ifdef _WINDOWS_
 	int outputSizeNeeded = WideCharToMultiByte(CP_UTF8, 0, &input[0], wcslen(input), NULL, 0, NULL, NULL);
-	char* temp = (char*)STBI_MALLOC(outputSizeNeeded);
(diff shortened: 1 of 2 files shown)


rygorous:
This pull request currently rewrites both files entirely. Also note the original files use 3-spaces indenting, not 2-spaces indenting.

As-is, this pull request is basically impossible to review, and will make the entirely of the per-line file histories for stb_image and stb_image_write useless, so I'm rejecting this particular PR.

Can you clean this up by replicating the changes without all the erroneous formatting changes in another branch, and turn that into a pull request? You should be able to cherry-pick that from this branch.

PR #560, opened 2018-02-12
GIMP 32-bit ABGR BMP support

NuklearBomb:
I was getting this error when loading a 32-bit ABGR BMP exported from GIMP:
```
stb_image.h:5047: stbi__shiftsigned: Assertion `v >= 0 && v < 256' failed.
Aborted
```
This pull request contains a fix for that.

Image used for testing: [GIMP-ABGR8.bmp.zip](https://github.com/nothings/stb/files/1715503/GIMP-ABGR8.bmp.zip)

I also used stb_write_image.h (stbi_write_png) to write a .png of the loaded .bmp:
gimp-abgr8

Success!


nothings:
The correct fix is to change
    static int stbi__shiftsigned(int v, int shift, int bits)
to
    static int stbi__shiftsigned(unsigned int v, int shift, int bits)

Fixed for next release.

PR #563, opened 2018-02-13
fixed 'unreferenced formal parameter' warnings in stb_image

jamesdolan:
Fixed some warnings introduced recently.

stb_image.h(6339): error C2220: warning treated as error - no 'object' file generated
stb_image.h(6339): warning C4100: 'req_comp': unreferenced formal parameter
stb_image.h(6563): warning C4100: 'ri': unreferenced formal parameter


technosaurus:
You can just use `(void)param;` to silence the warning where the parameter is not used.
As it is, it removes functionality where it was used and deletes the wrong parameter in others.
Do not merge as is.

nothings:
Superseded by #662.

PR #568, opened 2018-02-21
Re added unicode filename support for stb_image and stb_image_write w…

jrsmith17:
@rygorous 

Apologies for needing to do this twice. I thought I had gone back and cleaned up enough of the white space issues to be clear, but I definitely missed some.

Original PR can be referenced [here](https://github.com/nothings/stb/pull/559)

### DIFF
--- a/stb_image.h
+++ b/stb_image.h
@@ -1143,12 +1143,40 @@ static void stbi__float_postprocess(float *result, int *x, int *y, int *comp, in
 
 #ifndef STBI_NO_STDIO
 
+char* stbi_convert_wchar_to_utf8(wchar_t* input) {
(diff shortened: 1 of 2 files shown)


rygorous:
You don't need to apologize! I was just trying to explain why the giant diffs are bad news. :)

This looks much clearer, thanks. We'll have a look at this for the next release.

nothings:
Merged for next release. Thanks for your patience.

I significantly changed it to eliminate the mallocs and to add an explicit #define that triggers it.

nothings:
Merged for next release. Thanks for your patience.

I significantly changed it to eliminate the mallocs and to add an explicit #define that triggers it.

PR #570, opened 2018-02-23
stb_dxt: fix constant color check

pps83:
original constant color check fix obviously didn't work when request was to compress without alpha.
Code is updated to fix the issue and avoid creating temporary copy of the input buffer.
Spurious spaces were also picked up by my editor, if you prefer I may update Cl to remove them.
Thanks for stb.

### DIFF
--- a/stb_dxt.h
+++ b/stb_dxt.h
@@ -24,6 +24,7 @@
 // contributors: 
 //   Kevin Schmidt (#defines for "freestanding" compilation)
 //   github:ppiastucki (BC4 support)
+//   Pavel Pavlov (constant color check)
 // 
 // LICENSE
 //
@@ -517,7 +518,7 @@ static void stb__CompressColorBlock(unsigned char *dest, unsigned char *block, i
 
    // check if block is constant
    for (i=1;i<16;i++)
-      if (((unsigned int *) block)[i] != ((unsigned int *) block)[0])
+      if ((((unsigned int *) block)[i] & 0xffffff) != (((unsigned int *) block)[0] & 0xffffff))
          break;
 
    if(i == 16) { // constant color
@@ -651,7 +652,6 @@ static void stb__InitDXT()
 
 void stb_compress_dxt_block(unsigned char *dest, const unsigned char *src, int alpha, int mode)
 {
-   unsigned char data[16][4];
    static int init=1;
    if (init) {
       stb__InitDXT();
@@ -659,15 +659,8 @@ void stb_compress_dxt_block(unsigned char *dest, const unsigned char *src, int a
    }
 
    if (alpha) {
-      int i;
       stb__CompressAlphaBlock(dest,(unsigned char*) src+3, 4);
       dest += 8;
-      // make a new copy of the data in which alpha is opaque,
-      // because code uses a fast test for color constancy
-      memcpy(data, src, 4*16);
-      for (i=0; i < 16; ++i)
-         data[i][3] = 255;
-      src = &data[0][0];
    }
 
    stb__CompressColorBlock(dest,(unsigned char*) src,mode);
--- a/stb_dxt.h
+++ b/stb_dxt.h
@@ -80,6 +80,9 @@ STBDDEF void stb_compress_bc5_block(unsigned char *dest, const unsigned char *sr
 #if !defined(STBD_ABS) || !defined(STBI_FABS)
 #include <math.h>
 #endif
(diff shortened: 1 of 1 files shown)


stephenwhittle:
Cancel

pps83:
second commit adds constant alpha optimization

nothings:
Yeah, please remove the whitespace changes (you can submit them as a separate PR if you want), it makes it too hard to read the diffs on the web.

stb_dxt.h
```diff
@@ -651,23 +664,15 @@ static void stb__InitDXT()
...
    if (init) {
       stb__InitDXT();
       init=0;
    }
 
    if (alpha) {
-      int i;
       stb__CompressAlphaBlock(dest,(unsigned char*) src+3, 4);
       dest += 8;
-      // make a new copy of the data in which alpha is opaque,
```
nothings: Why is this code block deleted? According to the comment, the color code expects the alpha to be opaque for the color code's constancy test to work. Fixing the alpha-constant case doesn't seem to help this.

pps83: See change above on line #521 Instead of copying color data without alpha to tmp buf, the check in the loop is modified to remove alpha from the test when checking equality.

nothings: Oh right, I missed that. Continued discussion moved to main thread.

nothings:
I'm confused what the actual bug is, other than the documentation not being clear. You always have to pass RGBA color, and if you're not compressing with alpha, the A should be constant across all pixels. If you do that, I don't think there's a bug?

As far as I can tell your fix handles the case where the user passes in RGBA with varying A and says to ignore the alpha (which I agree the current documentation makes it sound like it's ok). But your fix introduces the assumption that the platform is little-endian, which will become a drawback if big-endian machines ever become popular again. So just changing the documentation to match the implementation is probably a safer fix.

pps83:
> I'm confused what the actual bug is, other than the documentation not being clear. You always have to pass RGBA color, and if you're not compressing with alpha, the A should be constant across all pixels. If you do that, I don't think there's a bug?

Title of the commit could be improved to properly reflect what the change does. Overall, these are the changes:
 - properly handle color data with varying alpha even if `alpha=false` is specified
 - avoid copying to temporary stack buffer and clearing alpha if `alpha=true` is specified
 - fast path check for constant alpha (similar to fast path for constant color).

currently fast path for constant color won't work if `alpha=false` and color data actually contains varying alpha.

> As far as I can tell your fix handles the case where the user passes in RGBA with varying A and says to ignore the alpha (which I agree the current documentation makes it sound like it's ok). But your fix introduces the assumption that the platform is little-endian, which will become a drawback if big-endian machines ever become popular again. So just changing the documentation to match the implementation is probably a safer fix.

My code doesn't introduce that assumption, it was always there:

> data[i][3] = 255;

Also, imo it's not not about BE being popular one day again (which is unlikely), but about what color format is expected by dxt compressor: even on BE platform you'd probably need to have colors in in the same format.

nothings:
The old code assumed byte-order RGBA without any assumptions about endianness. In other words, I believe (although I'm not 100% certain) that it never converts between bytes and dwords anywhere other than that constant check.

pps83:
Yes, you are correct, my mistake. Code would need to be updated to handle BE platforms.

nothings:
Right, so I think it's better to just say that the "dummy" alpha channel has to be constant if you're not using alpha, rather than change the code. Make sense?

pps83:
> Right, so I think it's better to just say that the "dummy" alpha channel has to be constant if you're not using alpha, rather than change the code. Make sense?

Proper "fix" would be to update the title of the PR: it's adds a [few improvements (2,3)](https://github.com/nothings/stb/pull/570#issuecomment-462144938) rather than fixing anything.

In any case, constant alpha optimization is a very common case:, why would you compress something that doesn't need to be compressed?

nothings:
Is item 2 actually an improvement? Copying introduces overhead, but so does making the color test more complicated. Without performance testing it's not clear to me it's worth changing, and it's definitely not worth changing if it introduces an endianness dependency.

Item 3 seems like a perfectly reasonable idea, although shouldn't the index data be just all 0s? Maybe I'm missing something.

nothings:
Ok, submitter didn't reply to my questions, so I'm just going to make the suggested documentation change. Anything that still needs to change can be resubmitted, preferably as separate pull requests given the issues I had with this.

PR #571, opened 2018-02-23
stb_truetype: fix unused variable warning when asserts are disabled.

ocornut:
Minor warning fix when asserts are off, `pairSetCount` is only used for an assert.
(reported to me here https://github.com/ocornut/imgui/issues/1642)


nothings:
merged for next release

PR #575, opened 2018-03-08
stb_truetype: Added comment about stbtt_FindGlyphIndex() return value.

ocornut:
Added a simple comment:
`Returns 0 if the character codepoint is not defined in the font.`

Based on this thread:
https://twitter.com/ocornut/status/971757064983924742

Looking at the implementation for `stbtt_FindGlyphIndex`, it's clearly intended at the error value, but it is not obvious to me that that 0 isn't also valid index.


nothings:
merged in next release

PR #577, opened 2018-03-12
add indexed color image bmp

zmatsuo:
I add indexed color bmp in stbi_write_bmp().
With this PR, we will able to save 1,4,8 bit per pixel.
Please review this.


nothings:
I think I'd rather stay away from the complexity of indexed colored support in this library, since that's a huge can of worms. Thanks for your effort though.

PR #585, opened 2018-03-22
Fix `x` was multipled twice, resulting buffer-overrun bug. Fixes #567.

### LINKED ISSUE #567: stbi_image_write_hdr has incorrect stride calculation
Hi,

I recently swapped v. 1.07 with v.2.19 and I get an "access violation reading..." error in stbi_image_write_hdr. I tested it with a simple piece of code, and it still breaks:
```
        int w = 100;
	int h = 100;
	int n = 1;
	float* data = new float[w*h*n];
	for(int y=0;y<h;++y)
		for (int x = 0; x < w; ++x)
		{
			for (int i = 0; i < n; ++i)
				data[n*(x + y*w) + i] = float(y) / float(h);
		}
	stbi_write_hdr("test.hdr", w, h, n, data);
```
This and my other code works perfectly fine with v 1.07.

syoyo:
This PR should fix #567. I can provide unit test if required.

### DIFF
--- a/stb_image_write.h
+++ b/stb_image_write.h
@@ -694,7 +694,7 @@ static int stbi_write_hdr_core(stbi__write_context *s, int x, int y, int comp, f
       s->func(s->context, buffer, len);
 
       for(i=0; i < y; i++)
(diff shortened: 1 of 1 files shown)


rygorous:
I nominated PR #644 for the fix. We got multiple submissions. Thanks anyway! :)

Issue #592, opened 2018-04-08
stb_image_write.h v1.09: Access violation when writing a jpeg at 1680x1050 resolution.


maraakate:
Hi STB,

I've found that when trying to write a jpeg at 1680x1050 resolution Visual Studio will crash out with an Access Violation error and points to the following function: stbiw__jpg_processDU

The exact line it is unhappy with is r = imageData[p+0];
p = -10080

The parameters sent to stbi_write_jpeg:
	stbi_flip_vertically_on_write(1);
	bSuccess = stbi_write_jpg(checkname, grab_width, vid.height, 3, buffer, 100);

I am able to write out a PNG just fine at the same resolution with stb.

Hope this helps and thanks for all your hard work!
Frank

zao:
Reproduced this with a fresh project off master and a buffer of the accurate dimensions.
Only happens for heights that are not multiples of 8, which 1050 notably is.
Over-allocating storage around the image results in a successful write, but the file ends up larger in file size than it should be, which hints to reading unexpected data.

nothings:
Oh yeah, this is on me then, I wrote the flip_vertically support and didn't thoroughly test it. Don't worry about it further @DanielGibson.

DanielGibson:
Ok, no worries :)

I guess the
```c
                  if(row >= height) {
                     p -= width*comp*(row+1 - height);
                  }
```
part must also be adjusted for `stbi__flip_vertically_on_write` somehow

DanielGibson:
however, I find the following code more readable, and it's even shorter:
```c
                  float r, g, b;
                  int p;
                  if(row < height) {
                     p = (stbi__flip_vertically_on_write ? (height-1-row) : row)*width*comp;
                  } else {
                     // row >= height => use last input row (=> first if flipping)
                     p = stbi__flip_vertically_on_write ? 0 : ((height-1)*width*comp);
                  }
                  // if col >= width => use pixel from last input column
                  p += ((col < width) ? col : (width-1))*comp;

                  r = imageData[p+0]; // etc, no changes from here on
```

maraakate:
What's weird though is writing a PNG has no problems for me at 1680x1050.

nothings:
There's nothing weird about that, flip_vertically_on_write is handled internally by each of the 5 file formats, rather than being a single systematic approach, so it's totally separate code in each case.

maraakate:
[...]

PR #593, opened 2018-04-08
stb_image_write.h: Fix jpg flipping for non-multiple-of-8 sizes

DanielGibson:
JPG always encodes 8x8 pixel blocks. If the input image does not have
a width or height that's a multiple of 8, the last column or row is just
used multiple times for the remaining pixels of the block.
The original code first calculated p (the index into the pixel data)
with the "imaginary" row/colum (that might be up to 7 pixels too far
into each direction) and then subtracted the necessary amount of bytes
if row >= height or col >= width.
That was a bit cryptic (IMHO), and didn't get more readable/obvious when
vertical flipping was added - which introduced a bug, by not taking
stbi__flip_vertically_on_write into account when adjusting p for
row >= height, which lead to negative p, which lead to a crash...

The code should be more obvious (and less buggy) now.

(I wasn't sure about the comments, are they too obvious to be helpful?)

This fixes bug #592

### DIFF
--- a/stb_image_write.h
+++ b/stb_image_write.h
@@ -1424,14 +1424,16 @@ static int stbi_write_jpg_core(stbi__write_context *s, int width, int height, in
             float YDU[64], UDU[64], VDU[64];
             for(row = y, pos = 0; row < y+8; ++row) {
                for(col = x; col < x+8; ++col, ++pos) {
-                  int p = (stbi__flip_vertically_on_write ? height-1-row : row)*width*comp + col*comp;
                   float r, g, b;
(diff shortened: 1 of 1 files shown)


rygorous:
Patch looks good but I'll probably change it before merging to compute the clamped `row` in the outer loop, because this just feels bad as-is. :)

Anyway, thanks, will merge this soon-ish.

fungos:
Just got bitten by this one. I'm now using the patch from @DanielGibson.

nothings:
merged for next release

PR #572, opened 2018-02-26
Fix overflow in stbi_write_hdr_core()

poppolopoppo:
b056850ea9118b69325d973f8aef7f843527e299 left an additional multiplication by x,
leading to overflow in stbiw__write_hdr_scanline().

just removed this little rascal.

PR #594, opened 2018-04-11
fixed a bug in stbi_write_hdr_core that prevented hdr images from bei…

kopaka1822:
The scanline offset was multiplied by x*x instead of just x


nothings:
dupe of #572

Issue #515, opened 2017-10-29
Xcode 9 reports many instances of "Possible misuse of comma operator"


enci:
When building with Xcode 9, I get many "Possible misuse of comma operator" warnings. They are all from multiple consecutive value assignments. While I think they can be safely ignored, it's still better no to have them filling up the log.

nothings:
Can you capture the warning output so I can make sure I fix them all?

Piot:
I chime in here, because I have similar problems in `stb_image.h` :)

in `stbi__convert_format` (line 1541-):
```
      switch (STBI__COMBO(img_n, req_comp)) {
         STBI__CASE(1,2) { dest[0]=src[0], dest[1]=255;                                     } break;
         STBI__CASE(1,3) { dest[0]=dest[1]=dest[2]=src[0];                                  } break;
         STBI__CASE(1,4) { dest[0]=dest[1]=dest[2]=src[0], dest[3]=255;                     } break;
         STBI__CASE(2,1) { dest[0]=src[0];                                                  } break;
         STBI__CASE(2,3) { dest[0]=dest[1]=dest[2]=src[0];                                  } break;
         STBI__CASE(2,4) { dest[0]=dest[1]=dest[2]=src[0], dest[3]=src[1];                  } break;
         STBI__CASE(3,4) { dest[0]=src[0],dest[1]=src[1],dest[2]=src[2],dest[3]=255;        } break;
         STBI__CASE(3,1) { dest[0]=stbi__compute_y(src[0],src[1],src[2]);                   } break;
         STBI__CASE(3,2) { dest[0]=stbi__compute_y(src[0],src[1],src[2]), dest[1] = 255;    } break;
         STBI__CASE(4,1) { dest[0]=stbi__compute_y(src[0],src[1],src[2]);                   } break;
         STBI__CASE(4,2) { dest[0]=stbi__compute_y(src[0],src[1],src[2]), dest[1] = src[3]; } break;
         STBI__CASE(4,3) { dest[0]=src[0],dest[1]=src[1],dest[2]=src[2];                    } break;
         default: STBI_ASSERT(0);
      }
```

and in `stbi__convert_format16` (line 1590-):
```
      switch (STBI__COMBO(img_n, req_comp)) {
         STBI__CASE(1,2) { dest[0]=src[0], dest[1]=0xffff;                                     } break;
         STBI__CASE(1,3) { dest[0]=dest[1]=dest[2]=src[0];                                     } break;
[...]

PR #596, opened 2018-04-25
fix comma warnings when building with -Wcomma

kroko:
as the title says, this fixes comma warnings when building with `-Wcomma`.  it also addresses  https://github.com/nothings/stb/issues/515 as vanilla Xcode project configuration will call clang with this flag.


nothings:
This is pretty lame, because I use commas as a way to show a semantic grouping, like assigning all four channels of an RGBA pixel at once... but merged for next release.

kroko:
well, don't be sad, because IMHO semantic grouping stays as you are doing these assignments in one line meaning

```c
a=1 b=2 c=3
```

insted of

```c
a=1
b=2
c=3
```

and it already serves as semantic grouping.

thank you!

PR #598, opened 2018-04-28
Added appveyor support for Windows builds.

AndrewJDR:
This adds appveyor.com support for VS2008 - VS2017. This works similarly to travis-ci, giving build success reports any time someone submits a PR. You can see what it looks like here:
https://ci.appveyor.com/project/AndrewJDR/stb/build/1.0.24

And on my fork (see green checkmark): https://github.com/AndrewJDR/stb/commits/add_appveyor_windows

It also runs the test executables just to make sure they don't crash or whatever. Not sure if that's necessary, but why not.

You'll have to sign up for it to work, similar to travis-ci. Signup is free, and straightforward for OSS projects. Build-wise, I used nmake for now by making a copy of the original tests/Makefile and removing a few flags that cl doesn't understand, but you can replace nmake with whatever you want later by changing these lines in appveyor.yml:
```
  - call "C:\\Program Files (x86)\\%VSINSTALL%\vcvarsall.bat" %platform%
  - cd tests
  - nmake /f Makefile.nmake
```

I don't think travis supports windows, or I would have just used that. travis + appveyor seems to be a common combination on OSS projects for this reason.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,51 @@
+image: Visual Studio 2015
+
+platform:
+- x64
+- x86
+
+environment:
+  matrix:
+
+  - VS_VERSION: Visual Studio 2008
+    VSINSTALL: "Microsoft Visual Studio 9.0\\VC"
+
+  - VS_VERSION: Visual Studio 2010
+    VSINSTALL: "Microsoft Visual Studio 10.0\\VC"
+
+  - VS_VERSION: Visual Studio 2012
+    VSINSTALL: "Microsoft Visual Studio 11.0\\VC"
+
+  - VS_VERSION: Visual Studio 2013
+    VSINSTALL: "Microsoft Visual Studio 12.0\\VC"
+
+  - VS_VERSION: Visual Studio 2015
+    VSINSTALL: "Microsoft Visual Studio 14.0\\VC"
+
+  - VS_VERSION: Visual Studio 2017
+    VSINSTALL: "Microsoft Visual Studio\\2017\\Community\\VC\\Auxiliary\\Build"
+    APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2017
+
+build_script:
+  - echo build_script
+  - echo platform %platform%
+  - call "C:\\Program Files (x86)\\%VSINSTALL%\vcvarsall.bat" %platform%
+  - cd tests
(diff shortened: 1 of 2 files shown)


AndrewJDR:
Another quick note -- it does both x86 and x64 builds for vs2013-vs2017, but with vs2008-vs2012, x64 builds are not very nicely supported with appveyor so I just stuck with x86 there.

nothings:
At this point I already have two different makefile-like things--the Makefile for travis, and my build matrix for my local VS builds, test.sbm: https://github.com/nothings/stb/blob/master/tests/test.sbm

I don't want to maintain yet ANOTHER redundant Makefile, so I don't think I'm going to do this until I'm ready to have a single program generate all the makefiles/etc. from a single list of tests.

It doesn't need to be a complicated input file, e.g. it could just be each line is a list of all the c/cpp files to link for a single project, and then generate the rest of the Makefile by very simple logic with that. I wouldn't recommend anybody try to implement that without talking to me first, because I'm not sure what it'll take for it to be a system I'm comfortable with having.

AndrewJDR:
@nothings It may be quite possible to re-use the travis makefile for this purpose with very few changes to it, so that we are not adding an additional makefile into the mix.

Is that something you'd be interested in merging? If so, I can look into it.

nothings:
Sure.

AndrewJDR:
nmake and gnu make lack a unified way of expressing an if statement, so it's not trivial.

Is the logic so simple that it could be expressed via the C preprocessor? If both the input file and the logic are sufficiently simple, the input file could just be run through `cpp` (unix) / `cl /E` (windows)...

nothings:
I'd rather not use the C preprocessor. It probably makes more sense to just have a little program that generates the makefile and test.sbm anyway. Possibly use a revised test.sbm itself as the input.

AndrewJDR:
>  I wouldn't recommend anybody try to implement that without talking to me first, because I'm not sure what it'll take for it to be a system I'm comfortable with having.

I could take a crack at it, if you care to write a bit about what you're looking for. I'm busy on a project so it would just happen on a weekend or day where I find myself with some time to spare and/or I need a break. If I had to guess, sometime within the next 3-4 months?

If you see yourself most likely taking care of it yourself before then, it may not be worth your time to describe it.

PR #599, opened 2018-04-29
add tests to tests/test_sprintf.c

wojdyr:
I wanted to have some tests for `stb_sprintf.c` because I was making a few changes to this library. I tried to get good coverage with not too many tests. It's kind of fun watching as the coverage numbers go up. They went up to 99% of lines and 88% of branches.

The tests can be run with either stb_sprintf or glibc sprintf. To run it with glibc change 1 to 0 in the line:
```c
#define USE_STB 1
```
When the output differs I included both version (see `#if USE_STB` in the file). For example, default precision of `%a` differs, capitalization of inf/nan differs, and a couple of other insignificant details.

As for [my changes](https://github.com/project-gemmi/gemmi/blob/master/third_party/stb_sprintf.h) in `stb_sprintf.c` -- I'm not proposing them here because they add dependency on `string.h` and de-optimize the library a bit. But they avoid UB and warnings from sanitizers.

stb_sprintf is a fantastic library. For me the best feature is that it's locale-independent, so I can have locale-specific number formatting in the UI and locale-independent one in data files.

### DIFF
--- a/tests/test_sprintf.c
+++ b/tests/test_sprintf.c
@@ -1,1 +1,177 @@
-#include "stb_sprintf.h"
+
+#define USE_STB 1
+
+#if USE_STB
+# include "stb_sprintf.h"
+# define STB_SPRINTF_IMPLEMENTATION
+# include "stb_sprintf.h"
+# define SPRINTF stbsp_sprintf
+# define SNPRINTF stbsp_snprintf
+#else
+# include <locale.h>
+# define SPRINTF sprintf
+# define SNPRINTF snprintf
+#endif
+
+#include <assert.h>
+#include <math.h>    // for INFINITY, NAN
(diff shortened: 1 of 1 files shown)


cheako:
Why was this ignored?  I'm interested in working on stb_image and want to attempt detection of regressions.

nothings:
I do still hope to merge this at some point.

nothings:
I had to drastically change this to make it VC6 compatible. Also, you changed the semantics of the existing test_sprintf.c file; I changed it back so it can link with the other tests, but maybe it should be moved to a separate file.

wojdyr:
> but maybe it should be moved to a separate file.

Maybe -- I don't have any opinion here.

nothings:
Yeah, I realize you may not be actively working on it anymore.

PR #604, opened 2018-05-02
Fix hdr scanline error

tviksand:
Minor bugfix, hdr_write had an array out of bounds access due to a minor typo.


nothings:
dupe of #572

PR #606, opened 2018-05-06
stb_textedit: remove double initialization

mastensg:
clang's `scan-build` found this


nothings:
merged for next release

PR #607, opened 2018-05-07
stb_truetype: stbtt_PackFontRanges skip missing glyphs.

ocornut:
The functions used by `stbtt_PackFontRanges()` don't skip glyphs that are missing in the font, which result in packing and rasterizing whatever the "fallback" font glyph is. I am using the word "fallback" but I am not sure I really understand if `stbtt_FindGlyphIndex()` returning 0 could _also_ be a valid glyph. 
See #575 about `stbtt_FindGlyphIndex()` documentation.

If I'm taking "Arial Unicode.ttf" and trying to e.g. 32 to 255, notice the squares about 127:
image

Notice they take space in the font atlas:

image
(I disabled oversampling here to make this more readable)

While this extraneous space is bearable for a typical ASCII set, when you work with larger sizes, international font ranges, icon font ranges it can be taking more space.

Additionally, this also means that `stbtt_PackFontRanges()` provide no information on whether a character was successfully found/packed/rasterized: from the user's point of view, all of the glyphs packing are successful.

As an explicit problematic case: FontAwesome version 5 (latest version of a very popular icon font) recently started shipping over multiple font files instead of a single one, which makes is more difficult to use those icons by specifying a single font range (as all the fonts files are now sharing the same range, but each individual font file only cover some of the codepoint within that range).

One workaround right now is for the user to call `stbtt_FindGlyphIndex()` and recreate font ranges manually from the input font ranges. It's possible but not really trivial for a casual end-user application, and the number of holes may easily turn a single range into hundreds of ranges (this would be the case for the icon fonts mentioned above).

**PROPOSED SOLUTION**

The proposed PR skips missing glyphs.
This requires the end-user to check that the output `stbtt_packedchar` are valid by testing positions, e.g.

```cpp
            for (int char_idx = 0; char_idx < range.num_chars; char_idx += 1)
            {
                const stbtt_packedchar* pc = &range.chardata_for_range[char_idx];
                if (!pc->x0 && !pc->x1 && !pc->y0 && !pc->y1)
                    continue;
```

This was in theory _already_ required in the past but I think it would only happen on actual packing failure (no room) which arguably is rarer, so it's possible that some code didn't test for it. I don't imagine it to make much of a difference.

Note that the "square" shown in Arial Unicode above is a glyph itself and not all font show something that looks like a square at all.

**HOWEVER**

This is reliant on `stbtt_FindGlyphIndex() == 0` being _always_ an error value. It's unclear to me if some legit glyph could return 0, since the data is being pulled from font files and there are various code paths. I would expect that for something called "index" the value is 0 would be valid, and perhaps `stbtt_FindGlyphIndex()` should/could have be designed to return -1 on error. The only documentation I have is Sean's tweet in #575. I added asserts in `stbtt_FindGlyphIndex()` to verify that non-error return values are != 0 but I'm not hitting all return paths and either way it could depend on the font data. Perhaps `stbtt_FindGlyphIndex()` should be changed to return -1 instead of 0 (and maybe the public facing version would turn -1 into 0 for compatibility?). Or maybe 0 is always an error and then this is a non-issue.

**OPTIONAL?**

Perhaps still should be added as an option? e.g.

```STBTT_DEF void stbtt_PackSetSkipMissingGlyphs(stbtt_pack_context *spc, int skip);```

Which would set a flag in `stbtt_pack_context` and checked by the `stbtt_PackFontRanges**` function?

I'll change the PR to become that if you think it would be better this way. 

( Ref https://github.com/ocornut/imgui/issues/1703 https://github.com/ocornut/imgui/issues/1671 )

Thank you for reading and considering this.

### DIFF
--- a/stb_truetype.h
+++ b/stb_truetype.h
@@ -3967,13 +3967,17 @@ STBTT_DEF int stbtt_PackFontRangesGatherRects(stbtt_pack_context *spc, const stb
          int x0,y0,x1,y1;
          int codepoint = ranges[i].array_of_unicode_codepoints == NULL ? ranges[i].first_unicode_codepoint_in_range + j : ranges[i].array_of_unicode_codepoints[j];
          int glyph = stbtt_FindGlyphIndex(info, codepoint);
-         stbtt_GetGlyphBitmapBoxSubpixel(info,glyph,
-                                         scale * spc->h_oversample,
-                                         scale * spc->v_oversample,
-                                         0,0,
-                                         &x0,&y0,&x1,&y1);
-         rects[k].w = (stbrp_coord) (x1-x0 + spc->padding + spc->h_oversample-1);
-         rects[k].h = (stbrp_coord) (y1-y0 + spc->padding + spc->v_oversample-1);
+         if (glyph != 0) {
+            stbtt_GetGlyphBitmapBoxSubpixel(info,glyph,
+                                            scale * spc->h_oversample,
+                                            scale * spc->v_oversample,
+                                            0,0,
+                                            &x0,&y0,&x1,&y1);
+            rects[k].w = (stbrp_coord) (x1-x0 + spc->padding + spc->h_oversample-1);
+            rects[k].h = (stbrp_coord) (y1-y0 + spc->padding + spc->v_oversample-1);
+         } else {
+            rects[k].w = rects[k].h = 0;
+         }
          ++k;
       }
    }
@@ -4026,7 +4030,7 @@ STBTT_DEF int stbtt_PackFontRangesRenderIntoRects(stbtt_pack_context *spc, const
       sub_y = stbtt__oversample_shift(spc->v_oversample);
       for (j=0; j < ranges[i].num_chars; ++j) {
          stbrp_rect *r = &rects[k];
-         if (r->was_packed) {
+         if (r->was_packed && r->w != 0 && r->h != 0) {
             stbtt_packedchar *bc = &ranges[i].chardata_for_range[j];
             int advance, lsb, x0,y0,x1,y1;
             int codepoint = ranges[i].array_of_unicode_codepoints == NULL ? ranges[i].first_unicode_codepoint_in_range + j : ranges[i].array_of_unicode_codepoints[j];


nothings:
1) The semantics of index 0 are passed straight through from truetype; see https://docs.microsoft.com/en-us/typography/opentype/spec/loca.
    > By definition, index zero points to the “missing character”, which is the character that appears if a character is not found in the font. The missing character is commonly represented by a blank box or a space.

     (The intent is that the user doesn't need to know that this is the truetype definition, but as you note the return value of FindGlyphIndex hasn't documented in the first place.)

2) The intent is that 0 is returned on 'error' for glyph ids, but if you ask for a codepoint that is not present in the font, you get the index-0 box glyph. This means _it is not an error_! If you ask for a glyph that's not present in the font, you're not supposed to get nothing, you're supposed to get the fallback 'failure' glyph so that you (the programmer) don't have to do anything else. See nearly every app everywhere which has this behavior when you insert unicode characters not supported in the current font (browsers, word processors, etc). Also note that this is how stbtt works if you use the functions that are codepoint-based instead of glyph-based.

3) Requesting a codepoint be added with PackFontRanges is interpreted as 'asking for a codepoint' for the sake of (2).

Thus, getting the empty box character for each codepoint in a range that you requested is the intended behavior. The intent was that the burden was on you to only ask for characters you actually wanted. This is why PackFontRanges exists (so you can specify an arbitrary number of ranges, some/all of which could be single characters).

Now there are several possible approaches to fixing this:

1. Do nothing. The behavior as-is is "correct".
2. Share the box glyph in the atlas (i.e. still take up an stbtt_packedchar for each missing character, but only allocate one copy of the glyph per font).
3. Change the semantics
4. Add a function for batch requesting for a bunch of codepoints whether they're defined in the font (and possibly make the return value of that fill out font range structures ready to be sent to PackFontRanges).

Your solution is a form of (3) "change the semantics". I think I would prefer (2) or (4), since I prefer to maintain the semantics of 'if you ask for a codepoint you get some glyph, possibly the fallback'.

If we went with your solution, it has the issue that it conflates 'character didn't fit' and 'character wasn't present', which I think is bad. This could be addressed by adding a 'failure reason' to each packedchar struct (this wouldn't break anything since nobody would currently have been checking them, though it makes everyone pay with a small amount of memory). I would definitely want skipping-missing-chars to be optional (though it could be on a #define instead of adding a new runtime API).

(Note that technically approach (2) is orthogonal to both your-solution-if-optional and fix (4), but your solution and (4) are basically redundant.)

nothings:
Note, to clarify, the truetype specification isn't clear whether any character could legitimately use index 0 (i.e. if there were a unicode 'box' codepoint). This is part of why I passed this definition through rather than treat it as an explicit error.

I would assume the intent was "no, it can't", but since they don't explicitly forbid it, who knows. (And even if they did explicitly forbid it, who knows.)

ocornut:
Thanks for looking into it and the detailed answer.

----

Let me prefix my answer with the fact that I have came back to this again now and found a simpler user-side workaround (which involve looping twice and calling `stbtt_FindGlyphIndex` a third time. Being initialization code i'm not too fussed with the extra cost here \*.

(\* If the cost of `stbtt_FindGlyphIndex` matters, you could store the result the first time in `stbrp_rect` as it's already called twice per glyph in PackRanges.)

So my lazy work-around "only" calls it a third time and then overwrite `r->w = r->h = 0` between GatherRects and actual packing. Then after packing resets the `r->was_packed` flag so the render function can ignore them. This works because I have a custom version of the outer PackRanges function.

The second change of my PR, the line `stbtt_PackFontRangesRenderIntoRects()` that tests for non-zero width/height could possibly be safely merged:
`if (r->was_packed && r->w != 0 && r->h != 0) {`

Which would allow me to remove the second (simpler) loop, and the user code would crash or misbehave if either is zero anyway (unsigned values would underflow leading to 64k-sized glyphs etc.).

----

For the sake of pursuing this idea, but I could live with my workaround:
(you may want to just close and forget this.)

> 2. Share the box glyph in the atlas

Yes, and yes to being orthogonal to both 4 and my suggestion (since it wouldn't solve the problem for easily merging fonts that share overlapping ranges,).

> 4. Add a function for batch requesting for a bunch of codepoints whether they're defined in the font (and possibly make the return value of that fill out font range structures ready to be sent to PackFontRanges)

Yes. And possibly also make a function that returns the list of codepoints in the font? 

One advantage of my solution is that it comes with less processing/allocs. I think the error code would be stored along with `int was_packed` in `stbrp_rect` and not in `packedchar` (where all-zero positions are enough to signify "no output"). `was_packed` currently is an int, there wouldn't be an extra memory cost. But stb_rect_pack implement the same structure so I guess they may have to be updated together.

nothings:
Coming back to this again, what do you think of just having a #define that determines whether the pack functions embed glyph 0 or discard it? Still leave room for future revision where if you allow the box character (the old behavior) we implement a shared-box-character that doesn't bloat the atlas.

ocornut:
(To first clarify, my code evolved since the later May 2018 and this is a non-issue for me now. I'm first doing a pass calling stbtt_FindGlyphIndex() from which I am building the intersection of requested glyphs and available glyphs, and then passing that in array_of_unicode_codepoints. So I am not directly affected by this).

> Coming back to this again, what do you think of just having a #define that determines whether the pack functions embed glyph 0 or discard it?

I believe since we have a stbtt_pack_context structure and its initialization is opaque, it would do no harm to just have this flag in the runtime structure (as suggested in the later "OPTIONAL" part of my post).
That additional storage/check is not going to affect performance either way, technically the runtime flag makes things a little more flexible, and amount of code is going to be similar.

I can rewrite the PR to make this a compile-time or runtime option as you prefer (but it should be trivial enough you may want to just do it). I agree that either will be better than current PR.

nothings:
Yeah, I realized from the previous comment you'd worked around it, but I still value your opinion!

The only reason I was thinking compile time instead of runtime was to avoid API bloat and complexity, but given that there's already a context anyway, you're right, it's not like a single function vs a single #define is that big a deal. Given the amount of code, I can do it myself, especially if you're not even likely to use it yourself.

nothings:
Ok, there's a subtlety I missed here. You've set it up so some info is returned in the stbrp_rect, but the intent is that those are opaque, just storage for PackFont to do its work, which should never be looked at by the client. In theory any info about failed packing should be returned in the stbtt_packedchar. 

However, I'm guessing your intention here is that when you have multiple fonts with disjoint ranges, you want to use the same stbtt_packedchar, rather than one per font, so you don't want to stomp existing data that's in them? I think that's fine, we can just say you should initialize your stbtt_packedchar to 0, and then you can tell that a character failed if the results are all 0.

nothings:
Sorry, my point was correct but I was misremembering the details. If you use the broken-out functions that take stbrp_rects, then you can look at them if you want. But if you use the main function, stbtt_PackFontRanges, you never even see the stbrp_rects at all.

ocornut:
It wasn't my intention that user code would ever peek into a `stbrp_rect` structure.
The proposed PR only had stb_truetype code looking into them and eventually leading to `stbtt_packedchar` output to be zero.

I forgot it is officially documented that a zero width/height stbtt_packedchar is a failure, it was until now essentially a _packing failure_. The change would make it a packing failure OR a missing glyph failure.

ocornut:
To clarify, I believe that based on the discussion above, if you add a flag in the context structure + the corresponding function, then you can use existing PR as a base and only replace:

```if (glyph != 0) {```

with

```if (glyph != 0 || !spc->skip_missing_glyphs) {```

`char skip_missing_glyphs` being the new member, defaulting to 0.

It only makes it more evident that zero-sized packedchar are failure (they were already failures but less likely to happen so perhaps less likely to be handled by user code, which is a good reason for defaulting the new option to 0).

nothings:
Oh yeah, you're right, the packedchars are 0'd at init, I missed that.

nothings:
Anyway, yeah, your suggested change above is exactly what I had in my code (until I rewrote the test as 'glyph == 0 && spc->skip_missing' because I find the other form harder to reason about)

Issue #609, opened 2018-05-09
Unused functions warning when STBI_ONLY_PNG


dp304:
Hi! When I #define STBI_ONLY_PNG, g++ 6.3.0 gives me warnings:

warning: ‘void stbi__float_postprocess(float*, int*, int*, int*, int)’ defined but not used [-Wunused-function]
warning: ‘void stbi__vertical_flip_slices(void*, int, int, int, int)’ defined but not used [-Wunused-function]
warning: ‘int stbi__sse2_available()’ defined but not used [-Wunused-function]

nothings:
fixed in next release

PR #610, opened 2018-05-09
stb_image: fix three unused function warnings

dp304:
Some new and fixed preprocessor checks with the aim of fixing #609


nothings:
merged for next release

PR #613, opened 2018-05-22
stb_sprintf: fix return value to follow libc spec

### LINKED ISSUE #612: stb_sprintf.h: return value not according to spec
While the standard `snprintf` returns the theoretical maximum length (without zero-terminator), even when not calling it with a `NULL`  buffer and `0` length, the stb implementation returns the actual length written into the buffer (without zero-terminator).

```
char buf[8];
int len = snprintf(buf, 8, "this string is too long");
assert(len == 23); /* length of the string, without zero-terminator */
len = stbsp_snprintf(buf, 8, "this string is too long");
assert(len == 23); /* will fail, len is 7 */
```

I'll look into submitting a fix later today. Should be relatively easy by tweaking the clamp callback.
Although, I'd like to have your input, since this is a breaking change, and if somebody is depending on this behavior, a change might break code.

Leandros:
Pull request to fix #612. This is achieved by simply continuing to write into the temporary buffer, and counting the length in another variable.

### DIFF
--- a/stb_sprintf.h
+++ b/stb_sprintf.h
@@ -15,6 +15,7 @@
 //    Rohit Nirmal
 //    Marcin Wojdyr
 //    Leonard Ritter
+//    Arvid Gerstmann
 //
 // LICENSE:
 //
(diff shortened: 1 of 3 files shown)


nothings:
Had to merge this manually because I didn't want the Makefile or test_sprintf.c changes. Hope I didn't screw anything up!

PR #615, opened 2018-05-26
Update stb_tilemap_editor.h

sherjilozair:
Let's just use the proper name. More helpful for beginners.


nothings:
merged in next release

PR #621, opened 2018-06-05
stb_textedit: fix undo/redo after pasting large amount of text at once

### LINKED ISSUE #620: stb_textedit.h: undo/redo after a 32 KB+ paste fail because undo record store sizes as 'short'
Hello,

**Problem**

Undo/redo tends to break severely when pasting 32 KB+, because the `insert_length`, `delete_length`  fields of StbUndoRecord are typed `short` (values becomes negative, general havoc occurs).

This is regardless of the value stored in `STB_TEXTEDIT_UNDOCHARCOUNT` (default=999). Also note that even with a small value, the code should in theory allow to do a large paste followed by an undo, since undo-ing an insertion doesn't consume characters in the undo_char buffer.

This is the current structure:
```
typedef struct
{
   // private data
   STB_TEXTEDIT_POSITIONTYPE  where;
   short           insert_length;
   short           delete_length;
   short           char_storage;
} StbUndoRecord;
```

- Both `insert_length` and `delete_length` could probably use `STB_TEXTEDIT_POSITIONTYPE`.

- `char_storage` in theory should be derived from the value of `STB_TEXTEDIT_UNDOCHARCOUNT`, which is probably doable with a `#if STB_TEXTEDIT_UNDOCHARCOUNT < 0x10000` block. Or it could be another define parameters e.g. `STB_TEXTEDIT_UNDOCHARPOSITIONTYPE` or just `int`.
Using `STB_TEXTEDIT_POSITIONTYPE` for `char_storage` would be misleading and wrong if `STB_TEXTEDIT_UNDOCHARCOUNT > 32KB` and `STB_TEXTEDIT_POSITIONTYPE == short` _however_ (a) this is going to be rare and (B) this wouldn't be a regression from the current version, so for simplicity maybe this would be the best solution.

- In StbUndoState, `undo_char_point`, `redo_char_point` should use the same type choosen for `char_storage`.

**Suggestion**

(Easy mode)
Use `STB_TEXTEDIT_POSITIONTYPE` for all four fields.

With default STB_TEXTEDIT_UNDOSTATECOUNT==99 we are taking saving ~594 bytes per instance, I think this is the sort of memory people can afford if they are going to be interactively editing text. If people care about that sort of memory they can still tweak values and/or limits active instances.

**Patch**

- Replace 3 types in `StbUndoRecord`: insert_length, delete_length, char_storage
- Replace 2 types in `StbUndoState`: undo_char_point, redo_char_point
- Search for cast to short (`(short)`) and replace with proper type, There are 8 of them in the code.

I'm not pushing for a PR because by the time a decision is taken on the type to use it should be faster for you to just apply the three steps above, but let me know if you would prefer a PR.

Thank you.

ocornut:
Fixes #620.

Suggested line for Changelog:
> fix undo/redo from corrupting data after pasting large amount of text at once (over 32 KB)

(
I first tried to use an inferred `STB_TEXTEDIT_UNDOCHARPOSITIONTYPE` that would be either short either int, however since there is code mixing undo_char_pos and insert_length it didn't make much sense, so just switched to int and it is simpler. 

NB: The documentation incorrectly suggested the size would be:
`[4 + sizeof(STB_TEXTEDIT_POSITIONTYPE)] * STB_TEXTEDIT_UNDOSTATE_COUNT`
It should have said:
`[6 + sizeof(STB_TEXTEDIT_POSITIONTYPE)] * STB_TEXTEDIT_UNDOSTATE_COUNT`
And now it became
`[4 + 3 * sizeof(STB_TEXTEDIT_POSITIONTYPE)] * STB_TEXTEDIT_UNDOSTATE_COUNT`
)

### DIFF
--- a/stb_textedit.h
+++ b/stb_textedit.h
@@ -86,8 +86,8 @@
 //   moderate sizes. The undo system does no memory allocations, so
 //   it grows STB_TexteditState by the worst-case storage which is (in bytes):
 //
-//        [4 + sizeof(STB_TEXTEDIT_POSITIONTYPE)] * STB_TEXTEDIT_UNDOSTATE_COUNT
-//      +      sizeof(STB_TEXTEDIT_CHARTYPE)      * STB_TEXTEDIT_UNDOCHAR_COUNT
+//        [4 + 3 * sizeof(STB_TEXTEDIT_POSITIONTYPE)] * STB_TEXTEDIT_UNDOSTATE_COUNT
+//      +          sizeof(STB_TEXTEDIT_CHARTYPE)      * STB_TEXTEDIT_UNDOCHAR_COUNT
 //
 //
 // Implementation mode:
@@ -110,7 +110,7 @@
 // Symbols that must be the same in header-file and implementation mode:
 //
 //     STB_TEXTEDIT_CHARTYPE             the character type
-//     STB_TEXTEDIT_POSITIONTYPE         small type that a valid cursor position
+//     STB_TEXTEDIT_POSITIONTYPE         small type that is a valid cursor position
 //     STB_TEXTEDIT_UNDOSTATECOUNT       the number of undo states to allow
 //     STB_TEXTEDIT_UNDOCHARCOUNT        the number of characters to store in the undo buffer
 //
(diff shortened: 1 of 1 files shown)


nothings:
merged for next release

PR #623, opened 2018-06-08
Fix 1bit bmp

plzombie:
Fix to my previous pull request https://github.com/nothings/stb/pull/500.

Loading 1bit bmp now works properly if req == 4 or if s->img_x divisible by 8.
Sorry for not testing it before.


nothings:
merged for next release

Issue #622, opened 2018-06-07
Optimization of stbi__ldr_to_hdr


technik:
Current implementation of stbi__ldr_to_hdr has an if within a loop, that has become the bottleneck of loading this kind of images. The loop cam be easily broken into two parts, as shown in this compiler explorer capture:

https://godbolt.org/g/gbdU7y

that change alone, lets us put the if outside of the loop, reducing branching of the code, and letting the compiler do better optimizations. Personally, this change reduced loading time by half in one of my projects, so it's a big potential gain for a small change. Further optimizations can be made by specializing the loop for the case when  n==comp, but I didn't measure those.

I can submit a pull request for it, if the change seems reasonable.

nothings:
ldr_to_hdr capacity exists so that if you load an "LDR" image through the HDR API, you'll get reasonable behavior. It's not intended as a high-quality path.

So if performance of ldr_to_hdr is an issue for you, you really shouldn't use it. Load the image through the "ldr" API and do the conversion yourself. If you do the conversion yourself, you can:

1. Use a 256-entry lookup table so it's MUCH faster than calling pow() etc
2. Do sRGB-to-HDR conversion so it's more correct (assuming your LDR images are sRGB)

nothings:
All that being said, it's probably a reasonable change to make.

technik:
Totally agreed. I will prepare a quick pull request later, but I'll take your advice anyway.

nothings:
fixed in next release

PR #624, opened 2018-06-08
Optimize stbi  ldr to hdr

technik:
Optimize stbi__ldr_to_ldr.
Break a for loop into two parts, and extract an if out of it #622


nothings:
merted in next release

PR #625, opened 2018-06-13
stb_image: add APNG support

jcredmond:
I use stb_image in one of my projects and I needed to work with APNG files. To make my life easier I added APNG support to stb_image.h rather than having to pull in a patched libpng.

I know new formats aren't encouraged, but I'm contributing the code back just in case it is useful for someone else since the changes are not too invasive: it's essentially adding a couple of chunk types with error checking and some minor tweaks to the current PNG functions. The core PNG code has not been changed other than to remove the hardcoded assumption that the default image is the image being decoded.

I chose to pack the results so that the memory buffer returned can be free()'ed with one call and the user doesn't need to remember to free some other structures or some type of linked list. This makes it completely compatible with normal PNG files, and due to APNG being backward compatible itself the user doesn't pay for the animation data they may not use. I think this approach could be generalized for e.g. animated GIF, at the cost of adding a size_t pointer to the main API. Technical details are in the commit message and I'm open to changing around the internals.

### DIFF
--- a/stb_image.h
+++ b/stb_image.h
@@ -21,7 +21,7 @@
           avoid problematic images and only need the trivial interface
 
       JPEG baseline & progressive (12 bpc/arithmetic not supported, same as stock IJG lib)
-      PNG 1/2/4/8/16-bit-per-channel
+      PNG 1/2/4/8/16-bit-per-channel (including APNG)
 
       TGA (not sure what subset, if a subset)
       BMP non-1bpp, non-RLE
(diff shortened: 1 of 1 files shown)


jcredmond:
No problem. I use stb_image for all texture / image loading, and I much preferred this to pulling in a modified libpng. It was only a couple days work, and since stb_image has animated GIF I thought I'd formalize it in a PR just in case anyone else has a use for it.

Thanks for the great libraries!

PR #626, opened 2018-06-13
Fix double multiplication by width in write_hdr_core

troughton:
`stbiw__write_hdr_scanline` was overrunning the buffer due to multiplying by the width twice when calculating the row address.


nothings:
dupe of #572

PR #629, opened 2018-06-18
Add sizeOfBuffer argument for sprintf_s when STBI_MSC_SECURE_CRT is defined.

jmarshallstewart:


nothings:
This doesn't add a new #ifdef branch, but changes an existing one to add a new argument. Is it possible there was some version of MSVC where the old call (with one fewer arguments) is correct? This has happened with other snprintf variations in the past, so I would hate to break one. On the other hand I'm not sure how one could ever have an sprintf_s without the size of the buffer, so maybe it just never worked?

nothings:
This has been fixed in a different way. If you still have a problem with it, please report again or reopen.

jmarshallstewart:
No problems here, thanks for the fix (and for STB in general).

Issue #558, opened 2018-02-11
stb_rect_pack.h -  warning: comparison is always true due to limited range of data type [-Wtype-limits]


codeyash:
If possible please fix these warnings.

warning: In file included from c_src/stb_rect_pack.h:203:0,
warning:                  from c_src/stb_truetype.c:4:
warning: c_src/stb_rect_pack.h: In function ‘stbrp_pack_rects’:
warning: c_src/stb_rect_pack.h:559:31: warning: comparison is always true due to limited range of data type [-Wtype-limits]
warning:        STBRP_ASSERT(rects[i].w <= 0xffff && rects[i].h <= 0xffff);
warning:                                ^
warning: c_src/stb_rect_pack.h:559:7: note: in expansion of macro ‘STBRP_ASSERT’
warning:        STBRP_ASSERT(rects[i].w <= 0xffff && rects[i].h <= 0xffff);
warning:        ^
warning: c_src/stb_rect_pack.h:559:55: warning: comparison is always true due to limited range of data type [-Wtype-limits]
warning:        STBRP_ASSERT(rects[i].w <= 0xffff && rects[i].h <= 0xffff);
warning:                                                        ^
warning: c_src/stb_rect_pack.h:559:7: note: in expansion of macro ‘STBRP_ASSERT’
warning:        STBRP_ASSERT(rects[i].w <= 0xffff && rects[i].h <= 0xffff);

rygorous:
Fix for this was merged in 2019.

PR #630, opened 2018-06-21
stb_rect_pack: Silence gcc warning by removing always-true assert

rombankzero:
I removed an assert that was always true, since it was just asserting that a value was within its data type's normal numeric limits (a uint16 is always <= 0xffff). The comparison tripped gcc's `-Wtype-limits` warning (also enabled by `-Wextra`).

Related issue: #558.


nothings:
merged in next release

PR #633, opened 2018-07-03
stb_sprintf: fix wrong integer width for modifiers 'j', 'z', 't'

stefano-zanotti-88:
The 'j', 'z', 't' modifiers now behave according to the table at
http://www.cplusplus.com/reference/cstdio/printf/

A couple of new #include's were necessary for the fix.
To avoid them, the following hacks might be used, but I haven't tested them on various platforms to check that they actually work.

<pre>
sizeof(size_t)     ==  sizeof(sizeof(0))
sizeof(intmax_t)   ==  sizeof(long long)
sizeof(ptrdiff_t)  ==  sizeof((char*)0 - (char*)0)
</pre>

### DIFF
--- a/stb_sprintf.h
+++ b/stb_sprintf.h
@@ -159,6 +159,8 @@ PERFORMANCE vs MSVC 2008 32-/64-bit (GCC is even slower than MSVC):
 #endif
 
 #include <stdarg.h> // for va_list()
+#include <stdint.h> // for intmax_t
+#include <stddef.h> // for size_t, ptrdiff_t
 
 #ifndef STB_SPRINTF_MIN
 #define STB_SPRINTF_MIN 512 // how many characters per callback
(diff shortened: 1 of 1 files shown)


rygorous:
Can't just #include <stdint.h>, sadly, because we still support some pre-C99 compilers that don't have it!

Otherwise looks good, will merge next time we do a release.

stefano-zanotti-88:
You could also #include <stdint.h> if supported (based on the value of __STDC_VERSION__, maybe?), and fall back to the hacks that I mentioned in the comment in case the compiler really is pre-C99.
Something like: #if defined(__STDC_VERSION__) && __STDC_VERSION__ >= 199901L
Do you think that this could do?

rygorous:
Nope! There are targets with 64-bit registers (and 64-bit (u)intmax_t) but
32-bit pointers, size_t and ptrdiff_t. For example, Linux with the x32 ABI
on 64-bit x86, the PS3, or the Xbox 360. Admittedly, all three are becoming
less relevant with every passing year.

nothings:
size_t and ptrdiff_t are fine, but I just assumed 64-bit for intmax_t to avoid the stdint problem. if that's not a reasonable fix let me know.

fixed for next release (not a merge as I redid the changes by hand)

stefano-zanotti-88:
As far as I know, intmax_t is required to be >= 64 bit, and no compiler has 128-bit integers among their standard types (just as extensions), so the assumption is ok, also because handling 128-bit numbers would require far more extensive changes to the library, I guess.

PR #637, opened 2018-07-20
Fix unused variable warning.

tweenk:
pairSetCount is only used in an assertion. When STBTT_assert is defined to
nothing, this triggers an unused variable warning.


nothings:
Dupe of #571

PR #638, opened 2018-07-27
stb_image_write optimization for faster uncompressed writes with zlib define override

jarnoh:
I've been writing uncompressed PNG files, but I after timing my renders, I noticed that writing the actual files is still slow, even when I use zlib (with STBIW_ZLIB_COMPRESS) and set stbi_write_force_png_filter=0 and stbi_write_png_compression_level=0.  (note that for uncompressed data you want to have zlib 1.2.11, there is a huge performance benefit compared to 1.2.8)

I profiled my use case and most of the time was spent computing crc32 checksums.  stb_image_write's internal crc32 function is pretty oldschool, so I added new define STBIW_CRC32 which makes it possible to override it. zlib's crc32 function is 35x faster, and it is now simple to integrate it if you need the speed.

Another optimization is to skip loops when PNG is encoded with type=0 and just use memcpy.  Doesn't look like the current generation of compilers can optimize the switch case inside for loop, at least Xcode 9 does not do it.

With these changes, my use case (uncompressed PNG, zlib for compression and crc32) is now 4x faster, saving 4 megapixels image is down from 100ms per file to 25ms.

### DIFF
--- a/stb_image_write.h
+++ b/stb_image_write.h
@@ -923,6 +923,9 @@ unsigned char * stbi_zlib_compress(unsigned char *data, int data_len, int *out_l
 
 static unsigned int stbiw__crc32(unsigned char *buffer, int len)
 {
+#ifdef STBIW_CRC32
+    return STBIW_CRC32(buffer, len);
+#else
    static unsigned int crc_table[256] =
    {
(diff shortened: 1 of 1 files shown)


nothings:
Merged in next release, and I optimized the non-type-0 loops.

PR #640, opened 2018-08-08
Fixes crash when writing hdr files

cedricpinson:


rygorous:
We have multiple candidate fixes for that one and #644 won by virtue of being the first I looked at, so I'm closing this PR.

PR #644, opened 2018-08-17
Fix typo in stb_image_write.h

hb3p8:
There was a redundant multiply by X in stbi_write_hdr_core while computing a scan-line address.

Probably introduced in https://github.com/nothings/stb/commit/b056850ea9118b69325d973f8aef7f843527e299


rygorous:
This should fix #567.

PR #646, opened 2018-08-19
added Portable FloatMap (pfm) support:

slomp:
About PFM:
- lossless float32-based format for HDR images
- monochromatic (Y) or trichromatic (RGB) -- no alpha
- negative values permitted

Specs: http://www.pauldebevec.com/Research/HDR/PFM/

### DIFF
--- a/stb_image.h
+++ b/stb_image.h
@@ -31,6 +31,7 @@
       HDR (radiance rgbE format)
       PIC (Softimage PIC)
       PNM (PPM and PGM binary only)
+      PFM (Portable FloatMap)
 
       Animated GIF still needs a proper API, but here's one way to do it:
           http://gist.github.com/urraka/685d9a6340b26b830d49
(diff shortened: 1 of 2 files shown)


rygorous:
The effort is appreciated, but I wish you'd asked first before going through all the trouble of integrating this into stb_image, because we do _not_ currently accept patches for new image formats.

The problem is that every new format we supports adds an ongoing maintenance burden and is another potential vector for exploits. We already support too many formats, in the sense that several of the formats we do support have more bugs reported via fuzz testing than they have active users. :)

There are libraries that aim to support every image format under the sun, but stb_image isn't one of those.

slomp:
No worries, and thanks for the feedback. I'll just keep it in my fork, since it is handy for me.
I often find myself in need of a lossless float image format when working on research projects or when debugging intermediate image/vision/dnn steps, but I understand that such tasks might be tangential to what stb-image aims for.

PR #650, opened 2018-08-26
stb_gl - OpenGL/OpenGLES extension wrangler.

mchiasson:


nothings:
I don't want to have to maintain this library. You're welcome to submit it to single_file_libs, though. I recommend changing the name, though, since it's not an stb library (and I in fact have my own stb_gl.h).

mchiasson:
:-(

Issue #25, opened 2014-07-12
Convert stb_vorbis to a header only library [FEATURE REQUEST]


Symphonym:
As I know stb_image was converted from a .c file library to a .h file library I just thought I'd make it known that there's interest in having stb_vorbis as a header only library as well. Simply because it makes it so much easier to incorporate into an existing C/C++ library.

Seeing how the previous issue I posted was about some minor issues I had with mixing C and C++ files in a premake project. Whilst mixing C and C++ files should work just fine, making stb_vorbis header only would make it much easier for me to use, hopefully for others as well.

hsdk123:
I also feel this would make integration a whole lot easier. Is this still being considered?

bullno1:
Except for flipping `STB_VORBIS_HEADER_ONLY`, what else needs to be done? Optionally `static` qualify all functions based on some `define`?

nothings:
The main thing that has to be done is prefixing all global symbols with something like 'stbv_'.

BareRose:
I've submitted a PR addressing this issue 12 days ago: #653 

Let me know if there is anything missing, so far it works great in my own projects.

raysan5:
Any news on this issue? It would be really nice to have this library implementation aligned with the others stb_* libraries. I've been using @BareRose implementation for some time and it works great.

PR #653, opened 2018-09-04
Convert stb_vorbis to header-only library

BareRose:
I wanted to use stb_vorbis as it's the only public domain vorbis decoder I'm aware of, but it was in this weird sort-of-a-source-file-but-also-not state that made it unusable for me, meaning it would need to be converted to a proper header-only library. Since there was already demand for this in #25 I decided to bite the bullet and do the conversion myself based on the suggestions in that issue.

All I did was prefix everything with `stbv_` or `STBV_` respectively and add a `STB_VORBIS_STATIC` option similar to the one in stb_image. Still took me like two hours because of the sheer amount of code to go through. I also made a small fix regarding `alloca` getting redefined and breaking compilation on my MinGW.

On some other stb libraries the prefixes used for internals have two underscores, if this is preferred all you'll need to do is case-sensitively search-replace `stbv_` to `stbv__` and `STBV_` to `STBV__`. Personally I think a single underscore is sufficient here, but the decision is ultimately up to the maintainer(s).

Here's some additional things I noticed while going through the code for my prefixing:

1. The `check_endianness()` macro (now `stbv_check_endianness()`) does nothing. I suspect this is a leftover from an earlier implementation needing to do endian-specific stuff. If it really isn't needed anymore, it should be removed.

2. Using `__forceinline` seems more of a hassle than it's worth (see the issue with MinGW) and modern compilers are smart enough to inline stuff themselves anyway. Could be replaced either with `inline` if a hint to the compiler is desired, or removed entirely, thus letting the compiler make its own inlining decisions.

3. The `VORBIS_packet_comment` enum (now `STBV_VORBIS_packet_comment`) appears to be unused, possibly another leftover from something long-removed.

4. In the new header-only form the various `#ifndef` checks in the header section are likely not going to do anything. I've left them in because maybe there's a use to them that I'm not seeing, but if there isn't they probably should be removed and replaced with notices describing which functions are disabled by which defines.

I've added myself to the contributors list as "BareRose", hope that's okay.

### DIFF
--- a/stb_vorbis.h
+++ b/stb_vorbis.h
@@ -30,14 +30,14 @@
 //    Tom Beaumont       Ingo Leitgeb        Nicolas Guillemot
 //    Phillip Bennefall  Rohit               Thiago Goulart
 //    manxorist@github   saga musix          github:infatum
-//    Timur Gagiev
+//    Timur Gagiev       BareRose
 //
 // Partial history:
 //    1.14    - 2018-02-11 - delete bogus dealloca usage
 //    1.13    - 2018-01-29 - fix truncation of last frame (hopefully)
 //    1.12    - 2017-11-21 - limit residue begin/end to blocksize/2 to avoid large temp allocs in bad/corrupt files
 //    1.11    - 2017-07-23 - fix MinGW compilation 
-//    1.10    - 2017-03-03 - more robust seeking; fix negative ilog(); clear error in open_memory
+//    1.10    - 2017-03-03 - more robust seeking; fix negative stbv_ilog(); clear error in open_memory
 //    1.09    - 2016-04-04 - back out 'truncation of last frame' fix from previous version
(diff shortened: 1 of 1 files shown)


BareRose:
Looks like travis is getting thrown off by the renaming of the file. Not much I can do about this, you'll need to update your travis configuration to use the new filename.

BareRose:
@nothings Am I going to get any response here? Is there something wrong with my changes?

nothings:
This is still a valuable submission, it's just a lot of work to review so I've continued putting it off, sorry.

kerskuchen:
Using this in a unity-build with `STB_VORBIS_IMPLEMENTATION` and `STB_VORBIS_NO_CRT` defined will hammer `malloc`, `free` and `realloc` for everything under the `#include "stb_vorbis.h"` line. I would suggest defining `STB_VORBIS_MALLOC` and friends and using them instead of raw `malloc` etc. This would also be a more recognizable API for users of stb_image.h. :)

PR #661, opened 2018-10-01
hash through match in stb_compress to achieve slightly better compresion

kmar:
results (improvement over old compressor):
3.3% silesia [but 20% slower compression]
4.3% calgary
5.4% enwik8 [but 39% slower compression]
13.5% stb.h
16.4% ProggyClean.ttf
so not stellar for big files but decent for small ones, probably related to block size and/or window size

### DIFF
--- a/stb.h
+++ b/stb.h
@@ -10461,8 +10461,10 @@ static void outliterals(stb_uchar *in, int numlit)
    if (stb__out) {
       memcpy(stb__out,in,numlit);
       stb__out += numlit;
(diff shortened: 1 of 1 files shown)


nothings:
I think I intentionally didn't do this for speed reasons? I guess could make it a controllable switch.

kmar:
Yes, performance reasons was my first guess. I added stb_compress_level() for finer control, defaults to fast compression.

PR #663, opened 2018-10-12
Fixes for stb_leakcheck.h

Clownacy:
I ran into a couple of problems while trying to use this with my program:

My compiler (64-bit MinGW-w64 GCC 8.2.0) complained about not recognising `%lld` format specifiers.

The compiler also warned about non-guarding `if`s. I ignored it at first, but then I noticed my program crashed when it ran stb_leakcheck_dumpmem. Making the `if`s guard fixed that.

After that, the lib worked until I changed how it was included: instead of tacking a debug-only `#include` at the start of every file, I switched to using GCC's `-include` option to force-include it in every file. But doing that gave me errors about size_t not being defined. And after fixing that, I instead got errors about the #defines messing with stdlib.h. So to fix those I made the declaration-only part of the header include stdlib.h.

EDIT: One last "fix" was needed to make this work with my program, but I'm not sure if it's worth adding. stb_leakcheck.h was included in C++ code, but the implementation was in a C source file. To get them to play along, I had to wrap the header's contents in an `extern "C"`.

### DIFF
--- a/stb_leakcheck.h
+++ b/stb_leakcheck.h
@@ -91,19 +91,23 @@ void *stb_leakcheck_realloc(void *ptr, size_t sz, const char *file, int line)
 
 static void stblkck_internal_print(const char *reason, const char *file,  int line, size_t size, void *ptr)
 {
-#if (defined(_MSC_VER) && _MSC_VER < 1900) /* 1900=VS 2015 */ || defined(__MINGW32__)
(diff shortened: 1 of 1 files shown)


Clownacy:
I think I've figured out how to use `%zd` on MinGW-w64. Not sure if it's backwards-compatible with the original MinGW.

nothings:
merged for next release

PR #665, opened 2018-10-12
Image pnm decimal overrun

cheako:
Function stbi__pnm_getinteger can return negative numbers if the value in ascii decimal is too large.  This patch checks for values that should not be negative and fails, zeroing the value in question.


rygorous:
This one LGTM but for trivial indent issues which we can resolve when we merge.

PR #666, opened 2018-10-12
Unit tests

cheako:
This extends travis testing to include application code that produces perl's [TAP](https://testanything.org/), that is then processed by prove(an application that ships with perl).

Three example test are included that each test 4 things.  The 4 things are image pnm's ability to read and process headers and data in both pgm and ppm formats.  Of the three tests only one is implemented, binary reading when pnm's max value is 255.  The other two, reading in ascii formatted and low color count images, is either not supported or not supported correctly respectively.

The goal is to reduce or eliminate regressions and to provide targets for work currently in progress.  There is a lot of political baggage this introduces, but I'm ignoring this fact.


nothings:
Why the directory 't'?

Why are there files with a ".t" extension?

Why is the indentation random when I look at the "Files changed" tab?

cheako:
https://perldoc.perl.org/prove.html#Default-Test-Directory should answer the first few questions. 

The 3rd answer is that API code is differentiated from the test code by indentation for no reason at all.

Issue #664, opened 2018-10-12
stb_image.h pnm output is not scaled according to maxv/value.


cheako:
The code for pnm drops necessary header value.  The example feep images use smaller than 255 maximums and failure to account for this produces images that are much darker than expected.

[feep.pgm](http://netpbm.sourceforge.net/doc/pgm.html) uses a value of 15, this means that 15(0x0f) should be considered white and 14(0x0e) is 1/16th darker than white.  Though the values are copied unmodified meaning that 15 is only 15/256 brighter than black and 14 is 1/256 darker.

As this header value is not exposed there is no way to correct for this.

Hope this makes sense and I also pray that my understanding of the code is correct.

rygorous:
The "right" thing to do for the usual stb_image interface is definitely to scale the pixels to the canonical 0..255 range (as you're already doing).

PR looks reasonable (up to some formatting issues), will merge the next time we do a release.

PR #668, opened 2018-10-14
Image pnm color adjust

cheako:
See #664.

I've also implemented detection of the various pnm formats, intended to ease implementing these formats in the future.


rygorous:
See comments in #670.

PR #669, opened 2018-10-15
Image pnm ascii

cheako:
The long awaited support for ascii images and initial support for binary bitmap images.


rygorous:
See comments in #670.

PR #670, opened 2018-10-16
Image pnm decimal overrun rebase

cheako:
This is a rebase since the underlying code changed.


rygorous:
This has unfixed merge conflicts in it and touches a lot of code with no good reason that I can tell? Also there are definitely several problems in there:

1. The "<<<< HEAD" thing in there indicates something went wrong with your rebase, please check and fix it up.
2. That check for t in stb_image.h:7002 should be rewritten "t < '1' || t > '6'", I think. :)
3. Put "maxv" (renamed to "pnm_maxv") and "pnm_format" into stbi__context directly and remove stbi__pnm, that seems total overkill for two integer fields.
4. the STBI__PNM_BITMAP handling looks really wrong to me on multiple counts, and I'm pretty sure it's never been tested.
  * it assumes signed 9-bit bytes?!? (ctr counts up to 9, and there are 9 cases in the switch.)
  * no stbi_get8() on the first byte (only after ctr has counted to 9)
  * ctr needs to get reset to 0 on every row. (PBM spec: "[..] with don't care bits to fill out the last byte in the row.")
  * the handling for "case 0" assumes the MSB is the sign bit. Why are you using signed chars at all here, or reusing c for that matter? Should be "stbi_uc" (the unsigned char type returned be stbi_get8) instead.
  * Those bit masks you're using are octal literals, not binary. They are not testing the right bits!
5. lroundf is C99 and later or C++11 and later and does not exist in several compilers we support. Nor is there any real reason to use floating-point here. Just use `int adjust = (255 << 16) / p.maxv` and `out[i] = (out[i] * adjust + 0x8000) >> 16` here.

Could you fix these and submit another PR?

cheako:
We need to round floats because 16 is only an example, just as 15 or 14 and we can't use shifts in those cases.

rygorous:
The calculation I posted expands to 8.24 fixed point and then scales down, with rounding. It handles arbitrary values of maxv between 1 and 255. (Which is checked for elsewhere.)

rygorous:
Eh, sorry, 8.16 fixed point, not 8.24. My bad.

cheako:
I tried varying the precision of your equation, but it gets slightly different results on some edge cases for each.
https://onlinegdb.com/ryWybrS0m

Though I do wonder what is correct?

rygorous:
If you want to compare to a reference version, you should not compare against one using floats, since that's approximate and subject to round-off error too. In particular, your float version has the same basic issue of multiplying by a finite-precision approximation of `255/maxv`. In addition, the rounding semantics of `lroundf` and friends are somewhat biased, so they're not an ideal "best" reference either. But one thing at a time.

For an exact reference without rounding (or alternatively, rounding towards zero, which in this case is the same as truncation since all numbers in question are nonnegative), you just do the entire calculation in integers: `(x * 255) / maxv`.

For non-negative values, the rounding rule implemented by `lroundf` (towards nearest, round ties away from zero) is equivalent to adding 1/2 then truncating, i.e. `floor(x + 0.5)` (computed in infinite precision with no intermediate rounding). Still staying in exact integer form, and with even `maxv` (it usually isn't, but it's the easier case so let's do it first) this is readily expressed as `(x * 255 + maxv/2) / maxv`. To handle odd `maxv`, the easiest thing is to just scale the whole thing by another factor of two: `(x * (255*2) + maxv) / (2*maxv)`. This is an exact (no intermediate rounding ever) version of your `lroundf(x * 255.0f / maxv)` expression, but you could still argue whether it's correct, because it exhibits a slight bias: values that fall exactly on the halfway point between two representable integers will always get rounded up, which is a systematic error. Numerical computations normally use a different tie-breaker rule that guarantees exact halfway cases get rounded up and down equally as often, the most common being the round-to-nearest-breaking-ties-toward-even rule ("RTNE") - i.e. round exact halfway ties in the direction that makes the result even (LSB is 0). You could add something like that here as well but all of this really is complete overkill; practically speaking the accuracy of either your float version or the fixed-point approximation I posted is more than sufficient for this problem.

I suppose if you want to make it as precise as possible, and are concerned about cost (indeed an integer divide per pixel seems excessive), you could just compute a small 256-entry table of what byte values the inputs 0..maxv should map to. Then setting up the table can get fancy (I still wouldn't bother with round-to-nearest-even here though) but the actual per-pixel "computation" ends up being just a single table lookup.

rygorous:
Separately: why does your test say you want to support maxv up to 0xffff? Where is that coming from?

Don't take this the wrong way, but this patch series keeps moving the goalposts with every new iteration and I'm getting a bit concerned about the scope creep. Extending our max supported maxv up to 0xffff means we would need a path that reads the image as 16 bits/channel, which needs twice the memory, so either we have end up with two versions of the reading code (one that supports up to 8 bits and one for up to 16 bits) or we waste a lot of extra memory on 8-bit/channel images. It means that tricks like using a table of all possible values to remap from [0,maxv] to a canonical range (e.g. [0,255]) become unwieldy enough to be really unappealing. It means we need to deal with converting 16-bit/channel PNMs to 8-bit (when loaded through the 8-bit API), it means we need to plumb the PBM code into the `load_16`/`loadf` APIs so you can get at the unscaled values (else 16-bit/channel support isn't very meaningful); that's a fair amount of code that would all be part of the library going forward.

And if there's an actual demand to load 16-bit/channel PPMs or PGMs, all of that is justifiable, but is there? Are there 16-bit/channel PPMs or PGMs that you would like to load through stb_image but can't right now?

I'm OK with supporting ASCII PNMs, because it's a fairly small change on top of the code that is already there, and they're definitely around. I'm likewise fine with the `maxv` fix, because it likewise isn't that big and seems like the Right Thing to do. I'm not so sure about the PBM support because that code looks like it was written out of a feeling of obligation "this is in the spec too..." but then never actually tested, and things like that don't belong in stb_image, which generally intentionally skips support for rare format variants, like 12-bit JPEGs or arithmetically coded JPEGs. 16-bit/channel PNMs to me belong in that same category, and I'd much rather _not_ support it than have a lot of extra code that is never run and probably buggy.

cheako:
Thanks for letting me know more about the goals of the stb project.   What I don't want to do is write a remap that would be another obstacle for 16bit support.

cheako:
I take back my edit, 19 is an odd number so there should be at least one input that produces 127 as an output.

rygorous:
If we decide to add 16bit support later, we can worry about the remapping then; it's not a big deal either way. (Hey, it's a single integer expression.) For now, let's worry about the cases we actually support!

cheako:
I tested the code @rygorous provided and it did indeed support up to 16bit on the arch I tested.

https://onlinegdb.com/ryWybrS0m

I've finished and created a another PR.

PR #673, opened 2018-10-24
Strict aliasing rules and implementation defined shifting changes.

AdamAllison:
3 minor fixes,
1 strict alignment, I don't think the C standard requires alignment on global variables other that what is required for that type. I could be wrong, but my compiler spit out warnings, so that prompted the change

1 right shifting of something to get the sign bit is 'implementation defined', so I cast it to a u64.

1 minor casting of variable 'l' (length) to match 'pr' (precision).

Thank you,
Adam Allison

Here is want I used to test it:

#include <stdio.h>
#define STB_SPRINTF_IMPLEMENTATION 1
#define STB_SPRINTF_NOUNALIGNED 1
#define STB_SPRINTF_STATIC 1
#include "stb_sprintf.h"
/* expected is 1.01 12345 0x1.028F5CP+0 -0x1.028F5CP+0 -1.01 */
int main(void)
{
    char buffer[512];

    stbsp_snprintf(buffer, 512, "%G", 1.01F);
    puts(buffer);

    stbsp_snprintf(buffer, 512, "%d", 12345);
    puts(buffer);

    stbsp_snprintf(buffer, 512, "%A", 1.01F);
    puts(buffer);
    stbsp_snprintf(buffer, 512, "%A", -1.01F);
    puts(buffer);

    stbsp_snprintf(buffer, 512, "%G", -1.01F);
    puts(buffer);

    return 0;
}

### DIFF
--- a/stb_sprintf.h
+++ b/stb_sprintf.h
@@ -9,6 +9,7 @@
 //    Fabian "ryg" Giesen (reformatting)
 //
 // Contributors (bugfixes):
+//    Adam Allison
 //    github:d26435
 //    github:trex78
 //    Jari Komppa (SI suffixes)
--- a/stb_sprintf.h
+++ b/stb_sprintf.h
@@ -224,9 +224,15 @@ static stbsp__int32 stbsp__real_to_parts(stbsp__int64 *bits, stbsp__int32 *expo,
 
 static char stbsp__period = '.';
 static char stbsp__comma = ',';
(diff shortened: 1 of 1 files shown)


nothings:
I re-did these changes manually with a simpler way of forcing digitpair[] to have the right alignment (I hope). Nothing wrong with the way you did it, I just wanted to make the delta smaller.

In the next release.

PR #675, opened 2018-11-09
stb_sprintf: fix integer size for %ld

account-login:
`long int` may be 64 bit on some systems.


nothings:
yep, merged for next release

Issue #676, opened 2018-11-14
stb_vorbis fails to handle ogg's containing skeleton metadata 


krcroft:
The library fails to open vorbis stream that contain ogg skeleton metadata (reference: https://xiph.org/ogg/doc/skeleton.html).  

Specifically because the following check fails (as do all subsequent checks):

```c
if (f->segments[0] != 30) return error(f, VORBIS_invalid_first_page);
```

Example vorbis files with and without skeleton metadata are attached, which comply with the specification. 

- [libVorbis-1_3_6__libOgg-1_3_3-no_skeleton.ogg.gz](https://github.com/nothings/stb/files/2578783/libVorbis-1_3_6__libOgg-1_3_3-no_skeleton.ogg.gz)
- [libVorbis-1_3_6__libOgg-1_3_3-with_skeleton.ogg.gz](https://github.com/nothings/stb/files/2578784/libVorbis-1_3_6__libOgg-1_3_3-with_skeleton.ogg.gz)

To reproduce it, encode a wav file with `oggenc -k audio.wav`  or use the latest version(s) of OggDrop and toggle skeleton stream as shown: 

![oggdrop_skeleton_stb_vorbis](https://user-images.githubusercontent.com/1557255/48452605-b38e6680-e764-11e8-8aa1-c9fb7e4584e9.png)

I think the best approach would be include enough code just to sanely skip over it:

1. Detect the Ogg Skeleton's _fishead\0_ identifier in the first segment:

```
00000000  4f 67 67 53 00 02 00 00  00 00 00 00 00 00 6f 00  |OggS..........o.|
00000010  00 00 00 00 00 00 b6 13  29 e2 01 40 66 69 73 68  |........)..@fish|
00000020  65 61 64 00 03 00 00 00  00 00 00 00 00 00 00 00  |ead.............|
00000030  e8 03 00 00 00 00 00 00  00 00 00 00 00 00 00 00  |................|
00000050  00 00 00 00 00 00 00 00  00 00 00 00 4f 67 67 53  |............OggS|
00000060  00 02 00 00 00 00 00 00  00 00 68 00 00 00 00 00  |..........h.....|
00000070  00 00 a0 be 0b 2e 01 1e  01 76 6f 72 62 69 73 00  |.........vorbis.|
```

1. If it's present, handle and skip subsequent _fishbone\0_ secondary segments, which have no actual content packets.

krcroft:
[...]

PR #678, opened 2018-11-18
Add detection for Ogg skeleton metadata

krcroft:
Regarding #676: if a valid Ogg stream contains skeleton metadata, stb_vorbis will exit with `VORBIS_invalid_first_page`. 

This pull request will detect the skeleton metadata and exit with `VORBIS_ogg_skeleton_not_supported`, which is a small but helpful refinement.

### DIFF
--- a/stb_vorbis.c
+++ b/stb_vorbis.c
@@ -3579,7 +3579,7 @@ static int start_decoder(vorb *f)
    if (f->page_flag & PAGEFLAG_continued_packet)    return error(f, VORBIS_invalid_first_page);
    // check for expected packet length
    if (f->segment_count != 1)                       return error(f, VORBIS_invalid_first_page);
-   if (f->segments[0] != 30)                        return error(f, VORBIS_invalid_first_page);
+   if (f->segments[0] != 30) {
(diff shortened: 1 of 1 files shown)


stb_vorbis.c
```diff
@@ -3579,6 +3580,18 @@ static int start_decoder(vorb *f)
...
+      // check for the Ogg skeleton fishead identifying header to refine our error
+      if (f->segments[0] == 64 &&
+          getn(f, header, 6) &&
+          header[0] == 'f' &&
+          header[1] == 'i' &&
+          header[2] == 's' &&
+          header[3] == 'h' &&
+          header[4] == 'e' &&
+          header[5] == 'a' &&
+          get8(f)   == 'd')                         return error(f, VORBIS_ogg_skeleton_not_supported);
```
Daniel-Cortez: This check doesn't seem complete without the `\0` character. Can you also take that character into account?

krcroft: Got it (see 2nd commit). Thanks for the check!

nothings:
merged for next release

PR #679, opened 2018-11-23
Image pnm improvements 112318

cheako:
1. introduce unit tests
2. stb_image: pnm max value adjustment, detect format
3. stb_image: pnm add ascii image support
4. stb_image: pnm protect from decimal value overflow


cheako:
I forgot about folding the struct for pnm header fields into the core image handle struct..

nothings:
I'm concerned about adding the dependency on "prove", as I'm not clear how much value it adds in return for adding another dependency that's a PITA for Windows.

(I seem to remember starting to discuss this issue elsewhere but I can't find the PR with it.)

cheako:
Prove doesn't NEED to do much more than run the tests, so there could just be a small prove.bat that does that.  Make sure to run the compiled executables and not the perl...  if u have perl than u also have prove.

nothings:
If it's possible for you to separate this into three or four PRs, that would be helpful:

* fixes/improvements to PNM decoder
* unit tests embedded in stb_image
* external test programs and images
* integration with prove

You can skip the fourth one as I'm almost certainly going to replace that with something else, but if it makes your life simpler to keep it around that's fine, or if things don't divide cleanly along those lines you can divide them differently.

nothings:
Previous PR on the `prove` tester was #666

nothings:
Submitter never followed up on the request to split the PR.

PR #683, opened 2018-12-17
Some GIF fixes stb_image

Nuk510:
### Fix 1: Disposal method 2 clears canvas
The old code was restoring pixels from an earlier frame with disposal method 2 in this test GIF. This disposal method should instead clear the canvas.
Left: Original GIF
Right: `stb_image.h` behavior, pre-fix
test4b test4bstb

### Fix 2: First frame preserves transparency for unspecified pixels.
The old code was clearing the pixels surrounding the first frame to a background color. They should instead remain transparent.
Left: Boundaries of first frame
Center: Original GIF
Right: `stb_image.h` behavior, pre-fix
mariofallboundaries mariofall mariofallstb

### DIFF
--- a/stb_image.h
+++ b/stb_image.h
@@ -6375,8 +6375,10 @@ static stbi_uc *stbi__gif_load_next(stbi__context *s, stbi__gif *g, int *comp, i
             }
          }
       } else if (dispose == 2) { 
-	 // clear canvas to solid black transparency
-	 memset( g->out, 0x00, 4 * g->w * g->h );
+	 // clear canvas region to solid black transparency
+	 for (pi = 0; pi < pcount; ++pi)
+	    if (g->history[pi])
+	       memset( g->out + pi * 4,  0x00, 4 );
       } else {
          // This is a non-disposal case eithe way, so just 
          // leave the pixels as is, and they will become the new background


nothings:
Should g.background even be allocated anymore? It looks to me like you've removed the only use of it, unless I'm missing something. Which then makes me wonder if this fix is even correct or if it's possible you've broken other cases.

nothings:
@rygorous is going to investigate some related issues in this code at some point, so I'm assigning it to him.

nothings:
Pinging @rygorous

mosra:
Just a note from a random passerby in a hope it will be useful -- I was dealing with GIF animations recently and while looking for test files I found [dispose_bgnd.gif](https://github.com/python-pillow/Pillow/blob/master/Tests/images/dispose_bgnd.gif), [dispose_none.gif](https://github.com/python-pillow/Pillow/blob/master/Tests/images/dispose_none.gif) and [dispose_prev.gif](https://github.com/python-pillow/Pillow/blob/master/Tests/images/dispose_prev.gif) in the Pillow repository. At least the first one renders differently with stb_image than in the browser, didn't check the other two.

Biohazard90:
I have issues with GIF decoding as well and this PR unfortunately isn't quite right either.

The g->background array is not needed. This is used for dispose method 2. But instead the specification says it should be cleared to the *background color* which is g->pal[g->bgindex]. So to fix dispose method 2 one needs to copy g->pal[g->bgindex] over the array. Not use the g->background array and neither just set it to 0, like in this PR.

The PR also removes the block that begins with "if (first_frame && (g->bgindex > 0))" but I think that's wrong too. The condition should be changed to "if (first_frame && (g->bgindex > 0) && (g->eflags & 0x01) == 0)" instead.

Biohazard90:
There is also another problem, but with dispose method 3. I think to get the correct behavior, it should revert to the last non-disposable frame (dispose 0 or 1).
This is something that must be fixed in the client code, but the parameter name *two_back* should be renamed to *last_non_disposable* or something. Then the client should store the values for this parameter by checking for (dispose == 0 || dispose == 1). Hope that makes sense.

nothings:
We're going to fork animated gif support into a separate library, and this will have to be adapted when that happens.

sam20908:
> We're going to fork animated gif support into a separate library, and this will have to be adapted when that happens.

Oh cool! What's the library called?

PR #697, opened 2019-01-23
fix compiler warnings and const error

ab-cpp:
2 unused parameter warnings and a few parameters that should be 'const'.

### DIFF
--- a/stb_image.h
+++ b/stb_image.h
@@ -6336,7 +6336,7 @@ static stbi_uc *stbi__process_gif_raster(stbi__context *s, stbi__gif *g)
 
 // this function is designed to support animated gifs, although stb_image doesn't support it
 // two back is the image from two frames ago, used for a very specific disposal format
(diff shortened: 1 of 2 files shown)


nothings:
Unused parameters need to use STBI_UNUSED in the function body for C compatibility.

write_png const fix is good, but the static function shouldn't get const, it should be cast away in the caller (stb_image convention is no const in internal functions, only on API).

ab-cpp:
Thanks for your feedback ! Requested changes made.

nothings:
Great, thanks for the speedy fix! Merged for next release.

nothings:
(FWIW you put a bunch of C++ -isms in that I had to fix since it has to build as both C and C++, but no big deal.)

ab-cpp:
Aah, thanks, I forgot the C constraint (I guess const_cast<> didn't work so well :)

PR #699, opened 2019-01-29
Silence warning 'tc' may be used uninitialized in this function

NuklearBomb:


nothings:
merged in next release

PR #703, opened 2019-02-07
Fix build warning in stb_truetype.h when asserts are disabled

scramsby:


nothings:
Note to self: remove the variable entirely and move computation into assert.

nothings:
Actually, this is a dupe of #571, thanks though!

PR #704, opened 2019-02-07
Fix minor warnings emitted by PVS Studio.

ocornut:
Fix 3 warnings emitted by PVS Studio static analyzer for Visual Studio.

in stb_textedit.h:
- variable `i` is assigned twice.

in stb_truetype.h
- variable `classDefTable` assigned but never used afterwards.
Those assignments might be useful documentation in case of further modifications, and may be commented out instead.

(
FYI there are a dozen more warnings emitted by PVS Studio, the others being mostly false positive or less obvious to fix. Therefore those 3 fixes are a little superfluous in the sense that a user of PVS Studio will stumble on more. I can paste you the other warnings if you want.

PVS Studio offers free licence for open-source developers:
https://www.viva64.com/en/open-source-license/
)

### DIFF
--- a/stb_textedit.h
+++ b/stb_textedit.h
@@ -558,7 +558,6 @@ static void stb_textedit_find_charpos(StbFindState *find, STB_TEXTEDIT_STRING *s
 
    // now scan to find xpos
    find->x = r.x0;
-   i = 0;
    for (i=0; first+i < n; ++i)
       find->x += STB_TEXTEDIT_GETWIDTH(str, first, i);
 }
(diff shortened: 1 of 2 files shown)


nothings:
The stb_textedit one is already fixed locally from another PR.

The stb_truetype one look like it might be an incomplete implementation? It's not my code, so I'll have to dig into the table documentation and see what might be going on there. Leaving this PR open to stand in for an equivalent issue.

rygorous:
There was nothing missing there (see https://docs.microsoft.com/en-us/typography/opentype/spec/chapter2#classDefTbl).

Removed those assignments in the dev branch, will be in the next release.

Issue #707, opened 2019-02-10
stb_truetype 1.20 has new warnings


jamesdolan:
The comment block describing lines of code has some unintentional escape sequences that causes both Clang and GCC to throw different warnings. Both complain (with different warnings, the GCC one seems cant be disabled with a pragma for some reason) about the "Software rasterization" line where there are spaces after the backslash.

I suggest either replacing the escape characters with something else or putting non-whitespace after them to make the intentions clear.

```
// SOURCE STATISTICS (based on v0.6c, 2050 LOC)
//
//   Documentation & header file        520 LOC  \___ 660 LOC documentation
//   Sample code                        140 LOC  /
//   Truetype parsing                   620 LOC  ---- 620 LOC TrueType
//   Software rasterization             240 LOC  \                           
//   Curve tessellation                 120 LOC   \__ 550 LOC Bitmap creation
//   Bitmap management                  100 LOC   /
//   Baked bitmap interface              70 LOC  /
//   Font name matching & access        150 LOC  ---- 150 
//   C runtime library abstraction       60 LOC  ----  60
```

rygorous:
Fixed in #721.

PR #708, opened 2019-02-11
Remove trailing whitespace after backslash

lamarqua:
Remove trailing whitespace after backslash in `stb_truetype.h` that causes warnings on clang and GCC
"backslash and newline separated by space [-Wblackslash-newline-escape]"
 
Fixes https://github.com/nothings/stb/issues/707

### DIFF
--- a/stb_truetype.h
+++ b/stb_truetype.h
@@ -248,7 +248,7 @@
 //   Documentation & header file        520 LOC  \___ 660 LOC documentation
 //   Sample code                        140 LOC  /
 //   Truetype parsing                   620 LOC  ---- 620 LOC TrueType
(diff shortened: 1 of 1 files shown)


nothings:
This isn't usable, as it will cause warnings on other compilers which complain about comments ending in backslash (which is why the trailing whitespace was added originally). This is why #707 suggested eliminating the backslash entirely.

lamarqua:
This is a slight variation, I've just added some filler `...` which are visually light and should fix both cases of trailing whitespace after backslash and backslash in comment. If you prefer I can also just remove the backslashes and use `|` and `+`.

nothings:
I've just deleted the section entirely as it was out of date anyway.

PR #710, opened 2019-02-13
stb_sprintf: support for 'hh' modifier

stefano-zanotti-88:
The following code is now supported:
`#include <inttypes.h>`
`char buff[3];`
`stbsp_snprintf(buff, sizeof(buff), "%" PRIX8, (uint8_t)0xAB); // the format string expands to "%hhX"`


nothings:
I manually added the 'hh' support without removing the HALFWIDTH tracking just to minimize the changes.

PR #714, opened 2019-02-21
Support for custom stream using callbacks (i.e., C++ std::istream)

rpvelloso:
1. added stb_vorbis_open() to the API. This function receives a stream handle and four callbacks to manipulate a custom stream fgetc_cb, fseek_cb, ftell_cb, fread_cb. The stream is not closed along with decoder handler.

2. stb_vorbis_open_file()/filename()/file_section() are all using above callbacks with default values (i.e., stdio's fgetc, fseek, ftell, fread, fclose). There is an overload to mantain the existing API interface without callbacks as parameters.

3. supressed USE_MEMORY() as this is no longer necessary, the callbacks are set when a handler is created.

4. As an example, I'm using the callback open() to play from an C++ std::istream with the following callbacks supplied as parameters:

```cpp
int fgetc_cb(void *h) {
	auto stream = (std::istream *)h;
	return stream->get();
}

int fseek_cb(void *h, long int off, int dir) {
	auto stream = (std::istream *)h;
	stream->seekg(off, (std::ios_base::seekdir)dir);
	return 0;
}

long int ftell_cb(void *h) {
	auto stream = (std::istream *)h;
	return stream->tellg();
}

size_t fread_cb(void *buf, size_t len, size_t count, void *h) {
	auto stream = (std::istream *)h;
	size_t ret = stream->read((char *)buf, len*count).gcount();
	return ret == (len*count)?count:0;
}

auto stream = std::fstream("filename.ogg", std::ios::binary|std::ios::in);
auto stbvHandle = stb_vorbis_open(&stream, &err, nullptr, fgetc_cb, fseek_cb, ftell_cb, fread_cb);
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,29 +11,29 @@ by Jorge L. "VinoBS" Rodriguez, and stb_sprintf by Jeff Roberts.
 
 library    | lastest version | category | LoC | description
 --------------------- | ---- | -------- | --- | --------------------------------
-**[stb_vorbis.c](stb_vorbis.c)** | 1.14 | audio | 5462 | decode ogg vorbis files from file/memory to float/16-bit signed output
-**[stb_image.h](stb_image.h)** | 2.19 | graphics | 7462 | image loading/decoding from file/memory: JPG, PNG, TGA, BMP, PSD, GIF, HDR, PIC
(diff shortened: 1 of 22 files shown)


rpvelloso:
I have changed a single file from master branch, only: https://github.com/nothings/stb/pull/714/commits/55f2934f0551d0725137440bdaac1871806e0acc

Don't know why all those commits are showing here.

PR #715, opened 2019-02-21
Support for custom stream using callbacks (i.e., C++ std::istream)

rpvelloso:
1. added stb_vorbis_open() to the API. This function receives a stream handle and four callbacks to manipulate a custom stream fgetc_cb, fseek_cb, ftell_cb, fread_cb. The stream is not closed along with decoder handler.

2. stb_vorbis_open_file()/filename()/file_section() are all using above callbacks with default values (i.e., stdio's fgetc, fseek, ftell, fread, fclose). There is an overload to mantain the existing API interface without callbacks as parameters.

3. supressed USE_MEMORY() as this is no longer necessary, the callbacks are set when a handler is created.

4. As an example, I'm using the callback open() to play from an C++ std::istream with the following callbacks supplied as parameters:

```cpp
int fgetc_cb(void *h) {
	auto stream = (std::istream *)h;
	return stream->get();
}

int fseek_cb(void *h, long int off, int dir) {
	auto stream = (std::istream *)h;
	stream->seekg(off, (std::ios_base::seekdir)dir);
	return 0;
}

long int ftell_cb(void *h) {
	auto stream = (std::istream *)h;
	return stream->tellg();
}

size_t fread_cb(void *buf, size_t len, size_t count, void *h) {
	auto stream = (std::istream *)h;
	size_t ret = stream->read((char *)buf, len*count).gcount();
	return ret == (len*count)?count:0;
}

auto stream = std::fstream("filename.ogg", std::ios::binary|std::ios::in);
auto stbvHandle = stb_vorbis_open(&stream, &err, nullptr, fgetc_cb, fseek_cb, ftell_cb, fread_cb);
```

### DIFF
--- a/stb_vorbis.c
+++ b/stb_vorbis.c
@@ -37,7 +37,7 @@
 //    1.14    - 2018-02-11 - delete bogus dealloca usage
 //    1.13    - 2018-01-29 - fix truncation of last frame (hopefully)
 //    1.12    - 2017-11-21 - limit residue begin/end to blocksize/2 to avoid large temp allocs in bad/corrupt files
-//    1.11    - 2017-07-23 - fix MinGW compilation 
+//    1.11    - 2017-07-23 - fix MinGW compilation
 //    1.10    - 2017-03-03 - more robust seeking; fix negative ilog(); clear error in open_memory
 //    1.09    - 2016-04-04 - back out 'truncation of last frame' fix from previous version
 //    1.08    - 2016-04-02 - warnings; setup memory leaks; truncation of last frame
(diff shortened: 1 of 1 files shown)


rygorous:
Your heart's in the right place, but please don't submit big pull requests adding major features without asking first (and getting the design reviewed). In this particular instance we've gotten a very similar pull request before, and I rejected it for reasons that apply equally to this one:

https://github.com/nothings/stb/pull/456

In this particular instance, while replacing the load-from-memory paths with a generic implementation is the right direction, it is still not OK for `get8` to invoke a callback every time. A suitable replacement needs to have some buffering on the `stb_vorbis` side and only invoke a "refill" callback when its local buffer space runs out, not call back every single time.

`stb_image` already has an internal buffering layer of that sort that shows how it works: https://github.com/nothings/stb/blob/master/stb_image.h#L1448 - and while I'm mentioning stb_image, I think a pluggable IO interface should be similar or identical to stb_image's `stbi_io_callbacks` (https://github.com/nothings/stb/blob/master/stb_image.h#L356) if at all possible.

rpvelloso:
@rygorous, using callbacks the user can do the caching as he/she sees fit. IMHO caching is not a decoder responsibility. Anyway, caching is just another layer on top of this PR, I think this is a step further in the right direction.

PR #716, opened 2019-02-22
Fix recent optimisation; bring back pre-existing logic.

zellski:
It seems to me that https://github.com/nothings/stb/commit/e5d4d6fcce00ba8c4ec2ca0dec4c0d0b804d89fb#diff-469d4cb48b9965801e6c8cb7b92e46fe introduced a logic issue, by repeatedly incrementing `p`? InI now see reproducible SIGSEVs. This fix, while likely suboptimal, does at least revert the regression.


nothings:
Duplicate of #719

PR #719, opened 2019-02-23
Fix issue 706

rygorous:
Also adding a simple tester app to hopefully reduce this kind of breakage in the future.

(This test works on x64 Linux using clang 7.0.0 with no ASan warnings.)

### DIFF
--- a/stb_image_write.h
+++ b/stb_image_write.h
@@ -1476,17 +1476,13 @@ static int stbi_write_jpg_core(stbi__write_context *s, int width, int height, in
          for(x = 0; x < width; x += 8) {
             float YDU[64], UDU[64], VDU[64];
             for(row = y, pos = 0; row < y+8; ++row) {
-               int p;
-               if(row < height) {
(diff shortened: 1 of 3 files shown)


rygorous:
Fixes #706.

nothings:
FWIW, there's already a test that reads and writes images with stb_image / stb_image_write, I apparently just forgot to run it after making these changes: https://github.com/nothings/stb/blob/master/tests/image_test.c

nothings:
Oh, no, the bug is that that test doesn't write jpgs, because I never updated this after adding the jpeg writer.

rygorous:
I modeled the naming of the new test program after the existing image_test but

a) I didn't actually have a handy image to start from on hand (and either way I would prefer something self-contained)
b) the image_test doesn't exercise the vertical flipping which has been the source of problems in the past

and the new test solves both of these issues. :)

nothings:
Makes sense. (I have a whole test suite of images, but I don't have rights to put them on github unfortunately.)

PR #721, opened 2019-02-23
stb_truetype: Fix warning in comment.

rygorous:
Replace ASCII art with slightly crappier ASCII art that is not
going to make compilers complain about trailing backslashes.

Fixes issue #707.


nothings:
Those comments were so out of date and so irrelevant anyway I just deleted them.

Issue #608, opened 2018-05-08
[stb_image] Possible use of uninitialized memory


aheysse:
Within 
`load_jpeg_image`:

   // resample and color-convert
   {
      int k;
      unsigned int i,j;
      stbi_uc *output;
     ** stbi_uc *coutput[4]; **// <-- coutput is not defualt initialized

... starting line 3715 ...

            } else {
               stbi_uc *y = coutput[0]; // <-- coutput was never initialized here and is about to be derefed
               if (n == 1)
                  for (i=0; i < z->s->img_x; ++i) out[i] = y[i];
               else
                  for (i=0; i < z->s->img_x; ++i) *out++ = y[i], *out++ = 255;
            }

nothings:
It's initialized on line 3635.

aheysse:
@nothings Should have posted more info - oops. MSVC (15.6.4) warns: 
src-external\stbi\stb_image.h(3716): error C6001: Using uninitialized memory 'coutput'.

I've attached a min repro with the required MSVC build settings. I'd attempt to run on a version of MSVC >= 15.6.4. While making the min repro, I noticed there are other static code analysis warnings. 
[STBI.zip](https://github.com/nothings/stb/files/1985983/STBI.zip). In theory it could be a compiler bug, but I sort of doubt it.

nothings:
Check if initializing coutput to { 0 } fixes it.

nothings:
(PS: What do you mean by min repro? A minimal example that produces the error message, or a minimal example that crashes? The previous comment assumes you meant the errror message.)

rsachetto:
> Check if initializing coutput to { 0 } fixes it.

I've downloaded the solution and, with VS 15.4.4, I can reproduce the Warning. Initializing with {0} creates another warning:

Warning	C6011	Dereferencing NULL pointer 'y'. 	STBI	d:\sache\downloads\stbi\stbi\stb_image.h	

> (PS: What do you mean by min repro? A minimal example that produces the error message, or a minimal example that crashes? The previous comment assumes you meant the errror message.)

I think it is a  A minimal example that produces the error message, as it is only a main with exit(0)

nothings:
> Warning C6011 Dereferencing NULL pointer 'y'. STBI d:\sache\downloads\stbi\stbi\stb_image.h Line: 3718

I'll look through the code more carefully. Maybe it's a case that's broken but never been run in the wild. Or maybe it's just a bad false positive.

nothings:
[...]

PR #726, opened 2019-03-02
stb_image: Pacify some MSVC warnings.

rygorous:
Convince the compiler's dataflow analysis that yes, we are not
reading uninitialized values of coutput.

Fixes issue #608.

### DIFF
--- a/stb_image.h
+++ b/stb_image.h
@@ -3659,7 +3659,7 @@ static stbi_uc *load_jpeg_image(stbi__jpeg *z, int *out_x, int *out_y, int *comp
       int k;
       unsigned int i,j;
       stbi_uc *output;
-      stbi_uc *coutput[4];
+      stbi_uc *coutput[4] = { NULL, NULL, NULL, NULL };
 
       stbi__resample res_comp[4];
 


nothings:
FWIW I have a huge suite of changes I'm in the middle of which remove all warnings from all MSVC compilers to 2015 with /W2, as well as clang at -W2.

(I set up a thing last year to build everything with all the MSVC compilers I have installed, but it turned out they were building at default warning level, not /W2.)

rygorous:
I'm currently setting up fuzz testing. I did this mostly on Linux to start
with then noticed that apparently current Clang nightly has ASan +
libfuzzer support on Windows! It'll be a while before that's in a stable
release, but it seems to be working fine so far in my tests.

WIP branch at https://github.com/rygorous/stb/tree/image-fuzz, I'll turn
this into a PR once I've hammered on it a bit more (so far, mostly it just
runs out of memory or times out when files state huge sizes in their
header). That said the TGA loader legit takes around 49s (and over 2GB of
memory) to process a 48-byte file and that doesn't sit well with me; I have
a reworked TGA loader that's not quite as abysmally slow in progress (and
the reworking also results in a good place to put a "did we have any IO
errors? If so, bail" check) but I'm not going to turn that into a PR until
I've done a lot more testing. I want to make sure that the stuff that's in
there actually works.

-Fabian

PR #728, opened 2019-03-04
Remove trailing whitespace

akien-mga:
This is a suggestion for code quality, feel free to reject it if you think it's not worth the change (there are definitely pros and cons).

Many IDEs nowadays tend to automatically remove trailing whitespace on save, which can make for ugly commits downstream if you're not careful, and then ugly diffs the next time when you want to sync with upstream.

I suggest to remove all this trailing whitespace, which should have no impact on code formatting nor function. You can append `?w=1` to the GitHub diff view to ignore whitespace changes and confirm that this PR does only that: [diff of this PR](https://github.com/nothings/stb/pull/728/files?w=1). (I opened and saved all files in the Kate IDE with the "trim whitespace on save" option enabled.)

Pros:
- More readable/less visual noise in editors which display whitespace (e.g. `git diff`: [screenshot_20190304_133912](https://user-images.githubusercontent.com/4701338/53734003-eaf36100-3e82-11e9-8717-e7c6ef423568.png))
- More readable diffs when updating versions downstream, provided that upstream takes care not to re-add trailing whitespace

Cons:
- Needs some control to ensure that contributors don't readd trailing whitespace. Can be done via a check in a Travis build job (I can PR that if you're interested), and additionally prevented upfront with a pre-commit hook and/or a tool like `clang-format`. This can add some extra steps for contributors if they want to set up their build environment to avoid adding trailing whitespace.
- This one-time change will pollute `git blame` as this commit will be the first blame on all lines changed here. It can be worked around with `git blame -w` (ignores whitespace), but not everyone knows about the option.

Files like `.editorconfig` can provide hints to some IDEs (didn't check which ones) about the need to trim trailing whitespace or not. See e.g. https://github.com/godotengine/godot/blob/master/.editorconfig

### DIFF
--- a/stb.h
+++ b/stb.h
@@ -9,16 +9,16 @@
 
 
  ============================================================================
-   You MUST                                                                  
-                                                                             
-      #define STB_DEFINE                                                     
-                                                                             
+   You MUST
+
+      #define STB_DEFINE
+
    in EXACTLY _one_ C or C++ file that includes this header, BEFORE the
-   include, like this:                                                                
-                                                                             
-      #define STB_DEFINE                                                     
+   include, like this:
+
+      #define STB_DEFINE
       #include "stb.h"
-      
+
    All other files should just #include "stb.h" without the #define.
  ============================================================================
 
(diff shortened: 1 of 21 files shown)


nothings:
Unfortunately this dropped while I had a huge set of changes fixing warnings pending, so it's not auto-merging. If you're willing to fix it back up, I'll try to merge it very soon so it won't happen again.

akien-mga:
No problem, I'll rebase. That's another con I forgot to mention: it will add likely add conflicts with most pending PRs/feature branches.

akien-mga:
Rebased again.

As before, append `?w=1` to the GitHub diff URL to confirm that it's only whitespace changes: https://github.com/nothings/stb/pull/728/files?w=1
I simply opened all files in the Kate editor and let it strip trailing whitespace.

nothings:
Sigh, I forgot to do this right away.

nothings:
I've written a standalone tool to do this from the command-line.

PR #735, opened 2019-03-11
Suppressing compiler warnings on stb_ds.h

rsachetto:
Suppressing compiler warnings on stb_ds.h by using the return value of realloc inside an if statement


nothings:
This is fixed in a better way by #769

PR #737, opened 2019-03-12
stb_image: add png fuzz target

randy408:
This adds a fuzz target, dictionary and iphone png's for fuzzing the png reader. My plan is to eventually integrate with [OSS-Fuzz](https://github.com/google/oss-fuzz) which would continuously fuzz the code in `stb_image.h`.


randy408:
ping

randy408:
At this stage it would suffice if the initial integration through my fork was approved, this could be merged later once there is enough code coverage.

@nothings

nothings:
This doesn't compile for me -- uint8_t isn't defined in stb_png_read_fuzzer.cpp

But I'm merging it anyway.

nothings:
Hi, somebody is creating issues on bugs.chromium.org/p/oss-fuzz, presumably related to this, and I'm getting notification of the issues by the bug tracker emailed to sean@n...org, but I do not have permission to view the issues with my normal google account, which is NOT that email address. I don't have a google account for the sean@ email.

randy408:
Those issues are created automatically, it has been a while this MR was created and forgot about the Google account requirement (https://google.github.io/oss-fuzz/faq/#why-do-you-require-a-google-account-for-authentication), AFAIK there's no way around it. If you give me a Google account I'll update it (leave off the `@gmail.com` if you're worried about spam).

nothings:
nothings.org@

nothings:
Great, now I'm getting notifications to BOTH email accounts.

Seriously, why do I need to receive these at all?

randy408:
I received duplicate notifications for the existing issues, it might be temporary.

nothings:
Ok, but why do I need to be CC'd on this at all?

randy408:
These notifications are for potential security issues, I assume you already know about OSS-Fuzz (https://github.com/nothings/stb/issues/792), stb_image was [integrated](https://github.com/google/oss-fuzz/pull/3339) using the code and test images from this PR. If you don't want to be CC'd notifications can be disabled here: https://bugs.chromium.org/hosting/settings.

nothings:
Is there any reason to think those bugs are MORE of a security risk than the dozens of open issues that already exist on the github issue tracker that are already public? (Indeed, how many of the bugs on the oss-fuzz page are duplicates of already existing bugs on the github tracker?)

Did anything in this PR description explain that it was going to open up a parallel, independent issue tracking page that I am supposed to care about? (Nothing on the page you linked in the initial PR description makes that clear.)

And finally, if I don't want to be CC'd notifications about the bugs, there's one simple solution, which is to NOT PUT ME ON THE CC field. If somebody's email has to be CC'd for the system to work, then you can put your email there, because I don't want it. Your suggestion (changing the settings) doesn't make sense because the sean@ address *doesn't have a google account* so can't change those settings, whereas for the nothings.org@ I don't want to change my notification settings chromium-bugs wide. I want to be notified about bugs I interact with on bugs.chromium, just not random bugs on bugs.chromium/oss-fuzz. I can spam block them, but what the fuck.

At this point, it looks like merging this PR was a mistake, and I'm of a mind to undo it.

LindleyF:
I know little about how the alerting is set up on this, and I don't know much about what other bugs may be already tracked or how they relate in severity, but I have OWNERS on the Google-internal version of this code so I should offer my opinion.

Which is, "this test is failing, so please disable the test" is rarely the right attitude.

There are three possible outcomes of a failing test:
- A problem with the test. Disabling it may be appropriate in this case.
- A problem with the code under test which is easily fixable. Fixing it is appropriate in this case.
- A problem with the code under test which reveals fundamental problems that would take a major effort to fix. In this case, re-evaluating whether that code should be used at all is appropriate.

Too much testing is impossible, and failing tests are the best kind of tests because it means they're turning up actual issues. So I'd be against removing the fuzz testing altogether. I would be in favor of its results being reported in a way that's minimally spammy and consistent with other bug reporting on this component, if possible. In the extreme case, I'd be okay with fuzz testing remaining Google-internal and only upstreaming to github the fixes that arise from it, but that's not preferred because every difference makes pulling down updates from github harder in the future.

nothings:
> I know little about how the alerting is set up on this, and I don't know much about what other bugs may be already tracked or how they relate in severity, but I have OWNERS on the Google-internal version of this code so I should offer my opinion.
> 
> Which is, "this test is failing, so please disable the test" is rarely the right attitude.

So, I understand your impulse, but it has nothing to do with anything.

I have no problem with testing! I'm not saying anyone should disable the test.

But the reality is that I have dozens of unfixed bugs and limited time to work on them. I don't think I've fixed more than one or two bugs in the last two years myself, instead bugs get fixed primarily by PRs from other contributors. So (a) adding more friction to me dealing with this already-overwhelming open source project helps NOBODY, just takes away time from me working on it, and (b) bugs on a closed tracker that nobody else can see means nobody else is going to make PRs to fix them.

How do we compare severity of oss-fuzz bugs vs. existing bugs? Well, who knows, but for example, we can assume that they're the same average severity as existing bugs found by fuzz testing. So you could just run a search on this repo for 'fuzz'. Most people who fuzz test don't mention that they fuzz test, so this is a lower bound on the results of fuzz-testing, and in fact there are 7 existing issues that mention fuzzing. So I bet we already have 14+ bugs due to fuzz-testing. So how are 5 non-public fuzz-tested bugs on oss-fuzz worthwhile?

And now that I know how long it took, I should never have wasted my time replying to this.

randy408:
The timeout bug reported by OSS-Fuzz looks like it could consistently lead to DoS and the integer overflow bug could lead to allocating undersized buffers and possibly out-of-bounds writes, there are bugs on github with the same severity, there are no duplicates for those two. I don't think bug reports from fuzz testing belong in a public GitHub issue, they're usually more serious and less likely to be a false positive than warnings from e.g. a static analyzer.

The initial PR was created almost a year ago, the linked [README](https://github.com/google/oss-fuzz/tree/c2f88ee60f8a298762fb5c07429f0d2ec194fecd) at the time described the integration process (including the separate bug tracker) but this was moved to the documentation a few months ago, I admit that detail wasn't obvious when this was merged.

The duplicate notifications got sent when the old address was replaced with the @gmail.com one, it's probably a one-time bug and the old address should not receive any more notifications.

> If somebody's email has to be CC'd for the system to work, then you can put your email there, because I don't want it. 

My address is also on the CC list and I am intending to fix the bugs reported by OSS-Fuzz. Usually the project owner is required to have access to the private bugs (this is controlled with the CC field), I suppose removing your email from the CC field is an option while also keeping the tests running. Assuming that can be done, are you willing to keep this PR merged and accept future PRs maybe once every 1-2 months?

nothings:
> I don't think bug reports from fuzz testing belong in a public GitHub issue, they're usually more serious and less likely to be a false positive than warnings from e.g. a static analyzer.

I guarantee you there are equally serious bugs in public Github issues. This is not how security is handled on this project. That may seem irresponsible to you, but it's just how it is.

nothings:
> The duplicate notifications got sent when the old address was replaced with the @gmail.com one, it's probably a one-time bug and the old address should not receive any more notifications.

Or it's not a one-time bug, given that the wrong email address is STILL listed in the CC: field of the old bugs. It may not receive notifications for new bugs, but it hasn't actually been replaced/removed in that sense.

image

nothings:
> The initial PR was created almost a year ago, the linked [README](https://github.com/google/oss-fuzz/tree/c2f88ee60f8a298762fb5c07429f0d2ec194fecd) at the time described the integration process (including the separate bug tracker) but this was moved to the documentation a few months ago, I admit that detail wasn't obvious when this was merged.

Ok, fair.

randy408:
> Or it's not a one-time bug, given that the wrong email address is STILL listed in the CC: field of the old bugs. It may not receive notifications for new bugs, but it hasn't actually been replaced/removed in that sense.

You're right, it will only be removed for new bugs (https://github.com/google/oss-fuzz/issues/3373), if filtering the notifications is not an option I will ask the OSS-Fuzz team if fuzzing can continue with just me on the CC field.

randy408:
I'm planning on extending code coverage for this with https://github.com/google/oss-fuzz/pull/3784, this will trigger a lot of bug notifications for your gmail account, @nothings if you haven't blocked these or want your email removed entirely let me know.

PR #742, opened 2019-03-30
Removed unnecessary part of assertion

ErikWallstrom:
`v >= 0` is unnecessary since `v` is `unsigned`, which means it is guaranteed to be more than or equal to zero. It also caused warnings when compiled in gcc with `-Wall -Wextra`.


nothings:
Duplicate of #738.

PR #750, opened 2019-04-14
Add functions to extract SVG glyphs from font.

### LINKED ISSUE #748: stb_truetype: SVG support
Please can you add support for extracting SVG data from OpenType-SVG files.
I had a go at adding this myself but when I
```
t = stbtt__find_table(data, fontstart, "SVG ");
if (t) printf("off: %lx\n", ttULONG(data + t + 2));
```

The offset is always 0x0 (an illegal value), except in one font I found where it was 0xa.
For example, the Twitter Emoji font https://github.com/eosrei/twemoji-color-font/releases which definitely contains SVG font data gives me an illegal offset in the SVG table (edited; inserted wrong link originally).  otfcc https://github.com/caryll/otfcc shows the SVG data correctly.

chris-y:
The functions are:
stbtt_GetCodepointSVG - provides a pointer to the SVG data in the supplied argument, and returns the length of this data
stbtt_GetGlyphSVG - As above but takes the glyph index instead of the codepoint

These are low-level functions to get access to the data, and make no attempt to decompress it or render a glyph from it!

Fixes #748

### DIFF
--- a/stb_truetype.h
+++ b/stb_truetype.h
@@ -702,7 +702,7 @@ struct stbtt_fontinfo
 
    int numGlyphs;                     // number of glyphs, needed for range checking
 
-   int loca,head,glyf,hhea,hmtx,kern,gpos; // table locations as offset from start of .ttf
+   int loca,head,glyf,hhea,hmtx,kern,gpos,svg; // table locations as offset from start of .ttf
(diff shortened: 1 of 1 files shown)


nothings:
I did some rewriting of this so you might want to make sure I didn't break it.

PR #758, opened 2019-04-30
Fixed a memory leak in example code

Brotcrunsher:
The example code previously created a bitmap via `stbtt_GetCodepointBitmap` but never freed it, effectively creating a memory leak. This commit calls the proper stbtt free function `stbtt_FreeBitmap`.

### DIFF
--- a/stb_truetype.h
+++ b/stb_truetype.h
@@ -331,6 +331,9 @@ int main(int argc, char **argv)
          putchar(" .:ioVM@"[bitmap[j*w+i]>>5]);
       putchar('\n');
    }
+   
+   stbtt_FreeBitmap(bitmap, NULL);
+   
    return 0;
 }
 #endif 


Brotcrunsher:
As a side note: The example also never calls `fclose` to the corresponding `fopen`. This however could be seen as an irrelevant best practice thing which has nothing to do with the topic of the example code: showing stb_truetype.

nothings:
Yes, the sample code didn't do any cleanup of anything, which was consistent. It also doesn't do any error checking. I thought it was clearly supposed to be throw-away sample code, not cut-and-paste-able sample code, but clearly I was wrong, so I need to rethink this.

Brotcrunsher:
I volunteer to do this for stb_truetype. If you want this, just name me the scope of the checks:

1.) Should memory leaks be checked?
2.) Should return values be checked?
3.) Should C library functions be handled properly?
4.) Anything else that I miss?

nothings:
> 
> 
> I volunteer to do this for stb_truetype. If you want this, just name me the scope of the checks:
> 
> 1.) Should memory leaks be checked?
> 2.) Should return values be checked?
> 3.) Should C library functions be handled properly?
> 4.) Anything else that I miss?

I appreciate the volunteering, but since the sample code is embedded in the library I need to trade off all of these choices with length, so I have to do it myself I think.

PR #762, opened 2019-05-07
Remove check due to warning

BaCaRoZzo:
Since variable `v` is unsigned the first part of the check generates a warning which creates issues in our CI system. The first check is simply removed.

### DIFF
--- a/stb_image.h
+++ b/stb_image.h
@@ -5110,7 +5110,7 @@ static int stbi__shiftsigned(unsigned int v, int shift, int bits)
       v <<= -shift;
    else
       v >>= shift;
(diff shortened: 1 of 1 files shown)


Brotcrunsher:
Duplicate of: https://github.com/nothings/stb/pull/738
Dunno why it's not already in as it seems to be a reasonable change to me, probably a lack of time.

BaCaRoZzo:
Oh I see. Sorry for the duplicate. Do you think it is possible to have it merged soon? The lib is used from another lib we use and our CI system does not love to have such warnings hanging around. :sob:

nothings:
Closed as duplicate. (It'll get merged when it gets merged, there's no schedule.)

PR #763, opened 2019-05-11
Add write support for 16-bit PNG

WearerOfShoes:
This patch adds write support for 16-bit PNG.

I tried to mimic the way it is done on the read side, by having a special _16 function. I didn't modify the interface of any previously exposed functions.

Most of the functionality is shared with the 8-bit PNG writer, and I expect it to get a comparable performance and compression ratio. But I admit I haven't measured it.


stb_image_write.h
```diff
@@ -1056,7 +1063,7 @@ static void stbiw__encode_png_line(unsigned char *pixels, int stride_bytes, int
    }
 
    // first loop isn't optimized since it's just one pixel    
-   for (i = 0; i < n; ++i) {
+   for (i = 0; i < n*Bpc; ++i) {
```
Brotcrunsher: I don't know much about the png format, as such I can't comment on the quality of this code (even though at first glance it looks solid). The only minor thing that I found is that the `Bpc` parameter does not follow the naming convention of other variables/parameters. See for example the function `stbi__load_main`, it also has a `bpc` parameter with a small b.

WearerOfShoes: Thanks for reviewing. The reason for the B is exactly to differentiate it from bpc. bpc means bits per component, while Bpc means bytes per components.

I didn't want bpc/8 everywhere this is used, and also didn't want bytes_per_component everywhere.
Suggestions for better naming welcome.

WearerOfShoes: Updated pull request to have comp_bytes instead of Bpc. Purpose is to follow other namings, such as stride_bytes.

nothings:
Because this implementation supports both 8- and 16-bit channels by treating them as 1 or 2 bytes and using a multiply by the component count, it will slow down the common case of 1-byte channels unacceptably. It needs to use a separate path for the pixel loops.

WearerOfShoes:
Thanks for review, Sean.
Updated commit. Overhead *per frame* in 8-bit mode with patch vs current master:
2 mul, 1 div, 2 branches

nothings:
ok, yeah, that works too. i didn't realize it was a loop-constant that compiler MIGHT have fixed, but still worth changing.

stb_image_write.h
```diff
@@ -1062,23 +1069,32 @@ static void stbiw__encode_png_line(unsigned char *pixels, int stride_bytes, int
...
-      case 2: for (i=n; i < width*n; ++i) line_buffer[i] = z[i] - z[i-signed_stride]; break;
-      case 3: for (i=n; i < width*n; ++i) line_buffer[i] = z[i] - ((z[i-n] + z[i-signed_stride])>>1); break;
-      case 4: for (i=n; i < width*n; ++i) line_buffer[i] = z[i] - stbiw__paeth(z[i-n], z[i-signed_stride], z[i-signed_stride-n]); break;
-      case 5: for (i=n; i < width*n; ++i) line_buffer[i] = z[i] - (z[i-n]>>1); break;
-      case 6: for (i=n; i < width*n; ++i) line_buffer[i] = z[i] - stbiw__paeth(z[i-n], 0,0); break;
+      case 1: for (i=pixel_bytes; i < width*n; ++i) line_buffer[i] = z[i] - z[i-pixel_bytes]; break;
+      case 2: for (i=pixel_bytes; i < width*n; ++i) line_buffer[i] = z[i] - z[i-signed_stride]; break;
+      case 3: for (i=pixel_bytes; i < width*n; ++i) line_buffer[i] = z[i] - ((z[i-pixel_bytes] + z[i-signed_stride])>>1); break;
+      case 4: for (i=pixel_bytes; i < width*n; ++i) line_buffer[i] = z[i] - stbiw__paeth(z[i-pixel_bytes], z[i-signed_stride], z[i-signed_stride-pixel_bytes]); break;
+      case 5: for (i=pixel_bytes; i < width*n; ++i) line_buffer[i] = z[i] - (z[i-pixel_bytes>>1]); break;
```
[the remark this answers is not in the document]
jonfryd: @WearerOfShoes, I'm very interested in this, too...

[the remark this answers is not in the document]
jonfryd: @nothings: So it took me a while to figure out, but apart from the right-shift bug you pointed out, the code is actually correct with regards to `width * n` vs  `width * pixel_bytes` in the for-loops because `width` has already been multiplied by `comp_bytes` (i.e. 1 for 8-bit PNGs, 2 for 16-bit PNGs) in stbiw__write_png_to_mem_core():

```
   // all calculations in PNG are based on bytes in 16-bit mode,
   // so just treated it as a twice-as-wide 8-bit image works in most cases
   // exceptions are patches up by sending along comp_bytes
   x *= comp_bytes;
```

I also ran the tests in image_write_test and all PNG generated files are seemingly correct.

(BTW, I can make a new PR based on a fork of @WearerOfShoes work so we can finish this if you like)

Issue #733, opened 2019-03-08
stb_ds.h warning: ignoring return value of ‘realloc’, declared with attribute warn_unused_result


rsachetto:
When compiling stb_ds with gcc the compiler issues a warning: warning: ignoring return value of ‘realloc’, declared with attribute warn_unused_result [-Wunused-result]

One way to suppress this is to change the macro 

#define STBDS_FREE(x)  realloc(x,0) to 
#define STBDS_FREE(x)  if((void*)1==((void*)realloc(x,0))){;}

and change line 1450 in the function stbds_strreset to

STBDS_FREE(x) instead of using realloc(x,0);

I can make a pull request with this changes, but I am open an issue first to discuss if this code change 
is worth to include only to suppress some compiler warnings;

nothings:
what about just
  #define STBDS_FREE(x) (void) realloc(x,0)
?

rsachetto:
> what about just
> 
> ```
> #define STBDS_FREE(x) (void) realloc(x,0)
> ```
> ?

That was my first try. But it does not silence the warnings in the newer versions of gcc.

nothings:
What warning does it give?

rsachetto:
The same when using realloc(x,0)

rsachetto:
I simpler way is to change the macro to:

#define STBDS_FREE(x)  if(realloc((x),0));

It works on gcc

nothings:
using an if() like that in a macro breaks a following 'else'. embedding an else in the macro causes warnings in clang.

It looks like gcc has screwed the pooch on this, sigh. The best fix for now is to change STBDS_FREE to call stbds_free(p), and defined stbds_free to a function that calls realloc(p,0) and wraps it with whatever logic is necessary to prevent the warning. That way it's not macroified and avoids the macro issues.

rsachetto:
I´ve tested this and seems to work

```
#define STBDS_FREE(x)  stbds_free((x))

static void stbds_free(void *a)
{
  //ignore realloc return value and avoid compiler warnings
  if(realloc(a,0));
}
```

Manuzor:
Another way would be to define something like this:

```c
inline void stbds__ignore(...) { /* don't do anything here */ }
```

It can be used like this to quench that warning and is not subject to the same pitfalls as the `if(...);` solution:

```c
#define STBDS_FREE(x) stbds__ignore(realloc(x, 0))
```

This works for me at least for gcc 8.3.0. You could also use this in other situations than a void* return value.

[...]

PR #769, opened 2019-05-27
Use free() to free instead of realloc() - Fixes #768, #733

### LINKED ISSUE #768: stb_ds.h: STBDS_FREE(x) leaks memory on macos
STBDS_FREE(x) is defined as `realloc(x,0)`. While this frees `x`, on macos it also allocates new memory, that stb_ds then leaks. From the man page:

    If size is zero and ptr is not NULL, a new, minimum sized object is
    allocated and the original object is freed.

(On macos 10.13.6 (High Sierra, x64), the “minimum size” appears to 16 bytes.)

Is there a particular reason it doesn’t just use `free()`?

adurdin:
Three commits: take 1, or 1 and 2, or all:

1. Just  use free() instead of realloc(x, 0), no other changes.

2. Use STBDS_REALLOC and STBDS_FREE defines so user can override (globally, ugh).

3. Require STBDS_REALLOC and STBDS_FREE overrides in implementation only (needed stbds_free() wrapper function).

### DIFF
--- a/stb_ds.h
+++ b/stb_ds.h
@@ -232,7 +232,7 @@ DOCUMENTATION
         void sh_new_strdup(T*);
           Overwrites the existing pointer with a newly allocated
           string hashmap which will automatically allocate and free
-          each string key using malloc/free
+          each string key using realloc/free
 
       sh_new_arena
         void sh_new_arena(T*);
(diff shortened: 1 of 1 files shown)


nothings:
I added a 'context' parameter for STBDS_REALLOC/STBDS_FREE for future expansion, it's always passed NULL for now.

I removed the stbds_free() function that wrapped STBDS_FREE as I couldn't see any purpose to it.

PR #781, opened 2019-07-17
stb_ds.h: shgeti does not return -1 on missing key

sjml:
The documentation (and behavior of `hmgeti`) indicate that `shgeti` should return -1 if the key is not present in the hashmap. However, it seems to return 0 in that case. 

Here is a minimal program that displays the bug. 

```c
#include <assert.h>

#include "stb_ds.h"

int main (int argc, char *argv[]) {
    struct { int   key; int value; } *intToIntMap = NULL;
    struct { char* key; int value; } *stringToIntMap = NULL;

    assert(hmgeti(intToIntMap, 10) < 0);
    assert(shgeti(stringToIntMap, "10") < 0); // <-- assertion fails without patch
    
    hmput(intToIntMap, 10, 12);
    shput(stringToIntMap, "10", 12);
    
    assert(hmgeti(intToIntMap, 10) == 0);
    assert(shgeti(stringToIntMap, "10") == 0);
    assert(shget(stringToIntMap, "10") == 12);
}
```

I'll admit that I don't fully grok everything that's happening in stb_ds.h, but it looks like there's an off-by-one adjustment being made in `hmgeti` that is not repeated in `shgeti`. The change in this PR makes the example program run without assertion, but I haven't performed any more extensive tests than that.

### DIFF
--- a/stb_ds.h
+++ b/stb_ds.h
@@ -1,5 +1,5 @@
 /* stb_ds.h - v0.5 - public domain data structures - Sean Barrett 2019
-  
+
    This is a single-header-file library that provides easy-to-use
    dynamic arrays and hash tables for C (also works in C++).
 
@@ -26,7 +26,7 @@ COMPILE-TIME OPTIONS
   #define STBDS_NO_SHORT_NAMES
 
      This flag needs to be set globally.
-       
+
      By default stb_ds exposes shorter function names that are not qualified
      with the "stbds_" prefix. If these names conflict with the names in your
      code, define this flag.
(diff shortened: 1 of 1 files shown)


nothings:
#771/#770 was related, should probably make a pass through all the functions and make sure they're all right/consistent.

sjml:
If it would be useful, I could put together a test file to run the functions through their paces. I didn’t see any issues with NULL returns, so long as the empty map is properly set to NULL instead of uninitialized.

nakst:
I have also experienced this problem, and the fix in this pull request is the same that I discovered.

nakst:
You may also wish to wrap the "t" in brackets, as is done in hm~ functions.

phoekz:
I independently ran into the exact same problem and the exact same fix 😄.

nothings:
I fixed this independently, as the provided fix also had whitespace changes. I also added cases to trigger this to the existing unit tests. In next version.

sjml:
Thanks! Sorry for the whitespace changes; my editor can be overly aggro at times.

Issue #691, opened 2019-01-05
stb_truetype bug: gpos + kern table results in double kerning


bitc:
I finally was able to track down why my font rendering wasn't matching up with the results from other programs.

I believe that `stbtt_GetGlyphKernAdvance` has a bug. Here is the incorrect code:

```c
STBTT_DEF int  stbtt_GetGlyphKernAdvance(const stbtt_fontinfo *info, int g1, int g2)
{
   int xAdvance = 0;

   if (info->gpos)
      xAdvance += stbtt__GetGlyphGPOSInfoAdvance(info, g1, g2);

   if (info->kern)
      xAdvance += stbtt__GetGlyphKernInfoAdvance(info, g1, g2);

   return xAdvance;
}
```

If a font file has both a kern table, and a gpos table, then this will result in too much kerning being applied (exactly double, since the values in both tables are supposed to be the same).

This [stack overflow answer](https://stackoverflow.com/questions/31782261/cannot-get-the-kerning-of-some-specific-ttf-fonts-with-freetype) confirms that in the case that both tables are present, only one should be used for kerning:

> if both are present, it's up to the font renderer to select one

And here is the reference from Microsoft linked from that answer: <https://docs.microsoft.com/en-us/typography/opentype/spec/recom#39kern39-table>

I am using the font "DejaVuSans.ttf", and can verify that the results look perfect when I use either the `info->gpos` table, or the `info->kern` table, but when both results are added together as is currently done I get incorrect kerning.

Thank you

nothings:
Good work, thanks.

khaledhosny:
The second `if` should be `else if`; prefer `GPOS` if present and use `kern` otherwise.

danielmaciel:
Oh wow, I didn't know that was possible.
It seems reasonable to just add an `else` clause in there indeed.
Thanks.

danielmaciel:
Note that the `kern` table was usually much faster to lookup in my (not super thorough) tests so if it does provide exactly the same results I'd probably want to prefer `kern` instead

rygorous:
Fix for this was merged in 2019.

PR #782, opened 2019-07-18
Prefer gpos over kern table. By default some newer font tools export …

nikolaydio:
…both tables(for backwards compatiblity) and it seems we should be using only gpos if both are available.(Or otherwise we get kern values * 2)

### DIFF
--- a/stb_truetype.h
+++ b/stb_truetype.h
@@ -2540,8 +2540,7 @@ STBTT_DEF int  stbtt_GetGlyphKernAdvance(const stbtt_fontinfo *info, int g1, int
 
    if (info->gpos)
       xAdvance += stbtt__GetGlyphGPOSInfoAdvance(info, g1, g2);
-
-   if (info->kern)
+   else if (info->kern)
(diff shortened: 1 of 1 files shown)


NiLuJe:
This doesn't work if there's a GPOS table, but with data in a format that stb doesn't understand. I think Noto is a good testcase for this, IIRC.

I settled for

```
   if (info->gpos)
      xAdvance += stbtt__GetGlyphGPOSInfoAdvance(info, g1, g2);
   if (info->kern && xAdvance == 0)
      xAdvance += stbtt__GetGlyphKernInfoAdvance(info, g1, g2);
```

which looks fairly terrible, but appears to behave.

nikolaydio:
Well, your solution looks fine. I don't think it is terrible at all.

It would be nice, however, if this was patched anyhow, since the default solution to font issues is to validate the font is rendered correctly by other software and update the truetype library.

And in this case the patch is a single line, regardless of the way we go, and the issue is known since January.
https://github.com/nothings/stb/issues/691

nothings:
The if..else construct is in the newest forthcoming release.

Issue #787, opened 2019-08-05
"no SOI" error when loading some PNGs


kg:
Some PNGs look enough like a JPEG that the JPEG file format test will break with the "no SOI" error instead of rejecting the image as a JPEG. The error is in ```stbi__decode_jpeg_header``` here:
```C
   if (!stbi__SOI(m)) return stbi__err("no SOI","Corrupt JPEG");
```

```return 0``` instead of ```return stbi__err(...)``` fixes it.

rygorous:
Can you give an example file? I'd like to handle this more systematically.

-Fabian

nothings:
See also this very short twitter thread for a little discussion. https://twitter.com/nothings/status/1158523980589891584

kg:
I'm not able to reproduce this now for some reason. I don't remember modifying the image, but I did update to a newer version of stb_image (in the process of creating my PR), so I guess something unrelated could have tweaked the file parsing?

It was reproducing consistently all day yesterday (always failing on the no SOI error) and not crashing so I'm not sure what else it could be. Given that I'm OK with not fixing it if we don't have an example. I might be able to come up with one if I look at the parser long enough (is this a task for a fuzzer?)

kg:
More testing: It does actually hit no SOI, which is normal I think? That's rejecting part of the PNG header. I think what was happening is that because the error flag from that is never cleared, the 'last failure message' was no SOI even though I was hitting a different failure (a crash, probably) somewhere else in the code. That failure appears to be fixed so my codepath that was picking up the no SOI message is no longer running. If I manually run that path I get 'no SOI' again.

nothings:
final conclusion is we should just test for JPEG later in the sequence

AnterCreeper:
Hello!
During my test, I found that while I switch on the -O2 option, it will return "no SOI" and return a wrong data;
![image](https://user-images.githubusercontent.com/30188772/102620616-34f12800-4179-11eb-95dc-847a24436973.png)
![image](https://user-images.githubusercontent.com/30188772/102620723-6d910180-4179-11eb-8588-eba79c52edc2.png)
However, it will work correctly if i turn off -O2.

rygorous:
The test sequence was reordered in a fix that's now on the dev branch and will be in the next version.

pedrobagatin:
I am getting this error 

> turn off -O2

What is -O2? I am getting similar error and would like to fix it without changing library version if possible.

nothings:
[...]

PR #788, opened 2019-08-05
[prototype changeset] Indexed PNGs, configurable buffer size

kg:
Hi, this is a small set of prototype changes I wanted to submit for review. I am happy to make any formatting changes or otherwise and I can remove any changes you don't want to merge. I can also split this into separate PRs, it was just easiest to submit a single diff (and I would hope this is easier for you to review). The changes are as follows:

* Add API for loading 8-bit indexed color PNGs or TGAs without converting them to RGBA.
  Palette is stored into an input buffer provided by the caller.
* Fix erroneous SOI error when testing whether an image is a JPEG (#787).
* Allow configuring the buffer size when including stb_image.
  This improves load performance.

### DIFF
--- a/stb_image.h
+++ b/stb_image.h
@@ -85,7 +85,7 @@ RECENT REVISION HISTORY:
                                            Jeremy Sawicki (handle all ImageNet JPGs)
  Optimizations & bugfixes                  Mikhail Morozov (1-bit BMP)
     Fabian "ryg" Giesen                    Anael Seghezzi (is-16-bit query)
-    Arseny Kapoulkine
+    Arseny Kapoulkine                      Katelyn Gadd (indexed color loading)
     John-Mark Allen
     Carmelo J Fdez-Aguera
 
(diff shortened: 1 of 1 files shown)


nothings:
Can you back out the SOI changes, given the discussion in #787?

kg:
Should be gone now.

nothings:
Actually, after thinking about it, I don't think I want to take the paletted/indexed PNG changes. I don't think I like either the increased API footprint or the extra code that needs to be maintained (and really, it should be even MORE code, since all other paletted formats should do this as well).

I'm sure you have a need for it, so I guess you'll have to live with keeping a fork of it. :(

I see the 'STBI_BUFFER_SIZE' definition, but I don't see it being used anywhere?

kg:
The buffer size probably got lost in the merge, since I wrote this originally on a few-months-old version of stbi.

No worries, I agree that it adds a lot of complexity. Thanks for the review!

PR #795, opened 2019-08-18
stb_truetype.h is actual font text rasterizer

rds1983:
Just a small README.md fix


zao:
@rds1983 Sean processes PRs in batches every now and then, it's normal for PRs to be inert for a long while in this repository.

nothings:
Yeah, the issue here was that the README.md is actually a generated file, so fixing it there didn't actually fix it correctly, I had to update the files it's generated from

rds1983:
I see.

PR #808, opened 2019-09-30
ckjpr1

ChiraagKakar:
ckpr1


nothings:
I'm merging this, but it's not actually consistent with how the library currently does kerning, which is now a combination of two tables, based on some previous PR.

PR #811, opened 2019-09-30
C++14

KoulickS:
Programming is life


nothings:
These have a lot of conflicts with current version of stb.h (to be released).

rygorous:
Also introduces some other problems with typecasts we'd have to fix later, and finally stb.h is now deprecated; closing this PR.

PR #818, opened 2019-10-04
Suppress a warning about always true comparison

nsauzede:
With nothings/stb 052dce1, in stb_image.h, I get this warning :
```
../stb/stb_image.h: In function ‘int stbi__shiftsigned(unsigned int, int, int)’:
../stb/stb_image.h:5114:18: error: comparison of unsigned expression >= 0 is always true [-Werror=type-limits]
 5114 |    STBI_ASSERT(v >= 0 && v < 256); // v is unsigned int, so '>= 0' is always true
      |                ~~^~~~
../stb/stb_image.h:5114:4: note: in expansion of macro ‘STBI_ASSERT’
 5114 |    STBI_ASSERT(v >= 0 && v < 256); // v is unsigned int, so '>= 0' is always true
      |    ^~~~~~~~~~~
```
Using this compiler :
```
$ g++ --version
g++ (Clear Linux OS for Intel Architecture) 9.2.1 20191001 gcc-9-branch@276412
Copyright (C) 2019 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

This trivial, one liner, PR fixes just that (but leaves the commented test, to not lose the original intent)

### DIFF
--- a/stb_image.h
+++ b/stb_image.h
@@ -105,7 +105,7 @@ RECENT REVISION HISTORY:
     Oriol Ferrer Mesia      Josh Tobin         Matthew Gregan     github:phprus
     Julian Raschke          Gregory Mullen     Baldur Karlsson    github:poppolopoppo
     Christian Floisand      Kevin Schmidt      JR Smith           github:darealshinji
-    Blazej Dariusz Roszkowski                                     github:Michaelangel007
+    Blazej Dariusz Roszkowski Nicolas Sauzede                     github:Michaelangel007
 */
 
 #ifndef STBI_INCLUDE_STB_IMAGE_H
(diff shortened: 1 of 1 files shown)


nsauzede:
Hello,
any chance to review this trivial PR ?
It causes issues when building applications using the great stb_image, with `-Wall -Wextra -Werror`

mupfelofen-de:
I agree, this compile-time assertion is kinda pointless and can safely be removed. It would even be appropriate here to limit the data type to a size of 8 bits (instead of checking if the value goes beyond that).

zao:
I would like to mention that PRs in this repo are processed in bulk when Sean gets around to it, it's quite normal to take time and shouldn't be considered a problem with the PR itself.

nsauzede:
aah ok, sorry for the newbie question ;-)
and thanks for the information

nsauzede:
Hello ?

nothings:
> 
> 
> I would like to mention that PRs in this repo are processed in bulk when Sean gets around to it, it's quite normal to take time and shouldn't be considered a problem with the PR itself.

nothings:
dupe

PR #819, opened 2019-10-09
Fix bare backslash n as being a valid whitespace character.

cnlohr:
The current implementation has trouble with the `#inject` feature on *nix systems.  Didn't add name to authors list because it was only one line.


cnlohr:
I have never made a pull request to STB.  (ok feels a little weird to use his name) but, I wanted to know if I did everything here correctly.

nothings:
Yes, looks good, I process PRs in batches every few months, so it may be a while.

cnlohr:
Thank you!

PR #828, opened 2019-10-26
Update stb_vorbis.c

rnkjoshi:
these change may prevent value of variable val from changing before its use.


nothings:
Not a bug.

PR #832, opened 2019-11-11
stb_include: fix stb_include_strings iteration

sp0lsh:
Upper bound was incremented each loop instead of iterator causing endless loop when called.


nothings:
haha, guess i never actually used that part

PR #834, opened 2019-11-16
Quell -Wcast-align warnings from Clang

a-e-k:
The `stbi__sbraw()` macro in stb_image_write.h causes Clang to spew about 24 warnings complaining that "cast from '`unsigned char *`' to '`int *`' increases required alignment from 1 to 4" when compiled with the `-Wcast-align` option.

In practice, this is spurious so long as `STBIW_MALLOC()` and` STBIW_REALLOC()` follow the usual alignment semantics for `malloc()` and `realloc()` in that they align sufficiently for any built-in type.

To quell the warning, we can cast through a `void` pointer as an intermediary.


nothings:
since it's inside an implementation macro i guess it's fine

a-e-k:
Thanks!

PR #849, opened 2019-11-27
Updates to snprintf to handle char32/char16

AdamAllison:
Added support for char32, char16 as the destination buffer,
   Added '%S' support (char16 string)
   Added '%lS' support (char32 string)

created vsprintfcbAny
change vsprintfcb to use vsprintfcbAny
duplicated snprintf functions with 'c16/c32'

created 'stbsp__push_char' macro to handle the old *bf++ = c

All original signatures are the same,
I only erase the types at the lowest level. 
    ie *char**-> *void** , or 
   *stbsp__char32** -> *void**
   *stbsp__char16** -> *void**

I don't know if the older stats are impacted. any compiler should see
that size_per_char is either 1, or sizeof(char16/char32) and doesn't change during the function

I haven't had any problems on testing on my end.
The whole thing has to be enabled with STB_SPRINTF_MULTICHAR 1

Thanks,
Adam

### DIFF
--- a/stb_sprintf.h
+++ b/stb_sprintf.h
@@ -80,28 +80,49 @@ int stbsp_vsprintfcb( STBSP_SPRINTFCB * callback, void * user, char * buf, char 
 void stbsp_set_separators( char comma, char period )
   Set the comma and period characters to use.
 
-== wchar_t
-int stbsp_sprintfW( wchar_t * buf, char const * fmt, ... )
-int stbsp_snprintfW( wchar_t * buf, int count, char const * fmt, ... )
-int stbsp_vsprintfW( wchar_t * buf, char const * fmt, va_list va )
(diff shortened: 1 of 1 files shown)


nothings:
I have to think about this because it's a pretty big change.

AdamAllison:
No problem, let me know if you have questions.

nothings:
I think we don't want to take on this complexity. If you want to fork it and maintain it, feel free!

We probably want to add support for printing wchar_t strings TO char*, i.e. doing UTF16/32 -> UTF-8 conversion, but I think we only want to support UTF-8 output. Once we add those, you could always use the callback function and do char -> char32/char16 conversion there.

AdamAllison:
No problem,
An interesting side note wchar_t: is an int in gcc, and a short in windows.

Thank you,
Adam

PR #882, opened 2020-01-19
Fix for buffer bugs in stb__splitpath_raw and stb_strncpy

bcollins526:
The stb__splitpath_raw function was incorrectly referring to buffer instead of path while parsing, and was passing the size of the buffer pointer (4 or 8 bytes) rather than the size of the buffer itself, which was causing a crash. Ideally the public interface would take the size of the destination buffer as a parameter also, but in lieu of that, changed to use stb_strncpy because it appeared to already be dealing with this by just assuming target buffer size is n+1.

Also, stb_strncpy had an off-by-1 bug that was truncating the last valid character of the string.

### DIFF
--- a/stb.h
+++ b/stb.h
@@ -2009,7 +2009,7 @@ char *stb_trimwhite(char *s)
 char *stb_strncpy(char *s, char *t, int n)
 {
    stb_p_strncpy_s(s,n+1,t,n);
-   s[n-1] = 0;
+   s[n] = 0;
    return s;
 }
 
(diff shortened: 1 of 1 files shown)


stb.h
```diff
@@ -2456,8 +2457,7 @@ static char *stb__splitpath_raw(char *buffer, char *path, int flag)
    }
 
    if (len) { stb_p_strcpy_s(buffer, sizeof(buffer), "./"); return buffer; }
```
nothings: missed one! :)

PR #890, opened 2020-02-01
Fix clang unused function compile warnings when compiling with a single decoder

BradleyMarie:
When building with clang and STB_PNG_ONLY, I got a compile warning about unused functions. This PR fixes this warning. For good measure, I also repeated the same process with all of the following build options:

STBI_ONLY_JPEG
STBI_ONLY_PNG
STBI_ONLY_BMP
STBI_ONLY_PSD
STBI_ONLY_TGA
STBI_ONLY_GIF
STBI_ONLY_HDR
STBI_ONLY_PIC
STBI_ONLY_PNM 

The conventions for #ifdef-ing out a block of code seemed to vary throughout the file, so I did my best to match my #ifdefs with the surrounding ones.


nothings:
That's gonna be a nightmare to maintain, probably should find a better way to do this in the future, but it'll do for now.

PR #891, opened 2020-02-02
Create separate LICENSE file

Croydon:
The text is copied without modifications from https://github.com/nothings/stb/blob/2f870b3c07e477a5424bc5cc7dc15f90e484ce8f/stb.h#L14415

This makes it a lot easier to find, access and copy LICENSE information.

I'm specifically interested in this as I'm packaging stb for the C/C++ package manager Conan

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,37 @@
+This software is available under 2 licenses -- choose whichever you prefer.
+------------------------------------------------------------------------------
+ALTERNATIVE A - MIT License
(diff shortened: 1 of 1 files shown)


nothings:
It's a problem if I ever want to release a library under a different license, which could happen e.g. if I have to start from some pre-existing library and use its license.

Croydon:
If this happens, then you could still modify the LICENSE file further with the new information or delete it, if absolutely necessary

nothings:
fair

PR #893, opened 2020-02-03
Fix a small typo in the __STDC_VERSION__ constant name

NiLuJe:
Introduced in today's update ;).

I haven't checked if it creeped into other libs, FWIW ;).

(This tripped a `-Wundef` warning).

### DIFF
--- a/stb_image.h
+++ b/stb_image.h
@@ -473,7 +473,7 @@ STBIDEF void stbi_set_flip_vertically_on_load(int flag_true_if_should_flip);
 // as above, but only applies to images loaded on the thread that calls the function
 // this function is only available if your compiler supports thread-local variables;
 // calling it will fail to link if your compiler doesn't
-#if __cplusplus >= 201103L || __STDC_VERSION_ >= 201112L || defined(__GNUC__) || defined(_MSC_VER)
+#if __cplusplus >= 201103L || __STDC_VERSION__ >= 201112L || defined(__GNUC__) || defined(_MSC_VER)
 STBIDEF void stbi_set_flip_vertically_on_load_thread(int flag_true_if_should_flip);
 #endif
 
@@ -576,7 +576,7 @@ STBIDEF int   stbi_zlib_decode_noheader_buffer(char *obuffer, int olen, const ch
 #ifndef STBI_NO_THREAD_LOCALS
    #if __cplusplus >= 201103L
       #define STBI_THREAD_LOCAL       thread_local
-   #elif __STDC_VERSION_ >= 201112L
+   #elif __STDC_VERSION__ >= 201112L
       #define STBI_THREAD_LOCAL       _Thread_local
    #elif defined(__GNUC__)
       #define STBI_THREAD_LOCAL       __thread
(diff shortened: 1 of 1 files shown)


nothings:
Wait, if that triggers an undef warning, why doesn't __cplusplus or the other tests?

NiLuJe:
`__cplusplus` does too (at least with a C compiler), but I'm not sure if there's a way around that?

NiLuJe:
Actual GCC output:

```
stb_image.h:476:5: warning: "__cplusplus" is not defined, evaluates to 0 [-Wundef]
 #if __cplusplus >= 201103L || __STDC_VERSION__ >= 201112L || defined(__GNUC__) || defined(_MSC_VER)
     ^~~~~~~~~~~
```

NiLuJe:
Something like that, maybe?

```diff
diff --git a/stb_image.h b/stb_image.h
index 8292642..5f9af10 100644
--- a/stb_image.h
+++ b/stb_image.h
@@ -473,7 +473,7 @@ STBIDEF void stbi_set_flip_vertically_on_load(int flag_true_if_should_flip);
 // as above, but only applies to images loaded on the thread that calls the function
 // this function is only available if your compiler supports thread-local variables;
 // calling it will fail to link if your compiler doesn't
-#if __cplusplus >= 201103L || __STDC_VERSION__ >= 201112L || defined(__GNUC__) || defined(_MSC_VER)
+#if (defined __cplusplus && __cplusplus >= 201103L) || __STDC_VERSION__ >= 201112L || defined(__GNUC__) || defined(_MSC_VER)
 STBIDEF void stbi_set_flip_vertically_on_load_thread(int flag_true_if_should_flip);
 #endif
 
@@ -574,7 +574,7 @@ STBIDEF int   stbi_zlib_decode_noheader_buffer(char *obuffer, int olen, const ch
 #endif
 
 #ifndef STBI_NO_THREAD_LOCALS
-   #if __cplusplus >= 201103L
+   #if (defined __cplusplus && __cplusplus >= 201103L)
       #define STBI_THREAD_LOCAL       thread_local
    #elif __STDC_VERSION__ >= 201112L
       #define STBI_THREAD_LOCAL       _Thread_local
```

(Fair warning: 5AM brain, I may be spouting complete nonsense).

nothings:
Yes, you can use defined() to avoid it, I didn't bother doing this because I had recently read the C99 spec that just said 'undefined macros have value 0' without anything about issuing diagnostics for it, so I didn't realize modern compilers were doing that.

But of course they are.

Also nobody flagged it when I asked for a code review on twitter. :(

nothings:
Anyway, I'll fix that, but it's separate from this PR.

nothings:
Here's my fix:

    // as above, but only applies to images loaded on the thread that calls the function
    // this function is only available if your compiler supports thread-local variables;
    // calling it will fail to link if your compiler doesn't
    STBIDEF void stbi_set_flip_vertically_on_load_thread(int flag_true_if_should_flip);

:)

nothings:
Released a fix, let me know if it doesn't work.

NiLuJe:
Yup, that did the trick ;).

PR #896, opened 2020-02-03
fix test_sprintf.c

wojdyr:
Thanks for merging the test_sprintf PR yesterday. I'd like to add three commits:

* fix compilation (CHECK1 -> CHECK2)
* update expected return values from stbsp_snprintf - it now returns the same value as snprintf, i.e. the number of characters that would be printed if the buffer was big enough
* if `TEST_SPRINTF` is defined include the implementation (I hope it doesn't affect other possible uses of test_sprintf.c), so that this test only can be run with:

      cc -I. -DTEST_SPRINTF tests/test_sprintf.c && ./a.out

### DIFF
--- a/tests/test_sprintf.c
+++ b/tests/test_sprintf.c
@@ -3,8 +3,6 @@
 
 #if USE_STB
 # include "stb_sprintf.h"
-//# define STB_SPRINTF_IMPLEMENTATION
-# include "stb_sprintf.h"
 # define SPRINTF stbsp_sprintf
 # define SNPRINTF stbsp_snprintf
 #else
@@ -44,6 +42,12 @@ typedef ptrdiff_t ssize_t;
 #define CHECK1(str, v1                                ) { int ret = SPRINTF(buf, v1                                ); CHECK_END(str); }
 
 #ifdef TEST_SPRINTF
+
+#if USE_STB
+# define STB_SPRINTF_IMPLEMENTATION
+# include "stb_sprintf.h"
+#endif
+
 int main()
 {
    char buf[1024];
--- a/tests/test_sprintf.c
+++ b/tests/test_sprintf.c
@@ -111,10 +111,10 @@ int main()
 #if __STDC_VERSION__ >= 199901L
 #if USE_STB
    CHECK4("Inf Inf NaN", "%g %G %f", INFINITY, INFINITY, NAN);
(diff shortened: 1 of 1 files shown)


nothings:
Yeah, this is what I meant about breaking the semantics and possibly making this a separate implementation. I link all the test_ files together in a single test application, one of which already defines STB_SPRINTF_IMPLEMENTATION, so it breaks my existing codebase to do this. So if you want a standalone test, define a new file instead of piggybacking off of an existing file that intentionally doesn't have the implementation.

nothings:
I fixed the first two issues manually, hopefully correctly.

wojdyr:
The compilation should not be broken, this PR moves STB_SPRINTF_IMPLEMENTATION to be guarded by `#ifdef TEST_SPRINTF`.
But since it breaks the semantics I can move it to another file (and leave only the #include statement in `test_sprintf.c`, as it was before?).
Can you propose a new filename?

nothings:
Putting it in ifdef TEST_SPRINTF doesn't address my issue. I compile all the test_* files into one program, which includes test_c_compilation.c which already does STB_SPRINTF_IMPLEMENTATION. Then I flip on one of the TEST_SPRINTF or TEST_TRUETYPE or etc. flags to choose which main() I get. You could put it inside ANOTHER flag, like "STANDALONE_SPRINTF", and have that define STB_SPRINTF_IMPLEMENTATION, or you can put it in another file.

The name can be anything really, as long as it's not an existing file in the directory. "sprintf_test.c" or whatever.

It's still useful for me to have the tests in my test app as well, maybe the "standalone" test can #include "test_sprintf.c"? Or maybe better to put the guts into "test_sprintf.inl" and include it from both C files.

nothings:
Note that "I fixed the first two issues manually" means I released a new version of it.

wojdyr:
If the the tests are useful where they are, I can simply compile it with `-D STB_SPRINTF_IMPLEMENTATION`.
For this, I'd need one of the #includes removed from test_sprintf.c:
```c
# include "stb_sprintf.h"
//# define STB_SPRINTF_IMPLEMENTATION
# include "stb_sprintf.h"
```

Thanks for maintaining these libraries!

nothings:
checked in the removed redundant #include

PR #918, opened 2020-02-27
fix integer arithmetic in stbi__zexpand()

randy408:
Adds overflow checks, changing the variable type to unsigned simplifies the patch, fixes: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=20572

The testcase is linked in the report, should I include it here?


randy408:
This PR is obsolete, the fix is included in https://github.com/nothings/stb/pull/960.

PR #922, opened 2020-03-15
Add #undef BUFFER_SIZE to avoid multiple definitions

czarnota:
When including stb_vorbis.c in other C file that also defines `BUFFER_SIZE,` you get a warning:

```
warning: "BUFFER_SIZE" redefined
```

It is a quite common name. This is simply fixed by `#undef BUFFER_SIZE`


nothings:
stb_vorbis.c is not intended to be #included

Issue #923, opened 2020-03-15
stb_ds: according to docs arraddn should return pointer but it returns an offset


michal-z:
To work as documented it should look like:
  (stbds_arrmaybegrow(a, n), stbds_header(a)->length += (n), **&(a)[stbds_header(a)->length - (n)]**)

instead of:
  (stbds_arrmaybegrow(a, n), stbds_header(a)->length += (n), **stbds_header(a)->length - (n)**)

rygorous:
This is fixed in the dev branch; also there are two versions, "addnptr" and "addnindex", that return a pointer and index respectively. The old (ambiguous) arraddn name is deprecated and should not be used.

PR #932, opened 2020-03-24
Fix arraddn returning index instead of pointer

HeroicKatora:
The documentation of that operation already said:
> Returns a pointer to the first uninitialized item added.

This also makes a lot of sense, allowing easy initialization. But the
implementation returned the index of the first uninitialized element
instead.

### DIFF
--- a/stb_ds.h
+++ b/stb_ds.h
@@ -397,6 +397,8 @@ CREDITS
 #define arrfree     stbds_arrfree
 #define arraddnptr  stbds_arraddnptr
 #define arraddnoff  stbds_arraddnoff
+// deprecated
+#define arraddn     stbds_arraddn
 #define arrsetlen   stbds_arrsetlen
 #define arrlast     stbds_arrlast
 #define arrins      stbds_arrins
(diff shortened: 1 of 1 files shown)


nothings:
Fixes #923.

nothings:
Both behaviors are useful and you can't easily implement one with the other, so the ideal solution is probably to have arraddn() which returns the index and arraddnp() which returns the pointer, or some such.

HeroicKatora:
I find a replacement for the current functionality more trivial: It is essentially calling `arrlen` prior to `arraddn` or it can be recalculated by subtracting the newly returned pointer from the new base. In comparison, getting a pointer to the new element _must_ be done afterwards as the array could have reallocated and requires pointer arithmetic. Both are subjectively more error prone.

nothings:
My point is that both of these are broken:
```
   item *newitem_ptr = arr + arraddn(arr,n);
   ptrdiff_t offset = arraddnp(arr,n) - arr;
```
because in both cases if the array relocates the computation may be wrong. So my preference is to provide both functions so people don't try to do either of these. You're correct that one is easier to synthesize, but, *shrug*.

HeroicKatora:
I didn't quite think anyone would write the second line but probably better to be cautious. But the naming of `arraddn` vs. `arraddnp` also doesn't seem quite distinct enough, very easy to miss, and wouldn't address that the current documentation may have already some buggy use. Maybe we can come up with slightly more distinct and descriptive names?

nothings:
`arraddnptr`
`arraddnoff`
?

nothings:
Great.

As loathe as I am to keep expanding the API, for the sake of not breaking existing code, let's also add an "arraddn()" which is explicitly cast to void so you can't use the return value. It doesn't need to be documented, just there to keep existing code that *doesn't* use the return value from breaking. So just a comment "// deprecated" on the #define is all that's needed.

nothings:
Note that I renamed `arraddnoff` to `arraddnindex`, but you probably weren't using that one anyway.

shachaf:
> Note that I renamed `arraddnoff` to `arraddnindex`, but you probably weren't using that one anyway.

It looks like the rename wasn't quite done -- `arraddnindex` is defined to `stbds_arraddnindex`, but the stbds_ version is still called `stbds_arraddnoff`.

PR #934, opened 2020-03-24
stb_image: fix assert failing when loading BMP

### LINKED ISSUE #897: Assertion `info.offset == (s->img_buffer - s->buffer_start)' failed  for particular .bmp
stb fails to read some .bmp images: 
/stb/stb_image.h:5327: void* stbi__bmp_load(stbi__context*, int*, int*, int*, int, stbi__result_info*): Assertion `info.offset == (s->img_buffer - s->buffer_start)' failed

Code snippet:
```

        int tmpWidth;
        int tmpHeight;
        int tmpChannels;

        using StbImageDataPtr = std::unique_ptr<unsigned char, decltype(&stbi_image_free)>;
        StbImageDataPtr stbData(stbi_load($filePath, &tmpWidth, &tmpHeight, &tmpChannels, 0), &stbi_image_free);
```
Image attached: 
[grace_hopper.zip](https://github.com/nothings/stb/files/4148088/grace_hopper.zip)

### LINKED ISSUE #909: .BMP file crashes when loaded via stbi_load_from_memory(), loads fine via stbi_load() [image attached]
I found a .BMP file which crashes when loaded via `stbi_load_from_memory()` with a failed assertion while it loads fine via `stbi_load()`. This is the code I used for loading from memory:

```
#include "stb/stb_image.h"

#include <assert.h>

int main(int argc, const char **argv) {
    int w = 0;
    int h = 0;
    int n = 0;
    char *imgencodedfile = NULL;

    FILE *f = fopen("modelpack11/forktruck.bmp", "rb");
    assert(f != NULL);

    fseek(f, 0L, SEEK_END);
    int32_t size = ftell(f);
    fseek(f, 0L, SEEK_SET);
    assert(size > 0);

    imgencodedfile = malloc(size);
    assert(imgencodedfile != NULL);
    int result = fread(imgencodedfile, 1, size, f);
    assert(result == size);

    fclose(f);
    
    unsigned char *data = stbi_load_from_memory(
        (unsigned char *)imgencodedfile, size, &w, &h, &n, 4
    );
    return 0;
}
```

This is the resulting crash:
```
stb/stb_image.h:5327: stbi__bmp_load: Assertion `info.offset == (s->img_buffer - s->buffer_start)' failed.
```

This is the zipped up "modelpack11" folder which contains the affected `modelpack11/forktruck.bmp` image (zipped as a whole to retain the license file, it's someone else's image):

[modelpack11.zip](https://github.com/nothings/stb/files/4210956/modelpack11.zip)

zturtleman:
This fixes two issues with an assert failing. I tested that the first part fixes #909 and the second fixes #897.

A.  Loading 16/24/32-bit BMP from memory caused an assert to fail (excluding 16-bit BMP with `hsz` 12).

`img_buffer` offset was always compared with the buffer for `stbi_load_from_file()` but `stbi_load_from_memory()` uses an external buffer.

Resolution: Change `s->buffer_start` to `s->img_buffer_original`.

B. Loading BMP with large header from file caused assert to fail.

`img_buffer` points to `stbi_uc buffer_start[128]` but the largest BMP supported has a 138 byte header (`hsz` 124) causing `img_buffer` to wrap around to an offset of 10. The assert fails because 138 (header size) != 10 (offset in temp read buffer).

Resolution: Add the previously read bytes to the offset in temp read buffer to get the absolute offset.

Fixes bug introduced in c440a53 ("stb_image: fix reading BMP with explicit masks")

----

The image attached to #909 only triggered the assert when loaded using stbi_load_from_memory() due to small BMP header size. The image attached to #897 triggered the assert in both cases (stbi_load_from_memory() and stbi_load_from_file()).

I personally ran into the issue with a 32-bit BMP exported from GIMP 2.10 which triggered the assert in both cases like #897 . Zipped BMP: [clover.zip](https://github.com/nothings/stb/files/4377932/clover.zip)

### DIFF
--- a/stb_image.h
+++ b/stb_image.h
@@ -107,7 +107,7 @@ RECENT REVISION HISTORY:
     Oriol Ferrer Mesia      Josh Tobin         Matthew Gregan     github:phprus
     Julian Raschke          Gregory Mullen     Baldur Karlsson    github:poppolopoppo
     Christian Floisand      Kevin Schmidt      JR Smith           github:darealshinji
-    Brad Weinberger         Matvey Cherevko                       github:Michaelangel007
+    Brad Weinberger         Matvey Cherevko    Zack Middleton     github:Michaelangel007
     Blazej Dariusz Roszkowski                  Alexander Veselov
 */
 
@@ -751,6 +751,7 @@ typedef struct
    int read_from_callbacks;
    int buflen;
    stbi_uc buffer_start[128];
+   int callback_already_read;
 
    stbi_uc *img_buffer, *img_buffer_end;
    stbi_uc *img_buffer_original, *img_buffer_original_end;
@@ -764,6 +765,7 @@ static void stbi__start_mem(stbi__context *s, stbi_uc const *buffer, int len)
 {
    s->io.read = NULL;
    s->read_from_callbacks = 0;
+   s->callback_already_read = 0;
    s->img_buffer = s->img_buffer_original = (stbi_uc *) buffer;
    s->img_buffer_end = s->img_buffer_original_end = (stbi_uc *) buffer+len;
 }
@@ -775,7 +777,8 @@ static void stbi__start_callbacks(stbi__context *s, stbi_io_callbacks *c, void *
    s->io_user_data = user;
    s->buflen = sizeof(s->buffer_start);
    s->read_from_callbacks = 1;
-   s->img_buffer_original = s->buffer_start;
+   s->callback_already_read = 0;
+   s->img_buffer = s->img_buffer_original = s->buffer_start;
    stbi__refill_buffer(s);
    s->img_buffer_original_end = s->img_buffer_end;
 }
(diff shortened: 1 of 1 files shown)


nothings:
Looks good, except your "Fixes: c440a53" doesn't make any sense. I understand it fixes #909 and #897, but I'm not sure what that comment means. Does it fix a bug *introduced* by c440a53?

zturtleman:
Yes, the issues were introduced by that commit. I've change the commit message to say this.

`Fixes: commit-id ("message")` format is used by the Linux kernel, and to a lesser extent Mesa 3D, to mean this commit fixes an issue introduced in the specified commit. I think it's used to help backport bug fixes to stable releases. Anyway, sorry for the confusion.

nothings:
No problem.

cxw42:
I also have a file exhibiting #897, and this PR (at commit e919bcd) fixed the problem for me.  Thanks @zturtleman!

halx99:
It's time to merge this PR?

nothings:
> It's time to merge this PR?

Don't post unhelpful comments.

PR #935, opened 2020-03-24
Added assert(f->valid_bits >= n); on line 1604

vickit144:
I propose to remove this line because  f->valid_bits will never be less than zero since, in the while loop, you're adding 8 to it. Therefore, it will always evaluate to false. This is to help remove redundant code!


nothings:
It would be better to turn it into an assert >= n.

vickit144:
Good idea! I just committed the change to my branch.

PR #938, opened 2020-03-26
Implement %S specifier to support string structs

Manuzor:
This adds support for passing strings represented as length+ptr structs. The struct layout can be configured with `#define STB_SPRINTF_STRSTRUCT struct my_custom_string_layout`. Documentation has also been added to describe this change with a short example.

### DIFF
--- a/stb_sprintf.h
+++ b/stb_sprintf.h
@@ -20,6 +20,7 @@
 //    Adam Allison
 //    Arvid Gerstmann
 //    Markus Kolb
+//    Manuel Floruß
 //
 // LICENSE:
 //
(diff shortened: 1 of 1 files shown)


rygorous:
I don't think we should accept this PR.

%S is already used by both the Single Unix Specification v2 and the Microsoft CRT to indicate a wchar_t string. The last thing stb_sprintf should do is start gratuitously overloading format specifiers that have a different meaning everywhere else.

Furthermore, if you want to print strings given by a length + payload, you can already do so using standard C format specifiers via printf("like this: %.*s", len, ptr), which works on every printf implementation.

Feel free to maintain your own copy with your own modifications and do whatever you want with them, but this kind of thing does not belong in the "official" copy.

Manuzor:
Thank you for your thorough response. I thought already that this one was a little out of bounds but I still wanted to share it just in case it's useful to you. Feel free to close this PR.

I didn't know about the `%S` specifier already being used so broadly. I think I'll keep my changes in my own copy, like you said, but since using `%S` doesn't sound like a good idea I suppose I should find a more suitable letter that's less likely to conflict. Suggestions would be welcome.

The `%.*s` combo is what I've been using so far. But in a code base where you mainly have strings in a length+ptr format it becomes tedious to always be required to have a local variable of your string, passing the internals of it separately. As far as I can see there's no way of wrapping calls to or extending stb_sprintf in a way that lets me handle just the unwrapping of my strings and let stb_sprintf do the rest. That's why I dug in and added it directly.

Issue #368, opened 2016-11-07
stbi_write_bmp: support for writing 32b rgba


pixelmager:
Writing BMP files currently only support RGB 24bit files. Support for writing RGBA 32bit files would be handy.

Looks like this might only require modifying stbi_write_bmp_core(..)

darealshinji:
Indeed, only a small adjustment:
``` diff
--- a/stb_image_write.h
+++ b/stb_image_write.h
@@ -472,11 +472,11 @@
 
 static int stbi_write_bmp_core(stbi__write_context *s, int x, int y, int comp, const void *data)
 {
-   int pad = (-x*3) & 3;
-   return stbiw__outfile(s,-1,-1,x,y,comp,1,(void *) data,0,pad,
+   int pad = (-x*4) & 4;
+   return stbiw__outfile(s,-1,-1,x,y,comp,1,(void *) data,1,pad,
            "11 4 22 4" "4 44 22 444444",
-           'B', 'M', 14+40+(x*3+pad)*y, 0,0, 14+40,  // file header
-            40, x,y, 1,24, 0,0,0,0,0,0);             // bitmap header
+           'B', 'M', 14+40+(x*4+pad)*y, 0,0, 14+40,  // file header
+            40, x,y, 1,32, 0,0,0,0,0,0);             // bitmap header
 }
 
 STBIWDEF int stbi_write_bmp_to_func(stbi_write_func *func, void *context, int x, int y, int comp, const void *data)
```

PR #945, opened 2020-04-01
Fixes writing alpha-channel for BMP

pixelmager:
See https://github.com/nothings/stb/issues/368

An alternative approach could be to calculate values from the `comp`-variable. Adding a separate path for RGBA seemed like the most pragmatic approach though, as 24bit/32bit are the most common uses for the format.

### DIFF
--- a/stb_image_write.h
+++ b/stb_image_write.h
@@ -467,11 +467,20 @@ static int stbiw__outfile(stbi__write_context *s, int rgb_dir, int vdir, int x, 
 
 static int stbi_write_bmp_core(stbi__write_context *s, int x, int y, int comp, const void *data)
 {
-   int pad = (-x*3) & 3;
-   return stbiw__outfile(s,-1,-1,x,y,comp,1,(void *) data,0,pad,
-           "11 4 22 4" "4 44 22 444444",
-           'B', 'M', 14+40+(x*3+pad)*y, 0,0, 14+40,  // file header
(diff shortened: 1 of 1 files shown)


stb_image_write.h
```diff
@@ -467,11 +467,20 @@ static int stbiw__outfile(stbi__write_context *s, int rgb_dir, int vdir, int x,
 
 static int stbi_write_bmp_core(stbi__write_context *s, int x, int y, int comp, const void *data)
 {
-   int pad = (-x*3) & 3;
-   return stbiw__outfile(s,-1,-1,x,y,comp,1,(void *) data,0,pad,
-           "11 4 22 4" "4 44 22 444444",
-           'B', 'M', 14+40+(x*3+pad)*y, 0,0, 14+40,  // file header
-            40, x,y, 1,24, 0,0,0,0,0,0);             // bitmap header
+   if (comp == 4) {
+      int pad = (-x*4) & 4;
```
nothings, line 471: This should be (-x*4) & 3, I imagine.

nothings: Which is just 0 when I stop and think about it.

rygorous:
I'll add an issue for this but this PR is broken, that's a malformed BMP; BMPs with alpha need to use a v4 or v5 header (even 32bpp v1 BMPs don't have alpha if you're being strict) and the size computation is wrong. I'll discard this PR and try to do an alternative implementation.

PR #946, opened 2020-04-01
stb_image_write: Add support for disabling encoders

flibitijibibo:
Title says it all. This is designed exactly the same as stb_image.h's configuration defines.

### DIFF
--- a/stb_image_write.h
+++ b/stb_image_write.h
@@ -122,6 +122,7 @@ CREDITS:
    Jon Olick              -    original jo_jpeg.cpp code
    Daniel Gibson          -    integrate JPEG, allow external zlib
    Aarni Koskela          -    allow choosing PNG filter
+   Ethan Lee              -    STBIW_NO/ONLY_* to suppress building encoders
 
    bugfixes:
       github:Chribba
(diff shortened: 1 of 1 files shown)


rygorous:
We regret having the individual disable switches for codecs in stb_image
because giving the user N independent binary toggles means 2^N possible
configurations with potential to cause problems such as warnings about
unused functions at high warning levels. This in turn results in a
profusion of #ifdefs all over the code to work around this, or #ifdefs
elsewhere to locally disable warnings via compiler-specific pragmas, one
set per compiler.

The switches in stb_image are not going away any time soon because we're
several years into this mess and don't want to break people's builds
unnecessarily, but having this has caused a disproportionately large
maintenance headache for a feature that very few people actually use, and
we're not eager to repeat this mistake, especially since stb_image_write
has a different entry point per format anyway.

Feel free to delete the functions you don't want in your local copy if you
don't want to use -ffunction-sections or similar to let the linker do the
dead stripping for you, but especially in stb_image_write, this offers next
to no benefit for what we already know will be painful to maintain in the
long run.

kg:
The file size is MUCH bigger without disabling codecs, though. Is there a method you recommend for mitigating this? For my project it's on the order of a few hundred KB and that matters a lot for browser builds and demos.

rygorous:
As said, compile with -ffunction-sections and you should only get the
encoders you use.

PR #947, opened 2020-04-02
Fix unused parameter warnings

vikaig:
```
stb_vorbis.c: In function 'do_floor':
stb_vorbis.c:3053:95: warning: unused parameter 'step2_flag' [-Wunused-parameter]
 3053 | static int do_floor(vorb *f, Mapping *map, int i, int n, float *target, YTYPE *finalY, uint8 *step2_flag)
      |                                                                                        ~~~~~~~^~~~~~~~~~
stb_vorbis.c: In function 'vorbis_decode_packet_rest':
stb_vorbis.c:3161:86: warning: unused parameter 'left_end' [-Wunused-parameter]
 3161 | static int vorbis_decode_packet_rest(vorb *f, int *len, Mode *m, int left_start, int left_end, int right_start, int right_end, int *p_left)
      |                                                                                  ~~~~^~~~~~~~
```

### DIFF
--- a/stb_vorbis.c
+++ b/stb_vorbis.c
@@ -31,7 +31,7 @@
 //    Phillip Bennefall  Rohit               Thiago Goulart
 //    github:manxorist   saga musix          github:infatum
 //    Timur Gagiev       Maxwell Koo         Peter Waller
-//    github:audinowho   Dougall Johnson
+//    github:audinowho   Dougall Johnson     github:vikaig
 //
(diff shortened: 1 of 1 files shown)


nothings:
Needs to be done with a proper macro to allow updating when compilers change things, and the macro needs to be more than a void cast, which doesn't work on all compilers. See STBI_NOTUSED in stb_image.h

PR #950, opened 2020-04-06
stb_image: make unpremultiply and de-iPhone flags thread_local as well

mosra:
Follows the change done for vertical flipping in eb48fbdcede98f93dab3a5e42ce5b0072b739cf1 for these two options as well -- I'm using these three flags basically the same way, and if one is `thread_local`, why not the others.

Tried to follow existing code style as closely as possible, let me know if something is off :)

### DIFF
--- a/stb_image.h
+++ b/stb_image.h
@@ -473,6 +473,8 @@ STBIDEF void stbi_set_flip_vertically_on_load(int flag_true_if_should_flip);
 // as above, but only applies to images loaded on the thread that calls the function
 // this function is only available if your compiler supports thread-local variables;
 // calling it will fail to link if your compiler doesn't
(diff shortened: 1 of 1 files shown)


nothings:
I'd actually prefer to just remove unpremultiply and de-iphone from stb_image entirely. Doing so poses a difficult management/release problem, so maybe I won't, in which case this is reasonable. But I may sit on this for a while.

mosra:
Okay, good to know -- then I'll gradually move this processing to the application instead.

To make the deprecation/removal easier, maybe it might make sense to wrap those two in an `#ifdef` at first, and then switch it off at some point?

rygorous:
The iPhone format PNGs still seem to be a thing (unfortunately), so whatever, I'll take it. This is now merged into the dev branch and will be in the next release.

PR #959, opened 2020-04-27
Fix infinite loops when using stdio callbacks

rcgordon:
This pull request fixes some cases where stb_image, with maliciously-crafted input, can get into an infinite loop if using the stdio callbacks, because stdio has some subtle quirks.

The fseek change is the most local fix for the described issue, but it's possible a better fix is not setting read_from_callbacks to zero, but I didn't want to risk unforeseen ramifications by messing with that.

I've been fuzzing stb_image after some discussion on Twitter, so there's probably going to be some other pull requests coming soon. Sorry in advance!  :)

### DIFF
--- a/stb_image.h
+++ b/stb_image.h
@@ -789,7 +789,12 @@ static int stbi__stdio_read(void *user, char *data, int size)
 
 static void stbi__stdio_skip(void *user, int n)
 {
+   int ch;
    fseek((FILE*) user, n, SEEK_CUR);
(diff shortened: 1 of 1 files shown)


nothings:
Thanks, Ryan. All PRs are appreciated!

rcgordon:
I'm going to close this PR and start a new one with this and the rest of the fuzzing fixes, so there's no confusion and everything merges cleanly.

Ignore this and look at https://github.com/nothings/stb/pull/960 instead, thanks!

PR #960, opened 2020-04-28
stb_image fuzzing fixes

rcgordon:
This PR has all the hardening work I did on stb_image:

- Fuzzing inputs with AFL, using a 2x2 image in bmp, gif, jpg, hdr, png, pnm, psd, and tga. (pic not tested). Those files are available in [stbi_afl.zip](https://github.com/nothings/stb/files/4547882/stbi_afl.zip) .
- Running this testsuite through Valgrind and AddressSanitizer: https://code.google.com/archive/p/imagetestsuite/downloads

Testing was done with and without SSE2 support.

### DIFF
--- a/stb_image.h
+++ b/stb_image.h
@@ -318,7 +318,14 @@ RECENT REVISION HISTORY:
 //   - If you use STBI_NO_PNG (or _ONLY_ without PNG), and you still
 //     want the zlib decoder to be available, #define STBI_SUPPORT_ZLIB
 //
-
+//  - If you define STBI_MAX_DIMENSIONS, stb_image will reject images greater
+//    than that size (in either width or height) without further processing.
+//    This is to let programs in the wild set an upper bound to prevent
+//    denial-of-service attacks on untrusted data, as one could generate a
+//    valid image of gigantic dimensions and force stb_image to allocate a
+//    huge block of memory and spend disproportionate time decoding it. By
+//    default this is set to (1 << 24), which is 16777216, but that's still
+//    very big.
 
 #ifndef STBI_NO_STDIO
 #include <stdio.h>
@@ -732,6 +739,10 @@ static int stbi__sse2_available(void)
 
 #ifndef STBI_SIMD_ALIGN
 #define STBI_SIMD_ALIGN(type, name) type name
+#endif
+
+#ifndef STBI_MAX_DIMENSIONS
+#define STBI_MAX_DIMENSIONS (1 << 24)
 #endif
 
 ///////////////////////////////////////////////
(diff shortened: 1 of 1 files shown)


randy408:
There's a PNG-only fuzzer on [OSS-Fuzz](https://github.com/google/oss-fuzz) that found [this](https://github.com/nothings/stb/pull/918) which doesn't seem to be addressed by this PR. I'm the only one who looks at the bug reports for that fuzzer so It hasn't been expanded to other formats because I reckoned there would too many  bug reports for me to handle.

If you're willing to address the bug reports that would come up from increasing code coverage I could add you as a contact for the project. Overall it's a simple process, issues are automatically created on chromium.org with test cases and are auto-closed when the fix is in master.

rcgordon:
> I maintain a PNG-only fuzzer on [OSS-Fuzz](https://github.com/google/oss-fuzz) that found [this](https://github.com/nothings/stb/pull/918) which doesn't seem to be addressed by this PR.

I added your fix to this PR (commit credited to you, of course!). It's possible other changes made mitigated the immediate need for this fix, and it's also possible my fuzzing just missed it, but it looked like a correct, localized and reasonable change in any case, so I added it in.

> If you're willing to address the bug reports that would come up from increasing code coverage I could add you as a contact for the project.

Thanks, but I don't have time to take on that responsibility.

stb_image.h
```diff
@@ -4079,7 +4112,12 @@ static int stbi__zhuffman_decode_slowpath(stbi__zbuf *a, stbi__zhuffman *z)
 stbi_inline static int stbi__zhuffman_decode(stbi__zbuf *a, stbi__zhuffman *z)
 {
    int b,s;
-   if (a->num_bits < 16) stbi__fill_bits(a);
+   if (a->num_bits < 16) {
+      if (stbi__zeof(a)) {
+         return -1;   /* report error for unexpected end of data. */
```
bjornbytes, line 4117: Is it possible that this is too aggressive now?  I'm using stb_image to decompress files in zip archives (long story) and I started seeing some unexpected failures after this change.  When this error condition is hit, the zbuf has 13 bits in it and it's trying to decode the last byte of the stream.  Still brushing up on huffman stuff so I might be missing something obvious.  I can scrape together a test case if that's helpful

nothings: What if you add an stbi_fill_bits(a) after the if num_bits and before the if stbi_zeof?

nothings: actually I think this is going to need a messier fix, since it can potentially intentionally "read past the end of file" while trying to do the "fast" huffman table lookup. ugh, need to figure out the right way to do it.

Issue #967, opened 2020-05-17
stbi: build error on GCC 4.8.4


recp:
Hello,

I'm getting build error after updated `stb_image.h`: 
https://travis-ci.org/github/recp/AssetKit/jobs/688117801

it says:

```bash
./lib/stb/stb_image.h:909:4: error: ‘stbi__g_failure_reason’ undeclared (first use in this function)
1099    stbi__g_failure_reason = str;
1100    ^
```

```bash
./lib/stb/stb_image.h: At top level:
1102./lib/stb/stb_image.h:1036:26: error: expected ‘=’, ‘,’, ‘;’, ‘asm’ or ‘__attribute__’ before ‘int’
1103 static STBI_THREAD_LOCAL int stbi__vertically_flip_on_load_local, stbi__vertically_flip_on_load_set;
1104           
```

```bash
./lib/stb/stb_image.h: In function ‘stbi_set_flip_vertically_on_load_thread’:
1106./lib/stb/stb_image.h:1040:4: error: ‘stbi__vertically_flip_on_load_local’ undeclared (first use in this function)
1107    stbi__vertically_flip_on_load_local = flag_true_if_should_flip;
1108    ^
```

```bash
./lib/stb/stb_image.h:1041:4: error: ‘stbi__vertically_flip_on_load_set’ undeclared (first use in this function)
1110    stbi__vertically_flip_on_load_set = 1;
1111    ^
```
... and so on

build passes on clang and VC++.

r.

nothings:
https://gcc.gnu.org/onlinedocs/gcc/Thread-Local.html

Presumably the above extension is not supported in older versions of gcc, although this is contradictory to what I've read elsewhere. Unfortunately we can't tell from the error message which version of thread-locals it's using.

So first test is to go to line 580, which is the __GNUC__ case, and insert a #error and verify that that is the path being taken. (Alternatively, comment it out and check if it successfully compiles. This is not a proper fix, just a way of checking that I've diagnosed it correctly.)

If that's NOT the case then I don't know what's happening.

If that is the case, the docs don't say what version of gcc is required, so there's nothing I can do directly. If you want to figure out what version is required for that extension, and figure out how to do an #if to detect that, then I can fix the test. (Unfortunately, that undermines the whole idea of doing this thread local thing, since I only did it on the assumption that the rule for explaining it would be simple.)

recp:
@nothings sorry for the late response,

[...]

PR #971, opened 2020-05-25
stbi: fix thread local selector

recp:
* GCC < 5 supports __thread and GCC >= 5 supports C11 with _Thread_local
* Skip _Thread_local for MSVC because it may not be supported

---

See https://github.com/nothings/stb/issues/967

---

Also I have found these:

1. [Bug 203066 _Thread_local detection is broken with GCC < 4.9](https://lists.freebsd.org/pipermail/freebsd-standards/2015-September/003092.html)
2. [ _Thread_local detection is broken with GCC < 4.9](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=203066)

I have updated thread selector like this:

before:

```C
#ifndef STBI_NO_THREAD_LOCALS
   #if defined(__cplusplus) &&  __cplusplus >= 201103L
      #define STBI_THREAD_LOCAL       thread_local
   #elif defined (__STDC_VERSION__) && __STDC_VERSION__ >= 201112L
      #define STBI_THREAD_LOCAL       _Thread_local
   #elif defined(__GNUC__)
      #define STBI_THREAD_LOCAL       __thread
   #elif defined(_MSC_VER)
      #define STBI_THREAD_LOCAL       __declspec(thread)
#endif
#endif
```

after:

```C
#ifndef STBI_NO_THREAD_LOCALS
   #if defined(__cplusplus) &&  __cplusplus >= 201103L
      #define STBI_THREAD_LOCAL       thread_local
   #elif defined(__GNUC__) && __GNUC__ < 5
      #define STBI_THREAD_LOCAL       __thread
   #elif defined(_MSC_VER)
      #define STBI_THREAD_LOCAL       __declspec(thread)
   #elif defined (__STDC_VERSION__) && __STDC_VERSION__ >= 201112L && !defined(__STDC_NO_THREADS__)
      #define STBI_THREAD_LOCAL       _Thread_local
   #endif
#endif
```

Now if GCC's version is lower than 5 e.g. 4.8 than __thread will be used. Because AFAIK, C11 is supported since GCC 5.

### DIFF
--- a/stb_image.h
+++ b/stb_image.h
@@ -574,13 +574,13 @@ STBIDEF int   stbi_zlib_decode_noheader_buffer(char *obuffer, int olen, const ch
 #ifndef STBI_NO_THREAD_LOCALS
    #if defined(__cplusplus) &&  __cplusplus >= 201103L
       #define STBI_THREAD_LOCAL       thread_local
-   #elif defined (__STDC_VERSION__) && __STDC_VERSION__ >= 201112L
-      #define STBI_THREAD_LOCAL       _Thread_local
-   #elif defined(__GNUC__)
+   #elif defined(__GNUC__) && __GNUC__ < 5
       #define STBI_THREAD_LOCAL       __thread
    #elif defined(_MSC_VER)
       #define STBI_THREAD_LOCAL       __declspec(thread)
-#endif
+   #elif defined (__STDC_VERSION__) && __STDC_VERSION__ >= 201112L && !defined(__STDC_NO_THREADS__)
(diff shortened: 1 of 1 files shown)


nothings:
Makes sense! Thanks for doing the investigative work.

recp:
@nothings any chance to merge this soon?

nothings:
Note that I only push to github when doing releases, so merges only retire on github when there's a release.

recp:
@nothings thanks

PR #979, opened 2020-06-23
Added missing prototype for stbtt_FindSVGDoc()

ITikhonov:
Was getting an error:

./stb_truetype.h:2690:24: error: no previous prototype for function 'stbtt_FindSVGDoc' [-Werror,-Wmissing-prototypes]

Or may be it was intended to be static function?


nothings:
?

PR #981, opened 2020-06-25
stb_ds: fix bug where arraddn would return index instead of pointer

gustavolsson:
I ported some code over from `stretchy_buffer` to `stb_ds` and found that `arraddn` returns _the index_ of the first of the newly added items instead of a pointer to it (as it states in the documentation).

This PR makes the behavior consistent with what is stated in the documentation and with stretchy_buffer's `sb_add` implementation.


nothings:
duplicate of #932

PR #984, opened 2020-06-30
stb_image: fix build errors

daemyung:
Build errors are generated on Microsoft (R) C/C++ Optimizing Compiler Version 19.26.28806 for x64.

Error messages are below:
stb_image.h(1681,10): error C2196: case value '17' already used
stb_image.h(1683,10): error C2196: case value '12' already used
stb_image.h(1684,10): error C2196: case value '9' already used
stb_image.h(1685,10): error C2196: case value '10' already used
stb_image.h(1688,10): error C2196: case value '33' already used
stb_image.h(1738,10): error C2196: case value '17' already used
stb_image.h(1740,10): error C2196: case value '12' already used
stb_image.h(1741,10): error C2196: case value '9' already used
stb_image.h(1742,10): error C2196: case value '10' already used
stb_image.h(1745,10): error C2196: case value '33' already used

### DIFF
--- a/stb_image.h
+++ b/stb_image.h
@@ -108,7 +108,7 @@ RECENT REVISION HISTORY:
     Julian Raschke          Gregory Mullen     Baldur Karlsson    github:poppolopoppo
     Christian Floisand      Kevin Schmidt      JR Smith           github:darealshinji
     Brad Weinberger         Matvey Cherevko                       github:Michaelangel007
-    Blazej Dariusz Roszkowski                  Alexander Veselov
+    Blazej Dariusz Roszkowski                  Alexander Veselov  Daemyung Jang
 */
 
 #ifndef STBI_INCLUDE_STB_IMAGE_H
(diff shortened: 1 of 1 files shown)


nothings:
That fix doesn't make any sense, the STBI_COMBO macro right above it is what generates the case expression, and it's parenthesized properly, so I don't see how that could fix it.

nothings:
I mean, the proposed fix is clearly better code quality regardless, I just don't understand how the bug arises. Is this some very new version of MSVC? It seems like a compiler bug, but maybe I'm missing something.

daemyung:
I agree your oppinion. Actually there is no different between previous code and new code. I don't know why this errors are generated on the latest MSVC compiler. It's up to you mergine this patch. If you don't want, I will use it on my local. But as far as you know this patch also enhances better code quality. :)

daemyung:
This is MSVC compiler information on my machine.

PS C:\Program Files (x86)\Microsoft Visual Studio\2019\Professional\VC\Tools\MSVC\14.26.28801\bin\Hostx64\x64> ./cl.exe
Microsoft (R) C/C++ Optimizing Compiler Version 19.26.28806 for x64
Copyright (C) Microsoft Corporation.  All rights reserved.

nothings:
Actually, looking at it in context, this is a pretty common convention for me to not parenthesize that kind of macro. There's no way for it to go wrong.

PR #993, opened 2020-07-12
stb_ds: fix shputs() for strdup and arena hash table

srdjanstipic:
This commit fixes the issue of inserting a struct to strdup and arena hastable.
Downside of this change is that it introduces additional instructions (also for default string hashtable)
but inserted code should have negligible overhead because it is easy to predict.

Feel free to reject this change if you think that it does not make sense
or it is not worth additional complexity.

### DIFF
--- a/stb_ds.h
+++ b/stb_ds.h
@@ -512,6 +512,7 @@ extern void * stbds_shmode_func(size_t elemsize, int mode);
 
 #define stbds_header(t)  ((stbds_array_header *) (t) - 1)
 #define stbds_temp(t)    stbds_header(t)->temp
+#define stbds_temp_key(t) stbds_header(t)->temp_key
 
 #define stbds_arrsetcap(a,n)  (stbds_arrgrow(a,0,n))
 #define stbds_arrsetlen(a,n)  ((stbds_arrcap(a) < (size_t) (n) ? stbds_arrsetcap((a),(size_t)(n)),0 : 0), (a) ? stbds_header(a)->length = (size_t) (n) : 0)
@@ -587,7 +588,8 @@ extern void * stbds_shmode_func(size_t elemsize, int mode);
 
 #define stbds_shputs(t, s) \
     ((t) = stbds_hmput_key_wrapper((t), sizeof *(t), (void*) (s).key, sizeof (s).key, STBDS_HM_STRING), \
-     (t)[stbds_temp((t)-1)] = (s))
+     (t)[stbds_temp((t)-1)] = (s), \
+     (t)[stbds_temp((t)-1)].key = stbds_temp_key((t)-1))
 
 #define stbds_pshput(t, p) \
     ((t) = stbds_hmput_key_wrapper((t), sizeof *(t), (void*) (p)->key, sizeof (p)->key, STBDS_HM_PTR_TO_STRING), \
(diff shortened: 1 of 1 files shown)


nothings:
Yeah, I was thinking about whether this would really be that hacky last night, and given the existing temp, it's not so bad. The only thing I would change is to just always write the key regardless, rather than bothering to test the hash table type.

srdjanstipic:
My initial solution was always writing the key but I thought that we could minimize number of writes.
I will modify the PR to update the key always.

stb_ds.h
```diff
@@ -587,7 +588,8 @@ extern void * stbds_shmode_func(size_t elemsize, int mode);
 
 #define stbds_shputs(t, s) \
     ((t) = stbds_hmput_key_wrapper((t), sizeof *(t), (void*) (s).key, sizeof (s).key, STBDS_HM_STRING), \
-     (t)[stbds_temp((t)-1)] = (s))
+     (t)[stbds_temp((t)-1)] = (s), \
+     (t)[stbds_temp((t)-1)].key = stbds_temp_key((t)-1))
```
srdjanstipic, line 592: Update the key.

stb_ds.h
```diff
@@ -1410,9 +1413,9 @@ void *stbds_hmput_key(void *a, size_t elemsize, void *key, size_t keysize, int m
       stbds_temp(a) = i-1;
 
       switch (table->string.mode) {
-         case STBDS_SH_STRDUP:  *(char **) ((char *) a + elemsize*i) = stbds_strdup((char*) key); break;
-         case STBDS_SH_ARENA:   *(char **) ((char *) a + elemsize*i) = stbds_stralloc(&table->string, (char*)key); break;
-         case STBDS_SH_DEFAULT: *(char **) ((char *) a + elemsize*i) = (char *) key; break;
+         case STBDS_SH_STRDUP:  stbds_temp_key(a) = *(char **) ((char *) a + elemsize*i) = stbds_strdup((char*) key); break;
+         case STBDS_SH_ARENA:   stbds_temp_key(a) = *(char **) ((char *) a + elemsize*i) = stbds_stralloc(&table->string, (char*)key); break;
+         case STBDS_SH_DEFAULT: stbds_temp_key(a) = *(char **) ((char *) a + elemsize*i) = (char *) key; break;
```
srdjanstipic, line 1418: Save the key in all sh cases.

srdjanstipic:
> The only thing I would change is to just always write the key regardless,
> rather than bothering to test the hash table type.

Done

nothings:
The cost of writing the key is small since we just wrote the structure, so it will still be in cache, and loading the temp key is cheap since we just wrote it into the temp field during the function call. Of course the cost the other way is also small, as the branch will always predict correctly. So the decisive factor is that it's slightly less code to just always do it.

srdjanstipic:
> So the decisive factor is that it's slightly less code to just always do it

I agree.

Should I fix some other things in the PR?

nothings:
no, it's fine. I only use github for releasing, so you won't see the PR merged until the next release.

nothings:
I moved the temp_key storage into the hash_table structure because the existing struct is used by arrays as well as hash tables, and I didn't want arrays to have to pay the storage overhead. This makes loading the temp key require one more indirection, but it should still all be in cache so it still doesn't matter.

PR #998, opened 2020-07-18
stb_image: Fix gcc 7.5.0 compiler warnings

jwatte:
sizeof() returns unsigned, so cast it to compare to signed.
STB_REALLOC_SIZED() by default swallows the "oldsz" argument,
which generates a "set but not used" warning, so define a
static function (that gets inlined) to swallow this warning.

Provided by Jon Watte (github.com/jwatte) and released under
the STB Image license or the "public domain" license,
whichever you prefer.


rygorous:
These are fixed (in a different way) in the current dev branch, closing this PR.

PR #1003, opened 2020-07-24
stb_ds: hmgetp_null stbds_temp usage fix

Vzaa:
I was testing out stb_ds.h's hashmap and I believe I may have found an error in `hmgetp_null`.

I was getting invalid pointers when using `hmgetp_null` while `hmgetp` was working fine. Comparing the macros I've noticed `hmgetp` uses `stbds_temp((t)-1)` while  `hmgetp_null` uses `stbds_temp(t)-1`. Changing the macro fixed the issue for me.

The assertions in the code below were failing before for me on gcc 7.5.0 and clang 10.0.1

```c
#include <stdio.h>
#include <assert.h>

#define STB_DS_IMPLEMENTATION
#include "stb_ds.h"

typedef struct {
	int key;
	int value;
} foo_t;

int main(int argc, char const *argv[])
{
	foo_t *hash = NULL;
	foo_t *a;
	foo_t *b;

	hmput(hash, 1, 42);
	hmput(hash, 2, 43);

	a = hmgetp(hash, 1);
	b = hmgetp_null(hash, 1);
	assert(a == b);

	a = hmgetp(hash, 2);
	b = hmgetp_null(hash, 2);
	assert(a == b);

	hmfree(hash);
}
```

### DIFF
--- a/stb_ds.h
+++ b/stb_ds.h
@@ -595,7 +595,7 @@ extern void * stbds_shmode_func(size_t elemsize, int mode);
 #define stbds_hmget_ts(t, k, temp)  (stbds_hmgetp_ts(t,k,temp)->value)
 #define stbds_hmlen(t)        ((t) ? (ptrdiff_t) stbds_header((t)-1)->length-1 : 0)
 #define stbds_hmlenu(t)       ((t) ?             stbds_header((t)-1)->length-1 : 0)
-#define stbds_hmgetp_null(t,k)  (stbds_hmgeti(t,k) == -1 ? NULL : &(t)[stbds_temp(t)-1])
+#define stbds_hmgetp_null(t,k)  (stbds_hmgeti(t,k) == -1 ? NULL : &(t)[stbds_temp((t)-1)])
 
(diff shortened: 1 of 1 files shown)


pixelherodev:
I can confirm this. The issue is present in both hmgetp and shgetp. I patched them both manually and was going to submit them here when I saw this PR.

Vzaa:
> I can confirm this. The issue is present in both hmgetp and shgetp. I patched them both manually and was going to submit them here when I saw this PR.

You're right. I've updated the PR to change both.

nothings:
fixed in next release

pixelherodev:
Thanks!

makuto:
When is the next release coming? I just ran into this issue as well, with c9064e3.

makuto:
I saw this is now fixed with 781609b7e9f79e2d9092afb2cdf6364a1f853b31, so should be closed.

nothings:
It's already closed.

Vzaa:
Thank you

PR #1005, opened 2020-07-27
stb_vorbis: Allocate memory for empty buffers in setup_malloc()

bqqbarbhg:
I had a valid file failing with `VORBIS_outofmem` as the comment list was empty and I wasn't using an allocation buffer causing `setup_malloc()` to return `NULL` that gets interpreted as an allocation failure. You _could_ return a sentinel value here and check against it in `setup_free()` as an optimization, but I'm not sure if it's worth the complexity as you'd want to use an allocation buffer if you care about performance anyway.

In my case `f->comment_list_length` was zero here, which I assume is valid. Applying the fix in the PR made `stb_vorbis` able to decode the file without any further issues.
```c
   f->comment_list = (char**)setup_malloc(f, sizeof(char*) * (f->comment_list_length));
   if (f->comment_list == NULL)                     return error(f, VORBIS_outofmem);
```

### DIFF
--- a/stb_vorbis.c
+++ b/stb_vorbis.c
@@ -934,7 +934,7 @@ static void *make_block_array(void *mem, int count, int size)
 
 // sentinel buffer returned for empty setup_malloc()
 // declared const to catch illegal writes
-static const char setup_malloc_null_buffer[1];
+static const char setup_malloc_null_buffer[1] = { 0 };
 
 static void *setup_malloc(vorb *f, int sz)
 {
--- a/stb_vorbis.c
+++ b/stb_vorbis.c
@@ -942,7 +942,7 @@ static void *setup_malloc(vorb *f, int sz)
       f->setup_offset += sz;
       return p;
    }
-   return sz ? malloc(sz) : NULL;
+   return malloc(sz);
 }
 
 static void setup_free(vorb *f, void *p)
--- a/stb_vorbis.c
+++ b/stb_vorbis.c
@@ -932,6 +932,10 @@ static void *make_block_array(void *mem, int count, int size)
    return p;
 }
 
+// sentinel buffer returned for empty setup_malloc()
+// declared const to catch illegal writes
+static const char setup_malloc_null_buffer[1];
+
 static void *setup_malloc(vorb *f, int sz)
 {
    sz = (sz+7) & ~7; // round up to nearest 8 for alignment of future allocs.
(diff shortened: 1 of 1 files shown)


nothings:
Thank you for the clear explanation!

astralinux:
Proposed patch
-   return sz ? malloc(sz) : NULL;
+  return malloc(sz);
is implementation dependent. If sz is 0 then malloc(0) may return 0 or some value suitable for free. If it returns 0 then there will be the same trouble as before patch.

bqqbarbhg:
@astralinux interesting, I thought `malloc(0)` was specified to return a non-zero pointer since that's the only thing I've observed. I fixed this PR to return a sentinel buffer that is checked against in `setup_free()`.

Flix01:
@bqqbarbhg: that fixed some code I'm writing to decode internet radio streams (for some radios ```stb_vorbis_open_pushdata(...)``` returned the error ```VORBIS_outofmem```, and then Valgrind complained about the line ```return sz ? malloc(sz) : NULL;``` AFAIR).

The only issue is that my version of gcc complains that ```setup_malloc_null_buffer``` is uninitialized, so I'm using:
```c
static const char setup_malloc_null_buffer[1] = {0};
```

bqqbarbhg:
@Flix01 good point, I think MSVC is grumpy about const-data without an initializer too! Committed a fix to this PR.

flibitijibibo:
We ended up pushing a similar fix to FAudio's copy of stb_vorbis:

https://github.com/FNA-XNA/FAudio/commit/345586a0cf62532e5c8f05a856f91275a578e1ea

julianxhokaxhiu:
I have found the very same issues after I upraded the `stb_vorbis.c` file on SoLoud to 1.20.

The file I'm currently using to test is this one: https://github.com/maxmods/bah.mod/blob/7dd86d73d6f6b9d9acc3a7637aa5393a38b02a16/soloud.mod/examples/audio/windy_ambience.ogg

Would you consider merging this fix?

DanielGibson:
I think this is overkill - `f->comment_list` being `NULL` should be ok, unless I missed something it's only accessed if `f->comment_list_length > 0`, so just leaving it at NULL (and not returning an error) should suffice to fix the issue (and works for me)

nothings:
Fixed this a different way. Thank you for reporting, my apologies for how long it took to address.

nothings:
pull current master

PR #1006, opened 2020-07-29
Fix alloca in stb_vorbis for Sun platforms and add generic fallback for platforms with GCC

GabrielRavier:
Should be pretty self-explanatory. Yes, this is not a joke, I do indeed use OpenIndiana. But so yeah, the first part of the PR should fix compilation on Sun platforms, and the second part should fix compilation on any platform that doesn't define `alloca` in `stdlib.h`, is not supported by the checks we do and uses GCC.


rygorous:
I'll take the fix for Sun platforms but not compiling when the right header to include for the target wasn't detected is a feature not a bug as far as I'm concerned.

PR #1007, opened 2020-07-30
Updated stb_image_write description

voidbert:
The header supports saving JPEG and HDR images but it wasn't in the description yet.


rygorous:
Has been updated in the dev branch but not with this PR (since there was a missing format in stb_image as well). Will be fixed when we next do a release.

PR #1008, opened 2020-08-01
Update stb_vorbis.c

astralinux:
If vendor or user comments omitted then do not setup_malloc() with zero length, do not report "out of memory errors", but just continue decoding.


nothings:
Is the fixing the same thing as #1005, which is a much simpler fix?

astralinux:
Not sure.  Proposed there malloc(sz) - implementation dependent if sz is 0.

bqqbarbhg:
If this solution is preferable to #1005 which is now fixed should probably add an `assert(sz > 0)` to setup malloc as it's not supported.

astralinux:
#1005 is OK now
this my commit can be discarded

nothings:
thanks.

PR #1009, opened 2020-08-03
Fix comment and indentation in stbtt_solve_cubic

adrianlizarraga:
This branch fixes:
- The comment above `stbtt__solve_cubic` to reflect the actual usage of parameters in solving the cubic equation. Refer to https://en.wikipedia.org/wiki/Cubic_equation
- The indentation in the same function (tab => 3 spaces)

These are very minor updates to a private implementation function, but in the words of Hilary Duff "Why not?".


rygorous:
This is now fixed in the dev branch, although not by merging this PR.

PR #985, opened 2020-06-30
Make file formatting comply with POSIX standards and remove trailing space characters

aaronfranke:
This PR does two things, split up into two commits:

* Add newlines to the ends of text files where they are missing (mostly in the `tests/` folder). To comply with POSIX standards, and for best compatibility with command-line tools, text files must end in newline characters. Also, GCC will refuse to compile files that don't end in newline characters when you're targeting old C++ versions.

* Remove trailing space characters. They do nothing, they are pointless just take up disk space.

I did not commit any changes to `dsp` files because I wasn't sure what those are, and I don't want to break anything.

If desired, I can make a follow-up PR to provide you with the script I used, and a CI config for automatically running checks.

PR #1012, opened 2020-08-06
Make file formatting comply with POSIX standards and remove trailing space characters

aaronfranke:
This is the 3rd PR attempting to solve this issue, it supersedes #728 and #985.

@nothings I see that you [made a commit recently to remove trailing spaces](https://github.com/nothings/stb/commit/e140649ccf40818781b7e408f6228a486f6d254b), but there are still a ton of trailing spaces in this repo. This PR is what happens when I run my formatting script on this repo. In particular, the `tests/`, `tools/`, and `deprecated/` folders have a ton of trailing spaces. I think it would be best to just merge this PR instead of closing it and saying you'll fix it later.

As a reminder, this PR makes file formatting comply with POSIX and Unix standards and removes trailing spaces. UTF-8, LF, no BOM, and newlines at the end of files. Trailing spaces do nothing so they should be removed.

I did not commit any changes to `dsp` files because I wasn't sure what those are, and I don't want to break anything.

If desired, I can make a follow-up PR to provide you with the script I used, and a CI config for automatically running checks.


nothings:
(a) I am unlikely to ever accept a PR like this since it's going to cause too many conflicts with other PRs

(b) listing the actual actionable issues would be more useful. E.g. you say this PR "removes trailing spaces, UTF-8, LF no BOM, and newlines at the end of files". Are you actually saying there are instances of UTF-8 and BOM in the source tree? Because that's what your choice of words says, but I'd be pretty surprised.

(c) congrats, I don't give a crap about the contents of tests/ or tools/ when it comes to trailing newlines, and I don't know why you do. Like, what is the actual problem you're trying to solve? Why do you think I care about solving it?

(d) I don't want your scripting solution because every time somebody says this it turns out they mean "something that requires some program that windows machines don't have installed by default". It's either going to get fixed long-term by having a program to do it in the tools/ directory that compiles on windows, or it's not going to get fixed.

(e) If you want to be taken seriously, don't write shit like this:

> I think it would be best to just merge this PR instead of closing it and saying you'll fix it later.

If you actually knew what was best, I wouldn't have had four earlier bullet points.

aaronfranke:
> (a) I am unlikely to ever accept a PR like this since it's going to cause too many conflicts with other PRs

Of course it will, but the alternative is that you never fix formatting problems, which sucks.

This kind of PR is very easy to re-create, so if you want to merge other PRs first, I can re-create this again.

> (b) listing the actual actionable issues would be more useful. E.g. you say this PR "removes trailing spaces, UTF-8, LF no BOM, and newlines at the end of files". Are you actually saying there are instances of UTF-8 and BOM in the source tree? Because that's what your choice of words says, but I'd be pretty surprised.

Sorry for the confusion. This PR only removes trailing spaces and adds EOL at EOF, it does not remove any BOMs or CRs because there are none, but I just wanted to tell you what things my formatting script uses.

> (c) congrats, I don't give a crap about the contents of tests/ or tools/ when it comes to trailing newlines, and I don't know why you do. Like, what is the actual problem you're trying to solve? Why do you think I care about solving it?

The problem I am trying to solve is bad formatting, which affects downstream projects. The easiest, most comprehensive, and overall best fix is to format the entire upstream repo, rather than just a specific set of files that are used downstream (besides, how can you know what's used downstream? maybe downstream uses tests or the tools?).

The question I have for you is, why don't you want well-formatted code everywhere?

> (d) I don't want your scripting solution because every time somebody says this it turns out they mean "something that requires some program that windows machines don't have installed by default". It's either going to get fixed long-term by having a program to do it in the tools/ directory that compiles on windows, or it's not going to get fixed.

The idea is to have it in CI checks on GitHub, to alert you of issues before merging PRs. If a user is unable to run the script on their local machine, they can still just push and let GitHub tell them if there's anything wrong.

nothings:
>> > (a) I am unlikely to ever accept a PR like this since it's going to cause too many conflicts with other PRs
> 
> Of course it will, but the alternative is that you never fix formatting problems, which sucks.

I don't know what your problem is, but I didn't bother reading past this sentence, since obviously this is false. So!

aaronfranke:
So basically, you're a cunt.

PR #1016, opened 2020-08-15
stb_image.h: Suppress warnings about out_size, delay_size

jackodirks:
These two variables are unused in a Linux environment (the macros that use them actually do not
use them), which generate warnings. But the macros are set up like this
for compatibility with other environments, so that these warnings are
not easily avoided.
This fix just tells the compiler that these variables might not be used,
and that it should not warn, which seems the best of all worlds.

### DIFF
--- a/stb_image.h
+++ b/stb_image.h
@@ -7212,8 +7212,8 @@ static int stbi__psd_is16(stbi__context *s)
        stbi__rewind( s );
        return 0;
    }
-   (void) stbi__get32be(s);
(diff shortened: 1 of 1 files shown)


nothings:
Instead of casting to void, there is an STBI_NOTUSED macro for this.

jackodirks:
Thank you for the feedback. I changed my commit (hence the force push) and also replaced the `(void)` part with STDI_NOTUSED somewhere else in the header.

rygorous:

Issue #1019, opened 2020-08-19
stb_image.h: Compiler warning (GCC 10.1)


mgerhardy:
```
../../src/modules/image/stb_image.h: In function ‘int stbi__zhuffman_decode_slowpath(stbi__zbuf*, stbi__zhuffman*)’:
../../src/modules/image/stb_image.h:4123:10: warning: comparison of integer expressions of different signedness: ‘int’ and ‘long unsigned int’ [-Wsign-compare]
 4123 |    if (b >= sizeof (z->size)) return -1; // some data was corrupt somewhere!
      |        ~~^~~~~~~~~~~~~~~~~~~
```

randomMesh:
I was about to report the same warning and found this issue.
However, there are 2 more warnings which are
```bash
src/video/stb_image.h: In function ‘void* stbi__load_gif_main(stbi__context*, int**, int*, int*, int*, int*, int)’:
src/video/stb_image.h:6778:11: warning: variable ‘out_size’ set but not used [-Wunused-but-set-variable]
       int out_size = 0;
           ^~~~~~~~
src/video/stb_image.h:6779:11: warning: variable ‘delays_size’ set but not used [-Wunused-but-set-variable]
       int delays_size = 0;

```

rygorous:
This is a duplicate and there are other bugs (and open PRs) for this already, closing this one.

PR #1024, opened 2020-09-02
Added cast to match types

Bargor:
Fixes issue #1019 : 
../../src/modules/image/stb_image.h: In function ‘int stbi__zhuffman_decode_slowpath(stbi__zbuf*, stbi__zhuffman*)’:
../../src/modules/image/stb_image.h:4123:10: warning: comparison of integer expressions of different signedness: ‘int’ and ‘long unsigned int’ [-Wsign-compare]
 4123 |    if (b >= sizeof (z->size)) return -1; // some data was corrupt somewhere!
      |        ~~^~~~~~~~~~~~~~~~~~~


rygorous:
This bug is fixed (in a different way) in the dev branch, closing this PR.

Issue #1020, opened 2020-08-19
stb_image_write.h: Compiler warning (GCC 10.1)


mgerhardy:
```
../../src/modules/image/stb_image_write.h: In function ‘void stbiw__write1(stbi__write_context*, unsigned char)’:
../../src/modules/image/stb_image_write.h:400:24: warning: comparison of integer expressions of different signedness: ‘int’ and ‘long unsigned int’ [-Wsign-compare]
  400 |    if (s->buf_used + 1 > sizeof(s->buffer))
      |        ~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~
../../src/modules/image/stb_image_write.h: In function ‘void stbiw__write3(stbi__write_context*, unsigned char, unsigned char, unsigned char)’:
../../src/modules/image/stb_image_write.h:408:24: warning: comparison of integer expressions of different signedness: ‘int’ and ‘long unsigned int’ [-Wsign-compare]
  408 |    if (s->buf_used + 3 > sizeof(s->buffer))
      |        ~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~
../../src/modules/image/stb_image_write.h: In function ‘int stbi_write_bmp_to_func(void (*)(void*, void*, int), void*, int, int, int, const void*)’:
../../src/modules/image/stb_image_write.h:502:32: warning: missing initializer for member ‘stbi__write_context::context’ [-Wmissing-field-initializers]
  502 |    stbi__write_context s = { 0 };
      |                                ^
../../src/modules/image/stb_image_write.h:502:32: warning: missing initializer for member ‘stbi__write_context::buffer’ [-Wmissing-field-initializers]
../../src/modules/image/stb_image_write.h:502:32: warning: missing initializer for member ‘stbi__write_context::buf_used’ [-Wmissing-field-initializers]
../../src/modules/image/stb_image_write.h: In function ‘int stbi_write_bmp(const char*, int, int, int, const void*)’:
../../src/modules/image/stb_image_write.h:510:32: warning: missing initializer for member ‘stbi__write_context::context’ [-Wmissing-field-initializers]
  510 |    stbi__write_context s = { 0 };
      |                                ^
../../src/modules/image/stb_image_write.h:510:32: warning: missing initializer for member ‘stbi__write_context::buffer’ [-Wmissing-field-initializers]
[...]

PR #1000, opened 2020-07-20
stb_image_write.h: Fixed compiler warnings

tay10r:
Replicate with gcc -Wall -Wextra -fsyntax-only stb_image_write.h

PR #1029, opened 2020-09-28
stb_image_write.h: fixed some more warnings (GCC v8.3.0)

RufUsul:
mgerhardy opened an issue for these in https://github.com/nothings/stb/issues/1020
ghost already fixed the other ones in #1000


rygorous:
This "fix" is incorrect. The write_contexts should be initialized, leaving them uninitialized and hoping for the best is not OK.

PR #1031, opened 2020-10-01
fix building by MSVC for Windows 10 on ARM

eugenegff:
I recently enabled NEON support for Windows 10 on ARM for Ogre3D project, https://github.com/OGRECave/ogre-next/, and this is all that was needed to build STB with MSVC for ARM64.


rygorous:
This is now merged on the dev branch and will be in the next version.

PR #1035, opened 2020-10-14
stb_truetype: fix disregarded GPOS kern pairs

kawa-yoiko:
stb_truetype incorrectly drops some kerning in the GPOS table (example at the end). The OpenType spec ([related section](https://docs.microsoft.com/en-us/typography/opentype/spec/chapter2#class-definition-table)) specifies that in a ClassDef table, "All glyphs not assigned to a class fall into Class 0." In stb_truetype however, the function `stbtt__GetGlyphClass` returns -1 for glyphs with unassigned classes, so all their kerning info are disregarded. This patch fixes the bug.

Example: extract the kerning info from [Mali-Regular.ttf](https://github.com/google/fonts/blob/master/ofl/mali/Mali-Regular.ttf).

```c
#define STB_TRUETYPE_IMPLEMENTATION
#include "stb_truetype.h"
#include <stdio.h>

int main()
{
  stbtt_fontinfo font;
  static unsigned char buffer[1 << 20];

  fread(buffer, 1, sizeof buffer, fopen("Mali-Regular.ttf", "rb"));
  stbtt_InitFont(&font, buffer, 0);

  printf("VA %d\n", stbtt_GetCodepointKernAdvance(&font, 'V', 'A'));
  printf("AV %d\n", stbtt_GetCodepointKernAdvance(&font, 'A', 'V'));
  printf("CA %d\n", stbtt_GetCodepointKernAdvance(&font, 'C', 'A'));
  printf("AC %d\n", stbtt_GetCodepointKernAdvance(&font, 'A', 'C'));

  return 0;
}
```

Actual output:
```plain
VA -150
AV 0
CA 0
AC 0
```

Expected output (cross-checked with FontForge, fontTools `ttx`, and multiple GUI applications; corrected by the patch):
```plain
VA -150
AV -150
CA 0
AC -60
```

Details: the glyph "A" appears in the Coverage table but not in the ClassDef1 table, and thus falls into Class 0. Kern pairs with Class 0 as the first glyph are ignored in the unpatched version.

### DIFF
--- a/stb_truetype.h
+++ b/stb_truetype.h
@@ -2486,7 +2486,7 @@ static stbtt_int32  stbtt__GetGlyphClass(stbtt_uint8 *classDefTable, int glyph)
         } break;
     }
 
-    return -1;
+    return 0;
 }
 
 // Define to STBTT_assert(x) if you want to break on unimplemented formats.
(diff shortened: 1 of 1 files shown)


rygorous:
This issue is now fixed in the dev branch, but with a different patch that also fixes some other issues in the surrounding GPOS handling code. Closing this PR, fix will land in a release soon.

PR #1037, opened 2020-10-16
stb_truetype Added a function for getting the default english name of a font

Cryru:

### DIFF
--- a/stb_truetype.h
+++ b/stb_truetype.h
@@ -1036,6 +1036,12 @@ STBTT_DEF const char *stbtt_GetFontNameString(const stbtt_fontinfo *font, int *l
 //     http://developer.apple.com/textfonts/TTRefMan/RM06/Chap6name.html
 //     http://www.microsoft.com/typography/otspec/name.htm
 
+STBTT_DEF const char *stbtt_GetFontNameDefault(const stbtt_fontinfo *font, int *length);
+// return the default english full name of the font as utf-8
+
+STBTT_DEF const char* stbtt__ConvertUTF16BEtoUTF8(const char* wideSrc, stbtt_int32 srcLen, stbtt_int32* newLen);
+// performs a basic conversion of a big endian utf-16 string to a utf-8 string
+
 enum { // platformID
    STBTT_PLATFORM_ID_UNICODE   =0,
    STBTT_PLATFORM_ID_MAC       =1,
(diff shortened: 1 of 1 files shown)


Cryru:
stbtt_GetFontNameString is cool and all, but I figure most projects don't want to deal with platformID, languageID, encodings and such. How about we add a function that covers most use cases and just returns the name of the font?

Cryru:
That allocation there is probably a bad idea. Maybe we can allocate it once and store the name in the font struct?

nothings:
1. Does this cover must use cases? Do 100% of fonts define this? The lack of information about this is why I didn't implement it in the first place.
2. Why not implement this by calling `stbtt_GetFontNameString`? That is, why iterate the table yourself?

Cryru:
1. For sure there are other english encodings (Mac ones?) but I couldn't find any information on them.
2. I would have to call it multiple times with different arguments, it didn't seem right when I could just loop through once.

nothings:
1. So the problem with your reply to #1 is, what is a user of the library supposed to do. "This probably returns the English name of the font, but it may return NULL, in which case good luck, you'll have to fall back to `stbtt_GetFontNameString` and write all that code anyway." doesn't seem like much of an improvement.

2. Unless you're worried about performance, that doesn't seem like a very good argument. And why worry about performance for a function that's going to be called once per font and probably isn't slow (how big is a typical name table?).

Cryru:
1. I think it covers enough cases to be reliable, of course with the vastness of font variations you can never guarantee it will always work. It sure beats having to research platformId, encodingId and such in addition to storing whether the string you got was UTF16-BE (and which combinations of the above arguments result in one) and then using a special function to check for those cases.

2. That's fair - I'll rework it if we resolve point 1.

nothings:
I guess maybe somebody (not necessarily you) needs to like download 1000 fonts from various sources and check how many it works on. If it's 99% I guess I'm ok with it.

Cryru:
I'm on it

nothings:
The most important (but tedious) thing is that they come from multiple sources. Not 1000 sources, but "enough". Like, my Windows/Fonts directory has 500 ttf files in it, but I wouldn't want to use more than 50-100 of them.

Cryru:
Right.

Regarding point 2, using stbtt_GetFontNameString might not be best idea as it would take away the ability to cast a wide net on some of the argument pairs. (Such as the various Unicode platformId encodings)

Cryru:
I tested it against the Google fonts repository (https://github.com/google/fonts) and got a 100% success rate. Here's a list of the fonts tested: https://pastebin.com/j0WgdyFN
That's 3869 fonts.

nothings:
But that's 3869 fonts from a single source. Even if some of them were authored somewhere else, Google might have a policy about going through and setting an English name field.

Cryru:
Right, I was just noting. There was a nice variety in encodings and arguments in that sample size.
I'll look for more sources.

nothings:
> These failed due to containing non-English characters in the English name field. Handling multi-byte characters in the UTF8 conversion function would solve that.

Sounds good to me. There's already a function call to compare UTF16 to UTF8 strings which works by converting the UTF16 string to UTF8, so adapting that (rather than using a platform-provided function) should be straightforward.

Cryru:
I tried to standardize the return value of stbtt_GetFontNameDefault by converting the string to UTF8 if it isn't. Though my conversion function doesn't handle all cases.

If we don't want the function to perform the conversion automatically but have the user use the appropriate comparison function we would need to indicate in some way what the format of the returned string is.

PR #1040, opened 2020-11-01
Creating meson.build to allow easy use with meson build system

noah1510:
The title says it all.
There is only one new file with 17 lines that probably won't change in the near future.
This file allows developers to easily use stb with the meson build system.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -14,4 +14,6 @@ stb_dep = declare_dependency(
     version: meson.project_version()
 )
 
-meson.override_dependency('stb', stb_dep)
+if meson.version().version_compare('>=0.54.0')
+    meson.override_dependency('stb', stb_dep)
+endif
--- a/meson.build
+++ b/meson.build
@@ -2,7 +2,6 @@ project(
   'stb',
   'c',
   'cpp',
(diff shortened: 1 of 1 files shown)


nothings:
There are two obvious problems with this.

1. I have no idea what the point of this is. What does it accomplish? One of the points of the stb libraries is to not require interaction with the build system at all.

2. It doesnt' scale. There are hundreds of build systems out there. I'm not going to support them all.

noah1510:
> I have no idea what the point of this is. What does it accomplish? One of the points of the stb libraries is to not require interaction with the build system at all.

This file removes the need for a subproject if meson is used as build system because the build system will automatically download the repository and include the header files into the project.
This also removes the need to write scripts that copy the header files, some specific compiler flags or simply having very outdated versions of those headers.

The interaction with the build system is very minimal.
All it does is declare that the whole folder should be include and define the dependency that can be included, so this fill will not need to change.

> It doesn't scale. There are hundreds of build systems out there. I'm not going to support them all.

I get that and I would agree if the files are larger than this.
The change needed is so small and need basically no maintenance that I personally think the advantages are bigger than the loss.

nothings:
Nope, not interested. Feel free to fork if you want.

Issue #994, opened 2020-07-14
stb_image.h error loading bmp from memory.


haozhan:
![stb_bug](https://user-images.githubusercontent.com/7542479/87393373-a9c7a100-c5e0-11ea-9731-8b88dfb6b140.png)

s->buffer_start is an uninitialized abandoned array (pointer) when loading bmp picture from memory, using it for an error.

haozhan:
I took a look at the logbook, this problem has long been feedback and solved. But the old code problem comes back from merging with a branch.
The specific version is:
Merge branch 'stb-image-fuzzing-fixes' of https://github.com/rcgordon… …
Sean Barrett committed 4 days ago

![DevCap_20200717_161949](https://user-images.githubusercontent.com/7542479/87765463-6c664c00-c84a-11ea-8ba7-fc2f107cc705.png)

 the problem of merging the above version.

nothings:
ah, good catch, thanks.

Moros1138:
I can confirm this issue and offer what appear to be, at least on my end, the solution.

My code that I used to test it:

```
data = stbi_load_from_memory((stbi_uc*)vecMemory.data(),.vecMemory.size(), &width, &height, &channels, 4);
```

returns NULL and if you run ``stbi_failure_reason`` you get "bad offset" or "Corrupt BMP" depending on whether the ``STBI_FAILURE_USERMSG`` is defined.

I was able to fix it locally by changing the code at around line 5298 (in the latest header) from: 

```
      if (info.offset != s->callback_already_read + (s->img_buffer - s->buffer_start)) {
```

to the following
```
      if (info.offset != s->callback_already_read + (s->img_buffer - s->img_buffer_original)) {
```

This resolved the error messages and restores the expected behavior of the function. Looking forward to an update.

P.S. Thanks for all the work you and your collaborators put into this project. It is truly awesome!

ITotalJustice:
> img_buffer_original

Thanks mate, your suggestion fixed the issue for me as well.

ClysmiC:
[...]

PR #1044, opened 2020-11-05
Fixes issue #994

JoshMarshall756:
Uses the fix suggested in the comments.

Made a new pull request because I made a mistake in the previous one by branching from another branch and not master. I didn't know how to fix the previous PR tidily.


rygorous:
Fix is pushed in dev branch, will be in the next release.

PR #1045, opened 2020-11-06
Fix comparison of different signs.

stolk:
Addresses a compiler warning on comparing integers of different signs:
```
/home/bram/src/stb/stb_image.h:4123:10: warning: comparison of integers of different signs: 'int' and 'unsigned long' [-Wsign-compare]
   if (b >= sizeof (z->size)) return -1; // some data was corrupt somewhere!
```


rygorous:
A different fix for this is now in the dev branch and will be in the next version, closing this PR.

PR #1048, opened 2020-11-15
Fix warnings under -Wall -Wextra -Wpedantic.

jeannekamikaze:


rygorous:
The second fix (size check) is present in the dev branch (albeit in a different way), the first fix causes warnings on other compilers, closing this PR; if you continue to have a problem with the realloc_sized warning after the recent batch of fixes is released (it should be fixed now anyway but I don't know what compiler/target to test since you didn't say) feel free to open an issue and we'll work it out.

PR #1053, opened 2020-11-26
ossfuzz: improve code coverage

randy408:
* fix .gif pattern matching in ossfuzz.sh
* add netpbm test images derived from the public domain pngsuite
* added more image types (downloaded in [Dockerfile](https://github.com/google/oss-fuzz/blob/master/projects/stb/Dockerfile))


rygorous:
Merged into dev branch, should go live later this week.

PR #1054, opened 2020-12-10
fix C4456 warning

samhattangady:
When I was compiling my program with level 4 warnings I saw this error:
...\stb_ds.h(1422): warning C4456: declaration of 'i' hides previous local declaration
...\stb_ds.h(1362): note: see declaration of 'i'

Link on Microsoft Docs
https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-4-c4456?view=msvc-160

To fix this, I have renamed the variable from `i` to `j` in the scope.
Thanks.

### DIFF
--- a/stb_ds.h
+++ b/stb_ds.h
@@ -379,6 +379,7 @@ CREDITS
     Andreas Molzer
     github:hashitaku
     github:srdjanstipic
+    github:samhattangady
 */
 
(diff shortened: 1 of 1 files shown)


nothings:
thank you for your submission. i fixed this in a different way which was a smaller change (but hopefully still fixes it)

PR #1058, opened 2020-12-14
oss-fuzz: integrate with CIFuzz

randy408:
This adds a GitHub Action to PR's to verify code changes do not introduce new bugs or cause the OSS-Fuzz build to break.
 
For more info: https://google.github.io/oss-fuzz/getting-started/continuous-integration/

### DIFF
--- a/.github/workflows/ci-fuzz.yml
+++ b/.github/workflows/ci-fuzz.yml
@@ -1,0 +1,23 @@
+name: CIFuzz
+on: [pull_request]
+jobs:
+  Fuzzing:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Build Fuzzers
+      uses: google/oss-fuzz/infra/cifuzz/actions/build_fuzzers@master
+      with:
+        oss-fuzz-project-name: 'stb'
+        dry-run: false
+    - name: Run Fuzzers
+      uses: google/oss-fuzz/infra/cifuzz/actions/run_fuzzers@master
+      with:
+        oss-fuzz-project-name: 'stb'
+        fuzz-seconds: 900
+        dry-run: false
+    - name: Upload Crash
+      uses: actions/upload-artifact@v1
+      if: failure()
+      with:
+        name: artifacts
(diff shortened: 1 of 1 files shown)


rygorous:
Merged into dev branch, will go live when we next do a release. (Should be in a few days.)

rygorous:
This does not currently work, see https://github.com/nothings/stb/pull/1161/checks?check_run_id=3028530474.

The `tar xvzf $SRC/stb/jpg.tar.gz` fails even though the earlier wget seems to have completed successfully. It looks like `$SRC` is not set to anything which seems wrong. This looks to be a configuration issue but I have no idea how the Docker containers here are meant to be configured in the first place, whether SRC is supposed to get set to anything, and what sets it if so, and the docs for oss-fuzz I can find don't seem to explain this either.

Can you have a look, randy?

randy408:
> 
> 
> This does not currently work, see https://github.com/nothings/stb/pull/1161/checks?check_run_id=3028530474.
> 
> The `tar xvzf $SRC/stb/jpg.tar.gz` fails even though the earlier wget seems to have completed successfully. It looks like `$SRC` is not set to anything which seems wrong. This looks to be a configuration issue but I have no idea how the Docker containers here are meant to be configured in the first place, whether SRC is supposed to get set to anything, and what sets it if so, and the docs for oss-fuzz I can find don't seem to explain this either.
> 
> Can you have a look, randy?

CIFuzz is overwriting `$SRC/stb`, I have a fix which involves changes to the Dockerfile in the oss-fuzz repo and the build script here, it would fix CIFuzz checks but break OSS-Fuzz builds which pulls from `master`, should I point the Dockerfile to the `dev` branch temporarily?

rygorous:
Is this going to be required for CIFuzz from any branches to work, or is this just a temporary thing while the integration isn't merged to master yet?

If it's the former, a solution would be desirable, if it's just until the current staged changes from dev land in master when we next cut a release (due soon), don't worry about it.

PR #1064, opened 2020-12-16
Fixed vorbis files with no comments emitting outofmemory error

anthofoxo:


nothings:
This appears to be the same issue as #1005, but a very different fix. I'd like to know whether this actually addresses all the cases that fix does.

anthofoxo:
I am unsure if it covers all cases, but in the case of me streaming audio it does.

sagamusix:
I arrived at the same conclusion as this PR and would say that it's more elegant than the one in the other PR.

anthofoxo:
I added a small commit to my branch will will guard the call to setup_malloc as suggested in my old thread

stb_vorbis.c
```diff
@@ -3642,7 +3643,9 @@ static int start_decoder(vorb *f)
    f->vendor[len] = (char)'\0';
    //user comments
    f->comment_list_length = get32_packet(f);
-   f->comment_list = (char**)setup_malloc(f, sizeof(char*) * (f->comment_list_length));
+   f->comment_list = NULL;
+   if (f->comment_list_length > 0)
+       f->comment_list = (char**) setup_malloc(f, sizeof(char*) * (f->comment_list_length));
    if (f->comment_list == NULL)                     return error(f, VORBIS_outofmem);
```
DanielGibson, line 3649: I don't think this works.
In 3646 you set it to NULL, afterwards you only set it to something else if length > 0, and here you still `return error(f, VORBIS_outofmem);` if it's NULL, so loading vorbis files with no comment will still fail.

DanielGibson:
for this to work the `if (f->comment_list == NULL)  return error(..)` needs to be guarded by `if (f->comment_list_length > 0)` as well.

sagamusix:
Alternatively, `if (f->comment_list_length > 0 && f->comment_list == NULL) return error(f, VORBIS_outofmem);` (this is what I used when I said above that I arrived at the same fix, but it's in fact slightly different).

stb_vorbis.c
anthofoxo: thanks for noticing that, i missed that, ill update the pr in a few

anthofoxo:
i went back to fix that little issue, should be all good now

DanielGibson:
yes, this looks good!

nothings:
Thank you for this submission, my apologies for how long it took to integrate.

PR #1066, opened 2020-12-17
fix stb's illegal memory access when find glyph beyond glyph table

leoxupku:


rygorous:
Not using this particular PR but a slightly different fix is pushed to the dev branch and will be in the next release.

PR #1067, opened 2020-12-23
Vorbis fix

Phildo:
If `comment_list_length` returns 0, it previously attempted to malloc 0 bytes, and interpreted that failing as `VORBIS_outofmem`. This fix instead checks if the length is 0, and skips any malloc for comment_list.


Phildo:
Ah- just realized a similar pull request already exists (https://github.com/nothings/stb/pull/1064). Nevermind!

PR #1069, opened 2020-12-30
stb_truetype: fixes SDF rendering artifacts with some fonts

TobiasBohnen:
Affects e.g. DejaVuSans' "u" and Calibri's "&".

Before and after:
dejavusans

Fix based on: https://bugzilla.gnome.org/show_bug.cgi?id=665384


nothings:
This is already correctly handled on the rasterization path, so I'd prefer not to make changes that potentially affect that path as well. (For example, I don't think this implementation will handle the case of multiple move_tos at the end). So I'll try to fix it directly in the SDF generator.

nothings:
Fixed in dev branch.

Before:
image

After:
image

PR #1074, opened 2021-01-09
Fix compiler error when using STB_C_LEX_0_IS_EOF in stb_c_lexer.h

TheSandvichMaker:
On line 611 in stb_c_lexer.h there is a variable "tok" used which does not exist, and should be "lexer" instead.


nothings:
Fixed independently, thank you for your PR anyway.

PR #1075, opened 2021-01-09
Replaced ldexp when decoding .hdr images with bit op equivalent

TheSandvichMaker:
This speeds up loading .hdr images somewhere between 2.6x-2.8x on my machine.

### DIFF
--- a/stb_image.h
+++ b/stb_image.h
@@ -558,7 +558,7 @@ STBIDEF int   stbi_zlib_decode_noheader_buffer(char *obuffer, int olen, const ch
 #include <limits.h>
 
 #if !defined(STBI_NO_LINEAR) || !defined(STBI_NO_HDR)
-#include <math.h>  // ldexp, pow
+#include <math.h>  // pow
 #endif
 
 #ifndef STBI_NO_STDIO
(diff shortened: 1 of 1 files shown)


nothings:
This is not a safe way to do bit operations between float and int. Current practice is to use a union to convert between them, although there are indications that compiler writers have decided that that is also undefined behavior as well and there's some other thing we're supposed to be doing, I don't even remember what that is.

TheSandvichMaker:
Probing the Molly Rocket discord, the answer for C89 appears to be memcpy, union isn't considered safe until C99.
Modern C/C++ compilers seem to happily do the right thing with memcpy:

C: https://godbolt.org/z/oYTb8oj9M
C++: https://godbolt.org/z/Msdn5393G

nothings:
One concern is that on older compilers they may explicitly do a memory copy, which will _behave_ correctly, but maybe will be even slower than ldexp?

TheSandvichMaker:
Yeah, that is a relevant point. Testing the oldest version available on godbolt of each of the "major three" still behaves correctly, but I can't speak for VC6 for example. Since type punning through unions appears to be happening in the stb codebases anyway, should I just switch to that?

nothings:
I don't think VC6 performance is worth holding thing sup, so if it works on oldest versions on godbolt, I'm ok with that.

N-R-K:
> union isn't considered safe until C99.

Hi, C11 adds following footnote (95) regarding type punning through unions. I don't see anything similar in C99, so it's probably unsafe in C99 as well unless I've missed something.

> If the member used to read the contents of a union object is not the same as the member last used to store a value in the object, the appropriate part of the object representation of the value is reinterpreted as an object representation in the new type as described in 6.2.6 (a process sometimes called "type punning"). This might be a trap representation.

PR #1081, opened 2021-01-25
gcc fixes

goblinhack:


nothings:
do these warnings occur without -Wall ?

goblinhack:
Good point - only with -Wall I think

goblinhack:
There are some others too that popped up with latest ubuntu and -Wall - not sure how to work around these: fyi:

In file included from stb_image.cpp:3:
stb_image.h: In function 'void* stbi__load_gif_main(stbi__context*, int**, int*, int*, int*, int*, int)':
stb_image.h:6778:11: warning: variable 'out_size' set but not used [-Wunused-but-set-variable]
 6778 |       int out_size = 0;
      |           ^~~~~~~~
stb_image.h:6779:11: warning: variable 'delays_size' set but not used [-Wunused-but-set-variable]
 6779 |       int delays_size = 0;
      |           ^~~~~~~~~~~

nothings:
I don't generally fix things that only show up with -Wall.

goblinhack:
np

Issue #1087, opened 2021-02-01
stb_image_write: possibility to avoid use of global variables


flriancu:
Currently, there are some global variables in `stb_image_write`, including:
```
stbi_write_png_compression_level
stbi_write_force_png_filter
stbi_write_tga_with_rle
```
To avoid using these variables, I made a patch that:
- adds four new public functions:
```
STBIWDEF int stbi_write_png_prefs(char const *filename, int w, int h, int comp, const void *data, int stride_bytes, const stbiw_png_prefs *prefs);
STBIWDEF int stbi_write_tga_prefs(char const *filename, int w, int h, int comp, const void *data, const stbiw_tga_prefs *prefs);

STBIWDEF int stbi_write_png_to_func_prefs(stbi_write_func *func, void *context, int w, int h, int comp, const void *data, int stride_bytes, const stbiw_png_prefs *prefs);
STBIWDEF int stbi_write_tga_to_func_prefs(stbi_write_func *func, void *context, int w, int h, int comp, const void *data, const stbiw_tga_prefs *prefs);
```
- adds one new internal function:
```
STBIWDEF unsigned char *stbi_write_png_to_mem_prefs(const unsigned char *pixels, int stride_bytes, int x, int y, int n, int *out_len, const stbiw_png_prefs *prefs)
```
- and changes one internal function:
```
static int stbi_write_tga_core(stbi__write_context *s, int x, int y, int comp, void *data, const stbiw_tga_prefs *prefs)
```

The `stbiw_*_prefs` are structures designed to allow backward-compatibility in the future via their `structSize` member.

The pull request should link automatically to this issue.

Please let me know what you think. Thanks!

flriancu:
One immediate advantage is, you can call `stbi_write_png_prefs` from multiple threads with different compression levels. Unless something major escapes me, these `*_prefs` functions are thread-safe (except for the remaining issue of `stbi__flip_vertically_on_write`).

nothings:
If the goal is to introduce thread-compatability for these global variables, I'd prefer to do it the way we do it in stb_image.c with thread-locals, rather than by significantly increasing the number of main API entrypoints, which adds a lot of complexity for users to understand with very little benefit.

nothings:
But really it might be better just to remove these variables from the public API (but possibly keep them as undocumented), as I doubt they receive signifcant usage.

flriancu:
> If the goal is to introduce thread-compatability for these global variables
[...]

PR #1088, opened 2021-02-01
avoid global vars for PNG and TGA

flriancu:
Hi!

This PR relates to ticket #1087 .

As far as I can tell, this patch does not break the existing API.

Please let me know what you think. Thanks!


flriancu:
I'm not sure if `stbi_write_png_to_mem` is (also) meant to be used as an external function. If not, I reckon it can be deleted.

flriancu:
Closing per discussion in ticket #1087 .

PR #1095, opened 2021-02-13
Prevent out of memory error and return when the vorbis file has no comment list

VirtuosoChris:


huwpascoe:
👍 Please merge

DanielGibson:
This is one of three pull requests for this issue; I'd prefer this one as it's shortest (but correct, as far as I can tell).

@nothings: Please merge a fix for this issue, not being able to load .ogg files that have no comment is a bit frustrating, even more so as one needs to debug it to figure out *why* it fails (with `VORBIS_outofmem`).
Thanks in advance! :-)

rygorous:
A different fix for this was merged into mainline 2 days ago, closing this PR.

PR #1102, opened 2021-02-20
Fix stbi_load_from_memory of stbi__bmp_load

DaydreamCoding:
stbi_load_from_memory do not have buffer

### DIFF
--- a/stb_image.h
+++ b/stb_image.h
@@ -5395,7 +5395,7 @@ static void *stbi__bmp_load(stbi__context *s, int *x, int *y, int *comp, int req
    }
    if (psize == 0) {
       STBI_ASSERT(info.offset == s->callback_already_read + (int) (s->img_buffer - s->img_buffer_original));
-      if (info.offset != s->callback_already_read + (s->img_buffer - s->buffer_start)) {
+      if (s->io.read && info.offset != s->callback_already_read + (s->img_buffer - s->buffer_start)) {
         return stbi__errpuc("bad offset", "Corrupt BMP");
       }
    }


zchrissirhcz:
@nothings Hi, I met same problem and was trying make an Pull Request ... and just find this unmerged PR, and it really works. Could we merge this PR? Thanks.

zchrissirhcz:
To reproduce this issue, use this code:

```c++
#define STBI_NO_PSD
#define STBI_NO_TGA
#define STBI_NO_GIF
#define STBI_NO_HDR
#define STBI_NO_PIC
#include "stb_image.h"

#include "stb_image_write.h"

#include <stdlib.h>
#include <string.h>
#include <string>

// read "IU.bmp" image by stbi_load_from_memory()
// then save as "result.png"
int main() {
    const char* filename = "IU.bmp";
    const char* savename = "result.png";

    int height;
    int width;
    int channels;
    
#if 0
    unsigned char* raw_data = stbi_load(filename, &width, &height, &channels, 0);
#else
    FILE* fin = fopen(filename, "rb");
    if (fin == NULL) {
        LOGE("failed to open file %s\n", filename);
        exit(1);
    }
    fseek(fin, 0, SEEK_END);
    long buffer_size = ftell(fin);
    rewind(fin);
    unsigned char* buffer = (unsigned char*)malloc(buffer_size);
    if (buffer == NULL) {
        LOGE("failed to allocate memory\n");
        exit(2);
    }
    size_t result = fread(buffer, 1, buffer_size, fin);
    if (result != buffer_size) {
        LOGE("failed to read from file %s\n", filename);
        exit(3);
    }
    fclose(fin);

    unsigned char* raw_data =
        stbi_load_from_memory(buffer, buffer_size, &width, &height, &channels, 0);
     free(buffer);
     buffer = NULL;
#endif

    printf("message: %s\n", stbi_failure_reason());
    printf("image info: width=%d, height=%d, channels=%d\n", width, height, channels);
    
    int stride_in_bytes = width * 3;
    stbi_write_png(savename, width, height, 3, raw_data, stride_in_bytes);

    free(raw_data);

    return 0;
}
```

rygorous:
This is the wrong fix; see issue #994 for the correct one. Either way, this is fixed in dev branch and will be in the next release.

PR #1106, opened 2021-02-28
stb_image_resize: Fix edge reflect overflow

jspohr:
Noticed that my lowest mipmaps contained random colors after switching to STBIR_EDGE_REFLECT. I believe this is the correct fix, but please correct me if I'm wrong. Cheers!

### DIFF
--- a/stb_image_resize.h
+++ b/stb_image_resize.h
@@ -959,7 +959,7 @@ static int stbir__edge_wrap_slow(stbir_edge edge, int n, int max)
     {
         if (n < 0)
         {
-            if (n < max)
+            if (-n < max)
                 return -n;
             else
                 return max - 1;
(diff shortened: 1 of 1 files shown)


sezero:
Has this patch been reviewed / well-tested?

jspohr:
Reviewed - not as far as I know.
Tested - it worked for me on Windows and Linux. We were using the EDGE_REFLECT mode only rarely, though.

stb_image_resize.h
```diff
@@ -959,7 +959,7 @@ static int stbir__edge_wrap_slow(stbir_edge edge, int n, int max)
     {
         if (n < 0)
         {
-            if (n < max)
```
jspohr, line 962: This condition is always true, because `n` is negative here, and `max` is always positive.
The overflow happens if `-n >= max`: the function still returns `-n` instead of `max - 1`.

sezero:
OK, thanks.

jspohr:
Looks like this was fixed in stb_image_resize2

PR #1110, opened 2021-03-11
Note GL blend state for stb_truetype

fluffels:
Maybe worth adding these as a note to the example in the documentation. Don't know about other people but I always forget them :)


rygorous:

PR #1113, opened 2021-03-13
stb_vorbis bug fix: audiofiles with empty comment section was unable to load

romanshuvalov:
Fixed an error which triggered VORBIS_outofmem when loading a file without comment section (comment_list_length==0). Also fixed couple of warnings about unused parameters.

### DIFF
--- a/stb_vorbis.c
+++ b/stb_vorbis.c
@@ -33,6 +33,7 @@
 //    Timur Gagiev       Maxwell Koo         Peter Waller
 //    github:audinowho   Dougall Johnson     David Reid
 //    github:Clownacy    Pedro J. Estebanez  Remi Verschelde
+//    Roman Shuvalov
 //
 // Partial history:
 //    1.20    - 2020-07-11 - several small fixes
@@ -904,6 +905,7 @@ static int error(vorb *f, enum STBVorbisError e)
    return 0;
 }
 
+#define UNUSED_PARAMETER(x) (void)(x)
 
 // these functions are used for allocating temporary memory
 // while decoding. if you can afford the stack space, use
@@ -3069,6 +3071,7 @@ static int do_floor(vorb *f, Mapping *map, int i, int n, float *target, YTYPE *f
       for (q=1; q < g->values; ++q) {
          j = g->sorted_order[q];
          #ifndef STB_VORBIS_NO_DEFER_FLOOR
+         UNUSED_PARAMETER(step2_flag);
          if (finalY[j] >= 0)
          #else
          if (step2_flag[j])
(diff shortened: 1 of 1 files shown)


stb_vorbis.c
```diff
@@ -3642,18 +3647,23 @@ static int start_decoder(vorb *f)
...
-         f->comment_list[i][j] = get8_packet(f);
-      }
-      f->comment_list[i][len] = (char)'\0';
+          for(j=0; j < len; ++j) {
+             f->comment_list[i][j] = get8_packet(f);
+          }
+          f->comment_list[i][len] = (char)'\0';
+       }
+   }
+   else {
```
fluffrabbit, line 3665: `else` should be on the same line as the closing bracket.

stb_vorbis.c
```diff
@@ -4191,10 +4201,12 @@ static void vorbis_deinit(stb_vorbis *p)
...
-   for (i=0; i < p->comment_list_length; ++i) {
-      setup_free(p, p->comment_list[i]);
-   }
-   setup_free(p, p->comment_list);
+   if (p->comment_list != NULL) {
+       for (i=0; i < p->comment_list_length; ++i) {
+          setup_free(p, p->comment_list[i]);
+       }
+       setup_free(p, p->comment_list);
+   };
```
fluffrabbit, line 4209: Unnecessary semicolon?

stb_vorbis.c
```diff
@@ -904,6 +905,7 @@ static int error(vorb *f, enum STBVorbisError e)
    return 0;
 }
 
+#define UNUSED_PARAMETER(x) (void)(x)
```
fluffrabbit, line 908: This is a matter of opinion, but is defining `UNUSED_PARAMETER` necessary, as most people know what `(void)foo` does? I could see the use case of searching for the macro to find dead parameters, but only in cases where it's an actual dead parameter and not something that is compiled conditionally.

fluffrabbit (commented):
The latest revision of mainline stb_vorbis broke files without comments. Something to do with it trying to throw errors and whatnot. I want this bug fixed, but I have a couple of comments on this PR.

I'll probably use this patch in production like now, so thank you in any case. :)

ROTARTSI82:
I noticed this bug in my code when it suddenly stopped working :\

Also, this can be fixed by only changing this to ` if (f->comment_list_length && f->comment_list == NULL) ...`
https://github.com/nothings/stb/blob/c9064e317699d2e495f36ba4f9ac037e88ee371a/stb_vorbis.c#L3646
and adding that null check. Why is free called in a loop instead of just one free call to `comment_list`?

nothings:
Integrated a different PR with this fix. Thank you for reporting, my apologies for how long it took to address.

PR #1114, opened 2021-03-15
Replaced macro _MSV_VER to _WIN32 inside readdir_raw

StellaSmith:
This was done to fix a compilation error in msys2.


rygorous:
This is now fixed in the dev branch, although not via this PR.

PR #1130, opened 2021-04-09
 stb_tilemap_editor.h - Re-added calls to `stbte__hittest()`; fixed some compiler errors.

wernsey:
Some earlier commit removed calls to `stbte__hittest()`, which meant that mouse clicks were no longer working.

I just added those calls back from an older version, but removed the `over` local variables which were causing compiler warnings.

I also fixed some GCC warnings about unused variables when `STBTE__COLORPICKER` and `STBTE_ALLOW_LINK` is not defined

Also, the `STBTE_DRAW_TILE(x,y,t->id, slot == tm->cur_tile,0);` statement at line 3519 caused a compiler error because the line that declared it was removed, so I added it back (line 3512).


rygorous:
LGTM, merged into the dev branch with minor changes, will be in the next release.

PR #1133, opened 2021-04-16
tilemap_editor: Fix undefined variables

RobLoach:
I'm not sure if this is correct, but it does fix compilation and rendering.


rygorous:
Fix LGTM. Merged into dev branch, will be in the next release.

RobLoach:
Thanks!

Issue #512, opened 2017-10-15
stb_truetype.h: Implement support for Embedded Bitmaps.


kphillisjr:
This is mostly a request be able to support glyphs from TrueType fonts that use Bitmap data. 

- [ ] Add Support for Embedded Monochrome bitmap data.
  - [ ] add support for [Embedded Bitmap Location Table (EBLC)](https://www.microsoft.com/typography/otspec/eblc.htm)
   - [ ] Implement [Embedded Bitmap Data Table (EBDT)](https://www.microsoft.com/typography/otspec/ebdt.htm)
    - [ ] **Optional:** Implement Apple's Apple Advanced Typography (AAT) 'bloc' table - is done by copying and expanding the code used to work with the 'EBLC ' table.
    - [ ] **Optional:** Implement Apple's Apple Advanced Typography (AAT) 'bdat' table - This is an expanded form of the 'EBDT' table. That said, the EBDT and bdat tables are not completely compatible.
- [ ] Add support for Embedded Color Bitmap Data.
  - [ ] Implement [Color Bitmap Location Table('CBLC')](https://www.microsoft.com/typography/otspec/cblc.htm)
  - [ ] Implement [Color Bitmap Data Table ('CBDT')](https://www.microsoft.com/typography/otspec/cbdt.htm) support.
    - [ ] Add Support for uncompressed Data - This is in an BGRA format.
    - [ ] Add support for Compressed Data - This is a subset of the PNG Format where specified PNG headers are included.
- [ ] Implement [Embedded Bitmap Scaling Table ('EBSC')](https://www.microsoft.com/typography/otspec/ebsc.htm)
- [ ] Add Support for the [Standard Bitmap Graphics Table('sbix')](https://www.microsoft.com/typography/otspec/sbix.htm).
  - [ ] Gracefully Notify user of unsupported formats/tags. This table could define data using pdf data.
  - [ ] Add support for JPEG Encoded Data. This uses the 'jpg ' tag.
  - [ ] Add support for PNG Encoded Data. This uses the 'png ' tag.
  - [ ] **Optional:** Add support TIFF Data. This uses the 'tiff' tag, and currently stb_image.h does not support this file format.
- [ ] Add Support for Color Pallets - This is not actually a Bitmap format, but instead is a method for using Multiple glyph strikes to color an image. 
  - [ ] Implement [Color Table(COLR)](https://www.microsoft.com/typography/otspec/colr.htm)
[...]

PR #1135, opened 2021-04-18
Glyph layered color font support (CPAL/COLR)

pabloko:
Continuing from https://github.com/nothings/stb/issues/512 this PR adds support for opentype's extension tables CPAL and COLR 

https://docs.microsoft.com/es-es/typography/opentype/spec/cpal (version 0 as it seems theres no fonts using CPAL v1)
https://docs.microsoft.com/es-es/typography/opentype/spec/colr

Also keeping an eye on CPAL/COLR spec proposals
https://github.com/googlefonts/colr-gradients-spec/blob/main/OFF_AMD2_WD.md

The tables load the first time when this new api is called imitating the SVG loading mechanism, Color palettes can also be used with SVG data as per specification. During the first lookup of COLR table i've choosen to pre swap the bytes on the Layer Records Table in order to be able to cast this memory for the operation result. This was not needed on CPAL since the union allows to access separate color components and BGRA color is what opengl/d3d uses. This may cause trouble in some marginal cases using 3rd party libs (harfbuzz-ot-color) to read this table AND using this api subset at the same time, i think the benefit of pre-swapping the table is far superior so the user dont have to use more api to read the values.

This api has been tested with:  `Segoe UI Emoji` (old/new),  `FlagsWorldColor`,  `FirefoxEmoji`, `RocherColor` and others.

If someone has a font specimen with CPAL v1 please reach me to update this PR, but this extension isn't very juicy anyway. (palette flags for dark/clear bg, and names)

New api proposed:
```c
//// Color palettes (CPAL) ///////////////////////////////////////////////////
typedef union {
    unsigned int color;
    struct {
        unsigned char b, g, r, a;
    } argb;
} stbtt_color;
STBTT_DEF bool stbtt_FontHasPalette(const stbtt_fontinfo *info);
STBTT_DEF unsigned short stbtt_FontPaletteCount(const stbtt_fontinfo *info);
STBTT_DEF unsigned short stbtt_FontPaletteGetColors(const stbtt_fontinfo *info, unsigned short paletteIndex, stbtt_color **colorPalette);
//// Glyph layers (COLR) /////////////////////////////////////////////////////
typedef struct {
    unsigned short glyphid, colorid;
} stbtt_glyphlayer;
STBTT_DEF bool stbtt_FontHasLayers(const stbtt_fontinfo *info);
STBTT_DEF unsigned short stbtt_GetGlyphLayers(const stbtt_fontinfo *info, unsigned short glypId, stbtt_glyphlayer **glyphLayer);
STBTT_DEF unsigned short stbtt_GetCodepointLayers(const stbtt_fontinfo *info, unsigned short codePoint, stbtt_glyphlayer **glyphLayer);
```

Test harness results (d3d9) - Original + Color + Glyph layers + Palette
- SegoeUI Emoji (new) 

- RocherColor "HelloSTB" where each letter uses a different palette
image

### DIFF
--- a/stb_truetype.h
+++ b/stb_truetype.h
@@ -1499,6 +1499,8 @@ static int stbtt_InitFont_internal(stbtt_fontinfo *info, unsigned char *data, in
       info->numGlyphs = 0xffff;
 
    info->svg = -1;
+   info->colr = -1;
+   info->cpal = -1;
 
    // find a cmap encoding table we understand *now* to avoid searching
    // later. (todo: could make this installable)
--- a/stb_truetype.h
+++ b/stb_truetype.h
@@ -39,6 +39,7 @@
 //       github:IntellectualKitty
 //       Imanol Celaya
 //       Daniel Ribeiro Maciel
+//       Pablo Garcia "Pabloko"
 //
 //   Bug/warning reports/fixes:
 //       "Zer" on mollyrocket       Fabian "ryg" Giesen   github:NiLuJe
@@ -58,6 +59,7 @@
 //
 // VERSION HISTORY
 //
+//   1.25 (2021-04-17) support for glyph layered color fonts (CPAL/COLR)
 //   1.24 (2020-02-05) fix warning
 //   1.23 (2020-02-02) query SVG data for glyphs; query whole kerning table (but only kern not GPOS)
 //   1.22 (2019-08-11) minimize missing-glyph duplication; fix kerning if both 'GPOS' and 'kern' are defined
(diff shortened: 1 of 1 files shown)


nothings:
You're gonna have to fork the library if you want to do this. I don't want to expand the feature set of stb_truetype. I don't wnat to have it be a loading interface that doesn't implement the corresponding functionality, AND I don't want to implement or even maintain the corresponding functionality either.

pabloko:
Well I respect if you dont want to expand the feature set, but the last year SVG was added and it falls on the same category as COLR, or even worse because with SVG theres no rendering mechanism while glyph rendering is the core of this lib.

I don't understand the second part of your comment either... the interface i've posted is just a snip of the commited code, that is actually implemented and tested for full coverage. Did you took a look at the changes?

nothings:
You implemented the ability for clients of stbtt to access the data in these tables themselves, you didn't implement anything to DO anything with those tables, e.g. render colored bitmaps. stb_truetype is not a library for accessing data out of tables, it is a library for making font bitmaps. It doesn't make any sense to add functionality accessing data but not to provide any rendering for it. But I also don't want to support rendering color bitmaps (or SVG, for that matter).

pabloko:
The only rendering api needed for COLR is `stbtt_MakeGlyphBitmap`. The aim of this PR is not to access those tables, but provides palette colors and searching corresponding glyph substitution with arrays of glyphs to compose a layered colored glyph.

PR #1137, opened 2021-04-24
stb_dxt: Initialize tables at compile time

### LINKED ISSUE #1117: is stb_dxt.h thread safe?
I'm not 100% sure that stb_dxt is thread safe, mostly because of this block in stb_compress_dxt_block():
`   
   static int init=1;
   if (init) {
      stb__InitDXT();
      init=0;
   }`

(I think the optimizer could potentially move init=0 in front of stb__InitDXT but I havent exactly tested this extensively, I just don't trust it)

I get I could probably avoid this by throwing a dummy block at stb_compress_dxt_block on the main thread before doing any compression, but that feels a little bit clunky

fahickman:
Also fix a "potentially uninitialized variable" warning.

Resolves #1117


rygorous:
(Comment to self) this will need some minor reworking to include the table generator in some form. (It's fine for the tables to be static data but the code that generates them should be in the repo somewhere.)

rygorous:
Merged into dev branch.

PR #1143, opened 2021-05-11
expose STBRP__MAXVAL define even if STB_RECT_PACK_IMPLEMENTATION isn't defined

rollraw:
e.g. i intend to limit packed rect boundaries by STBRP__MAXVAL for backward compability in my header file (font.h for example) but i doesn't need full implemetation because this header included in other files where stb_rectpack.h will be unused, so it makes me possible to include "std_rectpack.h" in my font.h to use `STBRP__MAXVAL` define and define STB_RECT_PACK_IMPLEMENTATION before including font.h in font.cpp file where implementation used

### DIFF
--- a/stb_rect_pack.h
+++ b/stb_rect_pack.h
@@ -67,6 +67,12 @@
 #define STBRP_DEF extern
 #endif
 
+#ifdef STBRP_LARGE_RECTS
+#define STBRP__MAXVAL  0xffffffff
+#else
+#define STBRP__MAXVAL  0xffff
+#endif
+
 #ifdef __cplusplus
 extern "C" {
 #endif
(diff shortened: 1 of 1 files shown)


nothings:
STBRP_LARGE_RECTS is a configuration macro that is only required to be defined in the file that creates the implementation. (I.e. it can be a #define in that file, rather than a global macro defined in the build system.) This is a consistent design feature in all stb libraries to make it easier to use.

This means that during the "header file" section of stb_rectpack, the value of STBRP_LARGE_RECTS is unknown. If it's in the implementation file, it must be the final value, but anywhere else, it's unknown.

Since STBRP_MAXVAL is computed from STBRP_LARGE_RECTS, it's impossible to know in the header file what the value of STBRP_LARGE_RECTS is.

If font.h is a single-file library, that means it's impossible to automatically know what the value of STBRP__MAXVAL is in your header file at compile time.

If you need it at compile time (e.g. to size a data structure), the best approach is to require the user to pass it in to you: require '#define FONT_SBTRP_LARGE_RECTS' or '#define FONT_STBRP_MAXVAL <n>' be defined by the user.

If you only need it at run time (e.g. to rangecheck a parameter), we can add a function to stb_rectpack that returns STBRP__MAXVAL.

If font.h isn't part of a single-file library, then just your own MAXVAL define for your own use. Since the values are determined by datatype sizes, the values in stb_rectpack will never change.

rollraw:
thanks for answering
im not fully understand then why `stbrp_coord` is exposed before STB_RECT_PACK_IMPLEMENTATION when it also depends on STBRP_LARGE_RECTS
https://github.com/nothings/stb/blob/master/stb_rect_pack.h#L78-L82

rygorous:
This functionality is now in the dev branch, although with a slightly different implementation, and will be in the next release. Closing this PR.

PR #1145, opened 2021-06-01
shputs duplicate insert: fix stale temp_key

makuto:
Fix `temp_key` being stale on key re-insert. This pull request re-uses the existing key, which matches the previous behavior. Before the fix, there was a chance that `temp_key` would be the last inserted key, or effectively garbage data. This manifested for me in double frees and entries with incorrect keys.

See issue nothings#992 and pull request nothings#993.


makuto:
To reproduce the double-free issue:
```C
    s.key = "a", s.value = 1;
    sh_new_strdup(strmap);
    shputs(strmap, s);
    s.key = "b", s.value = 1;
    shputs(strmap, s);
    // At this moment, the "a" entry will actually receive the stale "b" temp_key
    s.key = "a", s.value = 1;
    shputs(strmap, s);```

PR #1148, opened 2021-06-11
Added 2 new macros, arrappn and arrappda

gimpycpu:
Added 2 new macros

 arrappn:
    void arrappn(T* a, T* v, int n);
      Appends n elements of v at the end of a, capacity might increase due to 
      arrsetcap being called

  arrappda:
    void arrappda(T* a, T* v);
      Appends dynamic array v at the end of a, capacity might increase due to 
      arrsetcap being called

unless I missed something I dont think there were any? 
I used them with success while pushing vertices, while loading a mesh data for example.
```
vec3* da = NULL;
//Iterate over all mesh
for (unsigned int i = 0; i < scene->mNumMeshes; i++)
{
    aiMesh* mesh = scene->mMeshes[i]; 
    stbds_arrappn(da, mesh->mVertices, mesh->mNumVertices); //append all vertices of the mesh to the dynamic array
}
```

### DIFF
--- a/stb_ds.h
+++ b/stb_ds.h
@@ -146,6 +146,16 @@ DOCUMENTATION
           Returns the number of total elements the array can contain without
           needing to be reallocated.
 
+      arrappn:
+        void arrappn(T* a, T* v, int n);
+          Appends n elements of v at the end of a, capacity might increase due to 
+          arrsetcap being called
+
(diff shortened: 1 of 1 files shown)


gimpycpu:
Found an issue, fixing

PR #1156, opened 2021-07-04
stb_image_write: Store uncompressed if compression was worse

a-e-k:
If the data was uncompressible and this deflate implementation expanded by more than the overhead of simply storing it uncompressed, fall back to deflate's uncompressed storage mode.  This bounds the worst-case maximum deflated size at the original size plus an overhead of 6 fixed bytes with another 5 bytes per 32767 byte block.


rygorous:
Merged into dev branch, will be in the next release, thanks for the fix!

Issue #734, opened 2019-03-11
stb_textedit: Undo weirdness after failed paste


Unit2Ed:
We've noticed some weirdness caused by the undo_point being erroneously adjusted after a failed paste.

We believe this is happening because `stb_textedit_paste_internal` decrements `state->undostate.undo_point` if the paste fails, but the undo hadn't been created yet (contrary to the comment above the decrement). It looks like the undo it's trying to remove is the one created by `stb_text_makeundo_insert`, which won't have happened if the insertion failed.

This is the current version of `stb_textedit_paste_internal`:
```
// API paste: replace existing selection with passed-in text
static int stb_textedit_paste_internal(STB_TEXTEDIT_STRING *str, STB_TexteditState *state, STB_TEXTEDIT_CHARTYPE *text, int len)
{
   // if there's a selection, the paste should delete it
   stb_textedit_clamp(str, state);
   stb_textedit_delete_selection(str,state);
   // try to insert the characters
   if (STB_TEXTEDIT_INSERTCHARS(str, state->cursor, text, len)) {
      stb_text_makeundo_insert(state, state->cursor, len);
      state->cursor += len;
      state->has_preferred_x = 0;
      return 1;
   }
   // remove the undo since we didn't actually insert the characters
   if (state->undostate.undo_point)
      --state->undostate.undo_point;
   return 0;
}
```

We've fixed this locally by removing the decrement, but I wanted to check your thoughts on this.

EDIT:
Forgot to mention that we've seen a soft-lock caused by this too in the case where the undo_char buffer has been filled up followed by a failed paste. `stb_text_create_undo_record` will lock-up in this while loop because `stb_textedit_discard_undo` does nothing if `undo_point==0`, which can happen due to the erroneous decrement.

```
   // if we don't have enough free characters in the buffer, we have to make room
   while (state->undo_char_point + numchars > STB_TEXTEDIT_UNDOCHARCOUNT)
      stb_textedit_discard_undo(state);
```

rygorous:
Fix for this is merged into dev branch, will be in the next release.

PR #1158, opened 2021-07-05
stb_textedit: Fix paste failure handling breaking undo stack (could lead to freezes)

ocornut:
This implement "Fix v1" described in #734 on the ground that it is simple.
(I can change it to "Fix v2" if deemed preferable)

Ended up spending a few hours researching and confirming this but in terms of crediting note that the actual fix was first suggested by Edward @Unit2Ed in #734.

----

**The failure conditions that are fixed are:**

(A)
- pasting does happen but WITHOUT a selection, stb_textedit_delete_selection() doesn't insert a undo point
- insertion fails
- code erroneously removed undo point

and

(B)
- pasting happens with a selection, but undo point cannot be created (deleted text larger than STB_TEXTEDIT_UNDOCHARCOUNT).
- insertion fails
- code erroneously removed undo point

----

**Repro**

https://github.com/ocornut/imgui/issues/4038 has a repro which should be easy to replicate outside of Dear ImGui (by using same inputs sequence and same buffer size). Dear ImGui doesn't change the default value of `STB_TEXTEDIT_UNDOCHARCOUNT`/`STB_TEXTEDIT_UNDOSTATECOUNT`.

### DIFF
--- a/stb_textedit.h
+++ b/stb_textedit.h
@@ -708,9 +708,7 @@ static int stb_textedit_paste_internal(STB_TEXTEDIT_STRING *str, STB_TexteditSta
       state->has_preferred_x = 0;
       return 1;
    }
-   // remove the undo since we didn't actually insert the characters
-   if (state->undostate.undo_point)
-      --state->undostate.undo_point;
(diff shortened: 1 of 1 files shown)


rygorous:
LGTM, merged into dev branch, will be in the next release.

PR #1161, opened 2021-07-09
stb_rect_pack: making functions cdecl for msvc qsort

ocornut:
(redoing #1160 without the branch fuckup)

Hello,

If you use stb_rect_pack without redefining `STBRP_SORT` it'll call `qsort()`, which expect the compare function to be using CDECL calling convention. If the default calling convention of the compile unit is not cdecl, compiler will choke on calling qsort `rect_height_compare()` and `rect_original_order()`.

I solved this on my side by adding a `STBRP__CDECL ` to those functions, based on the ground it won't hurt if those functions are always cdecl. 

Grepping `qsort` in stb/ folder the only other uses are: stb.h (ignoring) and stb_vorbis.c, which for the 2 functions used with `qsort()` does a similar thing:
```cpp
#ifdef _MSC_VER
#define STBV_CDECL __cdecl
#else
#define STBV_CDECL
#endif

static int STBV_CDECL uint32_compare(const void *p, const void *q)
{
   uint32 x = * (uint32 *) p;
   uint32 y = * (uint32 *) q;
   return x < y ? -1 : x > y;
}
```

### DIFF
--- a/stb_rect_pack.h
+++ b/stb_rect_pack.h
@@ -215,8 +215,10 @@ struct stbrp_context
 
 #ifdef _MSC_VER
 #define STBRP__NOTUSED(v)  (void)(v)
+#define STBRP__CDECL       __cdecl
 #else
 #define STBRP__NOTUSED(v)  (void)sizeof(v)
+#define STBRP__CDECL
 #endif
 
 enum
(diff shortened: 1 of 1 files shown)


rygorous:
Merged into dev branch, will be in the next release. (Ignore the CIFuzz failure, that's a problem with the CIFuzz integration, talking to Randy about how to debug that.)

nothings:
this includes a report about it being a potential issue in vorbis, has that been addressed?

ocornut:
Not an issue, was using this as an exemple that you already solved the same problem in the same way in stb_vorbis

nothings:
Ah, I read it reversed from that. All good!

PR #1165, opened 2021-07-11
Fix CIFuzz integration

randy408:
Follow-up from https://github.com/nothings/stb/pull/1058#issuecomment-877736188, by the way changes to the Dockerfile didn't break builds with `master` and everything should keep working after the merge, the solution wasn't that messy: https://github.com/google/oss-fuzz/pull/6023/files#diff-37336c5e77cb828028037c3ed7fa8f099fa00a0d25693f6b3f612e46bd119557R35


randy408:
Changes to the Dockerfile were merged a couple minutes ago, based on the logs CIFuzz is still using the older revision, might take a day for it to update.

PR #1169, opened 2021-07-15
Fix issue in stb_tilemap_editor.h which caused tiles in palette to be rendered incorrectly

SkaillZ:
The `STBTE__ID` was used instead of the slot, so an incorrect ID was passed to the `STBTE_DRAW_TILE` function.


nothings:
I don't think that fix is correct, you don't want to send back the internal slot number. It's only working because you're assigning IDs from 0..n-1 which matches the internal slot number. I think the correct value is t->id, but I haven't tested.

SkaillZ:
ah right thanks, I've updated it to use `t->id`

SushilRagoonath:
Thanks for this fix, saved me from doing it myself

Issue #937, opened 2020-03-26
stb_vorbis needs STB_VORBIS_MAX_PACKET_SIZE option with sane default


etc0de:
stb_vorbis really needs a STB_VORBIS_MAX_PACKET_SIZE option with sane default, like this:

```
#ifndef STB_VORBIS_MAX_PACKET_SIZE
#define (1024 * 1024 * 10)
#endif
```

Because while an ideal world according to specs, any packet size goes, in practice this means stb_vorbis currently will:

1. happily try to allocate up to 2GiB of memory even for tiny ogg files if they're faulty or malicious,
2. crash and likely corrupt its own memory (=could allow code execution) if a broken or malicious ogg file specifies a packet size between 2GiB and 4GiB, since the following function casts uint32 to sint32 with no overflow check:

   ```
   static int get32_packet(vorb *f)
   {
      uint32 x;
      x = get8_packet(f);
      x += get8_packet(f) << 8;
      x += get8_packet(f) << 16;
      x += (uint32) get8_packet(f) << 24;
      return x;
   }
   ```
The specs also recommend any packets to remain in 2-4KiB. While a max-size of considerably more might still be necessary for metadata like images (which is why I suggested a 10MiB default) it will otherwise likely never be used in any valid file. And a user who wants to be sure even files with giant ogg packet sizes will still work no matter the memory consumption, can then still use `#define STB_VORBIS_MAX_PACKET_SIZE (1024 * 1024 * 1024)`, or similar.

I would also recommend that test files are added with faulty large packets to ensure the limits actually work fine without a complete crash.

This issue makes stb_vorbis hardly usable for possibly broken or malicious files. It can also amplify other bugs, e.g. I ran into this because it failed to parse a likely perfectly valid file's header, ended up in wrong data, picked up a faulty >2GiB packet size in the wrong spot, and then self-destructed bringing down my program. It would be highly preferable if that use case just returned a parse failure and allowed me to resume operation. (I can email this file to someone interested, due to copyright reasons I don't think I am allowed to attach it here. Anyway, that is kind of a separate bug and the main issue IMHO really is the failure to enforce any sane packet sizes.

rygorous:
[...]

PR #1173, opened 2021-07-16
stb_vorbis: add STB_VORBIS_NO_COMMENTS to allow disabling comments.

AliceLR:
Another patch from libxmp, this time by @ccawley2011 (based on a separate patch by @sezero): adds a `STB_VORBIS_NO_COMMENTS` define to allow turning off comments support. This probably doesn't reduce code size very much, but being able to disallow arbitrary-sized allocations when they aren't needed seems useful.


nothings:
I believe this would be better solved by addressing #937, which is more flexible and addresses the core issue.

Issue #1170, opened 2021-07-15
arrclear (stb_ds.h feature request)


cajomar:
`arrsetlen(a, 0)` causes a warning when compiling with `-Wextra` with GCC:   
`warning: comparison of unsigned expression in ‘< 0’ is always false [-Wtype-limits]`  
Adding a new macro:
```c
#define arrclear(a) ((a) ? stbds_header(a)->length = 0 : 0)
```
gives a way to clear an array without warnings.  
There was a pull request (#731) for this, but it's owner deleted it without comment.

nothings:
It would be better to fix arrsetlen to not warn.

cajomar:
Then an easy fix to that would be to add 1 to both sides:
```diff
- #define stbds_arrsetlen(a,n)   ((stbds_arrcap(a) < (size_t) (n) ? stbds_arrsetcap((a),(size_t)(n)),0 : 0), (a) ? stbds_header(a)->length = (size_t) (n) : 0)
+ #define stbds_arrsetlen(a,n)   ((stbds_arrcap(a) + 1 < (size_t) (n) + 1 ? stbds_arrsetcap((a),(size_t)(n)),0 : 0), (a) ? stbds_header(a)->length = (size_t) (n) : 0)
```

nothings:
maybe cast both sides to an unsigned type

cajomar:
Like this?
```diff
-#define stbds_arrsetlen(a,n)   ((stbds_arrcap(a) < (size_t) (n) ? stbds_arrsetcap((a),(size_t)(n)),0 : 0), (a) ? stbds_header(a)->length = (size_t) (n) : 0)
+#define stbds_arrsetlen(a,n)   (((ptrdiff_t) stbds_arrcap(a) < (ptrdiff_t) (n) ? stbds_arrsetcap((a),(size_t)(n)),0 : 0), (a) ? stbds_header(a)->length = (size_t) (n) : 0)
```

Manuzor:
Or maybe flip the comparison from `<` to `>=`?

```diff
-#define stbds_arrsetlen(a,n)   ((stbds_arrcap(a) < (size_t) (n) ? stbds_arrsetcap((a),(size_t)(n)),0 : 0), (a) ? stbds_header(a)->length = (size_t) (n) : 0)
[...]

PR #1175, opened 2021-07-16
Add arrclear(), cast values to ptrdiff_t in arrsetlen() comparision

cajomar:
Pull request for issue #1170.

### DIFF
--- a/stb_ds.h
+++ b/stb_ds.h
@@ -131,6 +131,10 @@ DOCUMENTATION
           Deletes the element at a[p], replacing it with the element from
           the end of the array. O(1) performance.
 
+      arrclear:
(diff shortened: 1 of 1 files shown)


mohad12211:
can something similar to this be used for the hashmap?
setlen to zero but keep the allocated memory (capacity) for further use?

DontStopThinking:
This has been open for a while, any update on getting it merged please? I think being able to clear an array without freeing the allocated memory is a common use case, so it'd be useful to have this in stb_ds. Especially, since `std::vector` also has it with its `clear()` method.

Moreover, would it be worth adding some tests inside `stbds_unit_tests()` for this new macro?

PR #1176, opened 2021-07-18
stb_ds.h: update docs and changes overwritten by merge commit

laurelkeys:
- Added missing documentation for `arrlast` and `arrpush`
- Fixed two changes overwritten by the merge commit https://github.com/nothings/stb/commit/a32aeefd6bc270231b0000573699dabc05321c5a :
    - Readded call to custom free function in `stbds_arrfree`, from https://github.com/nothings/stb/pull/1144
    - Removed `stbds_arraddnoff`, which was rename to `stbds_arraddnindex` in https://github.com/nothings/stb/pull/1107 (since it isn't referenced anywhere else in the code and there's no short name `arraddnoff`, it doesn't seem necessary to keep it as a "synonym")

### DIFF
--- a/stb_ds.h
+++ b/stb_ds.h
@@ -95,7 +95,9 @@ DOCUMENTATION
           Removes the final element of the array and returns it.
 
       arrput:
+      arrpush:
         T arrput(T* a, T b);
+        T arrpush(T* a, T b);
           Appends the item b to the end of array a. Returns b.
 
       arrins:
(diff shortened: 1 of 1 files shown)


nothings:
I intentionally re-added `stbds_arraddnoff` just in case anybody was using it; it's possible people aren't using the short names, in which case they could have been using it.

`arrpush` is intentionally undocumented, as it as a compatibility definition for people used to stb_arr_push from stb.h, but `arrput` is the official name.

laurelkeys:
Oh, got it (my bad)! I removed `arrpush` from the documentation and re-added `stbds_arraddnoff`.

PR #1177, opened 2021-07-18
 Optional PNG text chunks support for stb_image

KivApple:
I am using stb_image for my small game engine project and thought that it would be great to embed a metadata (for example, char map for a bitmap font) in PNG files using some standard-conforming way. PNG specification (https://www.w3.org/TR/PNG-Chunks.html) allows compressed and uncompressed text chunks ("zTXt" and "tEXt" respectively) with keys (either standard like Title, Author etc or user defined) and some software and libraries (e. g. lodepng, libpng) support it. I've added this functional to stb_image, because it can be added just only by few lines of code.

Now you can set PNG text chunks handler (by default the option is completely disabled and shouldn't affect performance of the parser at all) and it would be called on each text chunk (both compressed and uncompressed chunks are supported). You can set some universal handler (also there is user_data pointer to be able to use the same handler multiple times with different behaivor or data structures) or change it before each decoding (and ever disable it again by passing NULL instead of the handler).

As I know there are similar features in some other formats (GIF comment extensions, JPEG APPn markers), so maybe it would be better to rename added functions in some format-agnostic way so it would be possible to add suport for other formats in future without introducing new APIs.

Thank you for your time!

P. S.: I am not sure how optimal I've implemented compressed text chunks handling - there are three allocations during the handling, but I am not enough familiar with stb_image source code :-(

### DIFF
--- a/stb_image.h
+++ b/stb_image.h
@@ -91,7 +91,7 @@ RECENT REVISION HISTORY:
  Optimizations & bugfixes                  Mikhail Morozov (1-bit BMP)
     Fabian "ryg" Giesen                    Anael Seghezzi (is-16-bit query)
     Arseny Kapoulkine                      Simon Breuss (16-bit PNM)
-    John-Mark Allen
+    John-Mark Allen                        Ivan Kolesnikov (PNG text chunks)
     Carmelo J Fdez-Aguera
 
  Bug & warning fixes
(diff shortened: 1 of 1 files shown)


nothings:
We want to keep stb_image as simple as possible for multiple reasons, so features like this are outside the scope of what we want to support.

KivApple:
Maybe it is at least possible to give users generic ability to set extra chunk handlers? It would give ability to implement any custom metadata handling (including text chunks), but it won't uncontrollable increase the complexity of the library in future (not something like "if there are already text chunks handling - let's implement one more special chunk type handling").

rygorous:
Sorry, but no. stb_image is for when all you want is the pixel data stored in a file. Everything else is out of scope.

We have been down this road before, more than once. Any such feature we add to the public API comes with an unspoken promise that it actually works robustly, which, no offense, it never does on the first try, and a subsequent flurry of bug reports, pull requests and security issues, as well as requests to add similar support for some of the other image formats. Even if we just add a "unknown chunk" callback for PNGs, we would be getting a steady stream of questions about how to use that API, as well as requests of assistance to debug incorrect uses, as a result.

The amount of spare time Sean and I have to spend on the stb libraries is very limited and we have learned the hard way just how much time even the tiniest-seeming feature additions end up costing us in the long run.

PR #1183, opened 2021-08-01
Add a getter for stbi__vertically_flip_on_load

laurelkeys:
Added a getter for the flag `stbi__vertically_flip_on_load`, so that we can restore it to its previous value after calling `stbi_set_flip_vertically_on_load()`, e.g.:
```c
int restore_flag = stbi_get_flip_vertically_on_load();

int x,y,n;
stbi_set_flip_vertically_on_load(true);
unsigned char *data = stbi_load(filename, &x, &y, &n, 0);

stbi_set_flip_vertically_on_load(restore_flag);
```

### DIFF
--- a/stb_image.h
+++ b/stb_image.h
@@ -521,6 +521,10 @@ STBIDEF void stbi_set_unpremultiply_on_load_thread(int flag_true_if_should_unpre
 STBIDEF void stbi_convert_iphone_png_to_rgb_thread(int flag_true_if_should_convert);
 STBIDEF void stbi_set_flip_vertically_on_load_thread(int flag_true_if_should_flip);
 
+// returns a non-zero value if images will be flipped vertically on load,
+// this can be changed with the stbi_set_flip_vertically_on_load function
+STBIDEF int stbi_get_flip_vertically_on_load(void);
+
 // ZLIB client - used by PNG, available for other purposes
 
 STBIDEF char *stbi_zlib_decode_malloc_guesssize(const char *buffer, int len, int initial_size, int *outlen);
@@ -1113,6 +1117,11 @@ STBIDEF void stbi_set_flip_vertically_on_load_thread(int flag_true_if_should_fli
                                          ? stbi__vertically_flip_on_load_local  \
                                          : stbi__vertically_flip_on_load_global)
 #endif // STBI_THREAD_LOCAL
+
+STBIDEF int stbi_get_flip_vertically_on_load(void)
+{
+   return stbi__vertically_flip_on_load;
+}
 
 static void *stbi__load_main(stbi__context *s, int *x, int *y, int *comp, int req_comp, stbi__result_info *ri, int bpc)
 {


nothings:
From reading it, I don't think this interacts correctly with the thread-local support.

laurelkeys:
I'm not really familiar with the use of thread-local variables, so I am sorry if this might not be a good example (code at the end).
Here I created a 1x2 image and launched 10 different threads that: 
1. save the current value of `stbi__vertically_flip_on_load`
2. set it to a new value based on whether the thread number is even or odd, then load the test image
3. check if the pixels are correctly flipped (or not), and that `stbi_get_flip_vertically_on_load()` returns the expected value
4. restores `stbi__vertically_flip_on_load`, and checks its value once again with `stbi_get_flip_vertically_on_load()`

Please let me know if there's some test case you have in mind.

---

Testing it on MSVC, GCC (MingW) and GCC on WSL2, I have gotten the results I expected (i.e. they all match):

image
image

---

```cpp
#include <cstdio>
#include <thread>

#include "stb/stb_image.h"

// @Note: test_image.png is an image with a (top) red pixel stacked on a
// (bottom) mid-gray pixel. Their values are: 0xff0000 and 0x808080.

#define RGB(pRGB) (((pRGB)[0] << 16) | ((pRGB)[1] << 8) | ((pRGB)[2] << 0))
#define TOP 0xff0000
#define BOTTOM 0x808080

void test_thread_local(int i, int set, int top, int bottom)
{
    int old = stbi_get_flip_vertically_on_load();
    stbi_set_flip_vertically_on_load_thread(set);

    int x, y, n;
    unsigned char *data = stbi_load("test_image.png", &x, &y, &n, 0);
    assert(x == 1 && y == 2 && n == 3);

    int get = stbi_get_flip_vertically_on_load();
    std::printf(" %d %s %d  |  %x   %x  |  %x   %x  (thread: %d)\n",
        set, ((set == get) ? "==" : "!="), get,
        top, RGB(data), bottom, RGB(data + 3), i);

    stbi_set_flip_vertically_on_load_thread(old);
    assert(stbi_get_flip_vertically_on_load() == old);
}

int main()
{
    std::printf("set  get |   top      top    |  bottom   bottom\n");

    std::thread threads[10]; // @Note: flip vertically on even threads.
    for (int i = 0; i < 10; ++i) {
        if ((i % 2) == 0) {
            threads[i] = std::thread(test_thread_local, i, /*set=*/1, BOTTOM, TOP);
        } else {
            threads[i] = std::thread(test_thread_local, i, /*set=*/0, TOP, BOTTOM);
        }
    }
    for (int i = 0; i < 10; ++i) threads[i].join();

    std::printf("           (actual) (expect) | (actual) (expect)\n");
}
```

nothings:
>     1. save the current value of `stbi__vertically_flip_on_load`
>     2. **set it to a new value based on whether the thread number is even or odd, then load the test image**
>     3. check if the pixels are correctly flipped (or not), and that `stbi_get_flip_vertically_on_load()` returns the expected value
>     4. restores `stbi__vertically_flip_on_load`, and checks its value once again with `stbi_get_flip_vertically_on_load()`

what function did you call in step 2?

laurelkeys:
> what function did you call in step 2?

I used `stbi_set_flip_vertically_on_load_thread()`, same for step 4.

I believe it should be ok to have a single "getter" implementation, since the macro `stbi__vertically_flip_on_load` is the value that is always being checked before calls to `stbi__vertical_flip()`.

nothings:
I agree it correctly produces the value that will get _used_, but it's not possible to correctly restore the state since you can't tell whether you're getting the global or the thread-local value if you're using a mix of set calls. It only works right if you only use the '_.thread()` setter.

Now, I don't think it makes much sense to use a mix of the set calls, but I'm not sure that a getter that exposes that underlying behavior in an awkward, unrecoverable way is really a good thing to commit to for all time. I would argue that it will be more effective to not use get.

First, an example. It is possible that at some future time we will provide a new API that takes additional parameters, and those parameters will include flags like `vertically_flip` whose value would be used directly, ignoring the globals. Most people would accept that this is a _better_ API, as it avoids complexities with e.g. thread local variables. In that situation, obviously `get` would have no utility since any value it could get would be ignored. There's no need to restore a hidden variable to its old value if the correct value is always supplied on every call.

However, you can _right now_ write code that behaves in the same way, by simply always calling `stbi_set_flip_vertically_on_load_thread()` explicitly before every call to `stbi_load()`. This will behave exactly the same as that future API. So if you do this, there's no need for a `get` function, by the same argument.

Due to the complexities of the thread local variables (which were written that way to preserve back-compatibility), a straightforward `get` implementation is possible but will not actually work correctly in all cases for saving/restoring the hidden variables, and there is (as far as I know) no other use for such a getter. Rather than try to engineer something that handles that complexity, we can dictate by fiat that those cases aren't important. But if can dictate that, we can just dictate "always set `vertically_flip` on every call if you ever set it".

laurelkeys:
> (...) It is possible that at some future time we will provide a new API that takes additional parameters, and those parameters will include flags like `vertically_flip` whose value would be used directly, ignoring the globals. Most people would accept that this is a _better_ API (...)
> (...) you can _right now_ write code that behaves in the same way, by simply always calling `stbi_set_flip_vertically_on_load_thread()` explicitly before every call to `stbi_load()` (...)

Totally agree! I'm already calling `stbi_set_flip_vertically_on_load()` before every call to `stbi_load()`, but this can "pollute the global state" when interacting with other code that doesn't do the same (e.g. the `vertically_flip` flags are false by default, so if I set it to true, some library function that also uses stb_image will behave incorrectly if it expects it to be false).

However, considering the need for backwards compatibility, I think it's best not to add this function to the API (as a function like you suggested above would be the optimal solution). Hence, I am closing the PR.

PR #1199, opened 2021-08-27
variable bt may be uninitialized: stbsp__int64 bt

mgood7123:


moon-chilled:
The warning is bogus.  It is misleading to initialise twice.

mgood7123:
> The warning is bogus. It is misleading to initialise twice.

oh :/

can this warning be silenced in some way?

PR #1204, opened 2021-09-08
Remove stb_perlin from tests

musicinmybrain:
Since `stb_perlin.h` was removed in https://github.com/nothings/stb/commit/59e7dec3e8bb0a8d4050d03c2dc32cf71ffa87c6, this PR removes it from the tests.


musicinmybrain:
Now that `stb_perlin` is back, this PR is no longer useful.

Issue #992, opened 2020-07-12
stb_ds crashes when inserting struct


srdjanstipic:
The following program crashes:

```c
#define STB_DS_IMPLEMENTATION
#include "stb_ds.h"

int main() {
  struct { char *key; int value;} *t = NULL, s = {"a", 1};
  sh_new_strdup(t);
  shputs(t, s);
  shfree(t);
  return 0;
}
```

nothings:
This is a known limitation that is missing from the documentation.  The "struct" form of the string hash APIs can't be used with the arena/strdup modes.

(The issue is that internally the hash table attempts to strdup() the key field, but then the final update operation for struct operations is to overwrite the entire struct with the passed-in structure.)

I'll look into having it assert if it's used incorrectly.

makuto:
This issue should be closed after PR #993, right? Or was there a reason it is kept open?

srdjanstipic:
Should be closed.

PR #1207, opened 2021-09-15
Fix temp_key being stale on shput key re-insert

makuto:
63e4ada98c0e7feff1eb00d55a4399599ae13d7b copied the found key to temp_key so it would be re-used when a matching entry is found. However, I failed to add this to the beginning of bucket to pos search, so it only failed in a relatively rare circumstance where that branch was being used.

This fixes problems where re-inserting an entry (e.g. to update its data) will create a new entry with an identical key.

See previous issue #992 and pull request #993.


makuto:
I was unsure how to re-use my existing branch after merging changes from master. If I need to create a new clone off master to remove those extra commits, I can do that.

PR #1208, opened 2021-09-16
Avoid sign-compare warning in stb_sprintf with GCC 7.5.0

Robyt3:
This avoids the warning/error "signed and unsigned type in conditional expression [-Werror=sign-compare]" with GCC 7.5.0 on Ubuntu 18.04.


gonzus:
Please count this as an endorsement for this PR. Thanks!

PR #1211, opened 2021-09-21
stb_truetype: added stbtt_ScaleForPixelWidth

mossx-dev:
Added a function to calculate the pixel width of a letter. This is done by getting the start of the line gap and subtracting it by the end. This is done exactly like stbtt_ScaleForPixelHeight with different offsets.


rygorous:
Code looks good, doc comments misleadingly mention height; I'll merge this with fixes when I next do a stb_truetype pass.

PR #1217, opened 2021-09-28
High resolution image saving bug fix

GercogKaban:
Some variables types are changed to int64_t, because in some cases, for example, when saving an image in 32768x32768 resolution, int variable overflows (it is obvious to see that image size = 32768 * 32768 * 4 = 2^32 bigger than 2^32/2 (unsigned integer part).

Unfortunately, I was unable to find solution to fix png format saving.

### DIFF
--- a/stb_image_write.h
+++ b/stb_image_write.h
@@ -141,7 +141,6 @@ CREDITS:
       github:ignotion
       Adam Schackart
       Andrew Kensler
-      github:GercogKaban
 
 LICENSE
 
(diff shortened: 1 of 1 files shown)


nothings:
I'm not sure this is a good idea, as the stb_image reader has a similar 32-bit issue and cannot handle very large images. It would be a bit problematic if stb_image_write could write images that stb_image couldn't read.

GercogKaban:
Oh, i didn't think about it. I'm trying to implement megatexture in my graphics engine and it doesn't require loading big images into the memory.

Then, at least, i propose to handle this issue in code (raise exception in C++) and point on it in "documentation", because then we get an access violation error.

Btw, thank you for quick response.

rygorous:
Closing for now for reasons discussed in the thread.

PR #1223, opened 2021-10-07
Fixes a crash and an infinite loop in stb_image that could occur with specially constructed PGM and HDR files

NeilBickford-NV:
Fixes two availability issues in stb_image that could occur with specially
constructed HDR and PGM files. Please see issues #1224 and #1225 for full descriptions.

HDR: In certain conditions, the RLE decoder can get stuck in the decoding loop
at the end of a stream: `stbi__get8()` always returns 0 when at the end of a
stream, which is interpreted as a count and results in an infinite loop.
I believe the solution is to treat a run of 0 as invalid, following the
`RGBE_ReadPixels_RLE()` function in Bruce Walter's
[https://www.graphics.cornell.edu/~bjw/rgbe/rgbe.c](https://www.graphics.cornell.edu/~bjw/rgbe/rgbe.c).

PGM: Loading large monochrome 16-bit PGM files would cause an access violation
due to reading out of bounds when comverting from 16-bit to 8-bit. In addition,
when loading 16-bit PGM files, `stbi__pnm_load()` would call
`stbi__convert_format()` instead of `stbi__convert_format16()` to convert from
monochrome to the required number of channels, so the buffer would be
interpreted as the wrong type. This pull request also adds an error nessage for
when the `stbi__getn()` call fails.

Thanks!

### DIFF
--- a/stb_image.h
+++ b/stb_image.h
@@ -108,7 +108,7 @@ RECENT REVISION HISTORY:
     Cass Everitt            Ryamond Barbiero                        github:grim210
     Paul Du Bois            Engin Manap        Aldo Culquicondor    github:sammyhw
     Philipp Wiesemann       Dale Weiler        Oriol Ferrer Mesia   github:phprus
-    Josh Tobin                                 Matthew Gregan       github:poppolopoppo
+    Josh Tobin              Neil Bickford      Matthew Gregan       github:poppolopoppo
     Julian Raschke          Gregory Mullen     Christian Floisand   github:darealshinji
     Baldur Karlsson         Kevin Schmidt      JR Smith             github:Michaelangel007
                             Brad Weinberger    Matvey Cherevko      github:mosra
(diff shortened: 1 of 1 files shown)


NeilBickford-NV:
Following up on this - is there anything I can change to help merge this in? Thanks!

nothings:
stb_image is just slow to update

NeilBickford-NV:
Quick note from the survey of the first 10 ossfuzz issues I did: I think this pull request might resolve ossfuzz issues https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=37628&q=proj%3Dstb&can=2 (this seems to be the infinite HDR reader error) and https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=38394&q=proj%3Dstb&can=2 (I think this is due to the PGM issue). I'm unable to reproduce ossfuzz's crashes with these fixes on Windows with Address Sanitizer enabled.

rygorous:

PR #1228, opened 2021-10-12
stb_ds: Fix arrfree when custom allocators are used.

travisdoor:
Originally STBDS_FREE macro was used directly in arrfree definition,
this leads to invalid replacement in case custom realloc and free
are specified and arrfree is not used in "implementation" unit.


travisdoor:
Maybe I used it wrong, my configuration looks like this:
```
common.c

#define STB_DS_IMPLEMENTATION
#define STBDS_REALLOC(context, ptr, size) bl_realloc(ptr, size)
#define STBDS_FREE(context, ptr) bl_free(ptr)
#include "stb_ds.h"
```

Now when I use `arrfree` in `common.c` file, everything is OK, the original `free` is replaced with `bl_free`, however, in other files only including `stb_ds.h` without any additional defines, I get default `free` instead of `bl_free`. So i.e. an array allocated from `common.h` and freed from a different file causes failure (`bl_realloc` vs default `free`).
My small change is just to replace `STBDS_FREE` in `arrfree` macro with a function call (having correct free in its implementation).

aschuhardt:
Can this be merged?

nothings:
yes, it will be, sorry, i don't update very often since the pandemic started

Issue #1224, opened 2021-10-07
In stb_image's HDR reader, loading a specially constructed invalid HDR file can result in an infinite loop within the RLE decoder


NeilBickford-NV:
**Summary**
stb_image's HDR loader in versions 1.33 to and including v2.27 parsed truncated end-of-file RLE scanlines as an infinite sequence of zero-length runs. An attacker could potentially have caused denial of service in applications using stb_image by submitting specially constructed HDR files.

CVE number: [CVE-2021-42715](https://nvd.nist.gov/vuln/detail/CVE-2021-42715)

**Describe the bug**
In stb_image's HDR reader, loading a specially constructed invalid HDR file can
result in an infinite loop within the RLE decoder.

This issue includes a fix in pull request #1223, and a proof of concept file that
can be used to reproduce the crash. We're reporting this on GitHub Issues
following the guidance in issue #1213.

The issue occurs in [this loop](https://github.com/nothings/stb/blob/master/stb_image.h#L7184) within `stbi__hdr_load()`:

```
while ((nleft = width - i) > 0) {
   count = stbi__get8(s);
   if (count > 128) {
      // Run
      value = stbi__get8(s);
      count -= 128;
      if (count > nleft) { STBI_FREE(hdr_data); STBI_FREE(scanline); return stbi__errpf("corrupt", "bad RLE data in HDR"); }
      for (z = 0; z < count; ++z)
         scanline[i++ * 4 + k] = value;
   } else {
      // Dump
      if (count > nleft) { STBI_FREE(hdr_data); STBI_FREE(scanline); return stbi__errpf("corrupt", "bad RLE data in HDR"); }
      for (z = 0; z < count; ++z)
         scanline[i++ * 4 + k] = stbi__get8(s);
   }
}
```

The proof of concept file manages to get this part of the decoder into a state
where:

* `nleft` is equal to 11;
* `s` is at the end of the file.

Because `s` is at the end of the file, `stbi__get8()` always returns 0, since
`s->img_buffer == s->img_buffer_end` here:

```
stbi_inline static stbi_uc stbi__get8(stbi__context *s)
{
   if (s->img_buffer < s->img_buffer_end)
      return *s->img_buffer++;
   if (s->read_from_callbacks) {
      stbi__refill_buffer(s);
      return *s->img_buffer++;
   }
   return 0;
}
```

This means that `count` is always set to 0; this passes the error check, but
doesn't affect any program state, meaning that the loop runs forever, an
availability issue.

[...]

PR #1230, opened 2021-10-15
stb_image PNG reader: Adds checks for invalid DEFLATE codes, fixing an infinite loop found by ossfuzz.

NeilBickford-NV:
This pull request attempts to fix an issue where a specially formatted PNG file
could cause an infinite loop in `stbi__parse_huffman_block` (ossfuzz issue [24232](https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=24232&q=proj%3Dstb))

The file in ossfuzz issue 24232 appears to cause an infinite loop by making
literal 0 correspond to length 0, and then truncating the file early (resulting
in reading a stream of 0s). More specifically, the file contains a Huffman
length table in which literal 0 decodes to length code 287. It turns out that
length codes 286 and 287 are in fact invalid codes - they're illegal symbols,
but they're there because they're required for constructing the canonical
Huffman tree! stb_image handles these codes by setting `len`, the number of
bytes to output, to 0. However, the result is that the loops on lines [4274](https://github.com/nothings/stb/blob/master/stb_image.h#L4274) and
4276 do not advance the state. By doing some necessary setup (so that the range
checks on [lines 4265-4267](https://github.com/nothings/stb/blob/master/stb_image.h#L4265-L4267) pass) and then truncating the file early (so that
`stbi__zreceive` always returns 0, like in issue #1224), reading the ossfuzz
file results in an infinite loop.

Since length codes 286 and 287 are illegal, this pull request adds a branch to
reject them (which fixes the issue with this file).

This also adds a check to reject invalid distance codes 30 and 31 for completeness
(I think the only effect of setting `dist` to 0 by using a specially formatted
distance table would be that uninitialized data would be left in the output).

Another approach would be to check that len != 0 after line [4261](https://github.com/nothings/stb/blob/master/stb_image.h#L4261).
However, I think it's possible for 0 to appear as a valid literal value, 256,
in the fixed Huffman length table (RFC 1951 section 3.2.6 - or maybe
256 also represents "end of block" there?)

(This does not adjust the credits to avoid merge conflicts with PR #1223)

Some references used when checking that length codes 286 and 287 and distance
codes 30 and 31 are indeed invalid:

RFC 1951 (DEFLATE Compressed Data Format Specification) sections 3.2.5-3.2.7: https://datatracker.ietf.org/doc/html/rfc1951#page-11

* The tables in 3.2.5 present instructions for only length codes 0-285, and distance codes 0-29.

* Section 3.2.6 states

```
         Literal/length values 286-287 will never actually
         occur in the compressed data, but participate in the code
         construction.

         Distance codes 0-31 are represented by (fixed-length) 5-bit
         codes, with possible additional bits as shown in the table
         shown in Paragraph 3.2.5, above.  Note that distance codes 30-
         31 will never actually occur in the compressed data.
```

https://en.wikipedia.org/wiki/Deflate#Bit_reduction lists

```
286, 287: not used, reserved and illegal but still part of the tree.
```

and

```
30–31: not used, reserved and illegal but still part of the tree.
```

Thanks!

### DIFF
--- a/stb_image.h
+++ b/stb_image.h
@@ -4256,11 +4256,12 @@ static int stbi__parse_huffman_block(stbi__zbuf *a)
             a->zout = zout;
             return 1;
          }
+         if (z >= 286) return stbi__err("bad huffman code","Corrupt PNG"); // per DEFLATE, length codes 286 and 287 must not appear in compressed data
          z -= 257;
          len = stbi__zlength_base[z];
          if (stbi__zlength_extra[z]) len += stbi__zreceive(a, stbi__zlength_extra[z]);
          z = stbi__zhuffman_decode(a, &a->z_distance);
-         if (z < 0) return stbi__err("bad huffman code","Corrupt PNG");
+         if (z < 0 || z >= 30) return stbi__err("bad huffman code","Corrupt PNG"); // per DEFLATE, distance codes 30 and 31 must not appear in compressed data
          dist = stbi__zdist_base[z];
          if (stbi__zdist_extra[z]) dist += stbi__zreceive(a, stbi__zdist_extra[z]);
          if (zout - a->zout_start < dist) return stbi__err("bad dist","Corrupt PNG");


NeilBickford-NV:
Also following up on this - is there anything I can do to help merge this in?

I'm hoping to have a go at fixing the remaining ossfuzz issues, but was holding off since I feel like queueing up too many merge requests at once could be difficult to handle - if queueing up the pull requests to fix those would be the way to go, though, please let me know.

Thanks again!

nothings:
feel free to queue up more, i tend to process PRs in batches

NeilBickford-NV:
Awesome - thank you, will do!

NeilBickford-NV:
Quick update from the survey of the first 10 ossfuzz issues I did: I think this pull request fixes ossfuzz issues https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=24232&q=proj%3Dstb&can=2 and https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=24235&q=proj%3Dstb&can=2 (the latter seems to be a Huffman table that produces a really large run of blocks). It might also fix https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=37896&q=proj%3Dstb&can=2 and https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=24186&q=proj%3Dstb&can=2 .

The last one (24186) is a bit odd, though: what seems to be happening is that in stbi__png_load, z->expanded isn't written past the first byte. That corresponds to a 1-byte filter, plus uninitialized raw data. This data is copied into the output; then when the `else` branch of `stbi__compute_transparency16` compares `tc` (which is initialized) against `p`, we get use of an uninitialized value. It's possible that the file the fuzzer produced just happens to have an invalid Huffman code, and maybe there's a deeper issue underlying issue 24186. Probably the best way to find out (if I had to guess) would be to let the fuzzer run on the code with the pull request merged in, to see if it finds a way of getting uninitialized data to `stbi__compute_transparency16` without bad Huffman codes.

Thanks!

NeilBickford-NV:
Quick heads-up: I've rebased the branches used in this PR, #1223, and #1297 over the latest `master` branch commit. The diffs should be the same as before if I've done that correctly. Thanks!

nothings:
Thanks, sorry I'm so slow!

rygorous:
(At long last) merged to dev branch, will be in the next release.

rygorous:
Fixed in 2.28.

PR #1241, opened 2021-10-31
Added another sample program and updated a sample OpenGL call.

Underdisc:
I posted this question #1238 and ended up solving my own problem. I wouldn't have encountered this issue had it been demonstrated in a snippet within the file. It wasn't much effort to add and I thought it would be a useful hint to those that fall into the same trap.

Like I said in the bottom of #1238, GL_ALPHA is not a valid option in recent versions of OpenGL. This also updates that to use working parameters.

### DIFF
--- a/stb_truetype.h
+++ b/stb_truetype.h
@@ -291,7 +291,7 @@ void my_stbtt_initfont(void)
    // can free ttf_buffer at this point
    glGenTextures(1, &ftex);
    glBindTexture(GL_TEXTURE_2D, ftex);
-   glTexImage2D(GL_TEXTURE_2D, 0, GL_ALPHA, 512,512, 0, GL_ALPHA, GL_UNSIGNED_BYTE, temp_bitmap);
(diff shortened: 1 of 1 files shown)


Underdisc:
Feel free to close if this addition isn't desirable. Figuring out my problem took a while but passionately making a pr afterwards did not. I can understand why it might not be wanted since sample code shouldn't lean heavily towards a specific graphics api.

nothings:
The problem is that GL_R8 doesn't work in OLDER versions of OpenGL. It's not possible to have something work with both. I'm more likely to just delete the examples entirely than try to solve that conundrum.

Underdisc:
True. I didn't even notice the glVertex stuff because I am using a single vertex buffer to render all of the glyphs. I just focused on dealing with all the scaling issues that come with that.

If having the examples be compliant with older versions is necessary, removing examples that call any OpenGL functions is good to get rid of the confusion. The non OpenGL examples are still nice to have, though, and aren't restricted by the same constraint. What's the reason for keeping the example code compliant with old versions of the OpenGL api?

nothings:
because (a) the examples are simple and don't need to set up vertex buffers etc, and (b) because Windows opengl headers are still 1.1 so is the eaiest version to target on that platform

PR #1243, opened 2021-11-08
stb_image: Added support for RLE compressed BMP

Voldakk:
I was using stb_image to convert a ton of BMPs from an old game to PNG and some failed due to them being RLE compressed, so I added support for 4 and 8-bit RLE compressed BMPs. 
While this isn't a new file format i guess it will still increase the attack surface of stb_image, though I think it should be added for completeness.


rygorous:
The validation in here is not correct (in the 4bpp RLE decoder, x is tracked incorrectly; each repetition produces 2 pixels). Testing x <= img->x isn't safe either, this needs to be aware of whether pixel count is odd or not. 

Don't take this the wrong way, but this is a feature very few people care about and if I find security-relevant issues with the patch within a minute of looking at it, then I'd rather not take it.

PR #1266, opened 2021-12-24
stb_vorbis: allow disabling the seek api

sezero:
Bound to the new STB_VORBIS_NO_SEEK_API define.


nothings:
what problem does this solve?

sezero:
It helps with users discard code which is dead to them

nothings:
It's not worth doing a feature-by-feature disabling, as it becomes too hard to maintain the code due to all the various combinations of flags.

sezero:
As you wish, feel free to close.

PR #1272, opened 2022-01-04
add missing shget*ts and don't reassign hashmap in hmget*ts

### LINKED ISSUE #1271: hmget*_ts macros not completely thread safe
**Describe the bug**
hmget*_ts macros are not thread safe since they reassign the hash map pointer passed to them

**To Reproduce**

Use stb_ds's hash map with thread sanitizer enabled, example code (pthread_cancel etc ommited):

```
#include <pthread.h>
#define STB_DS_IMPLEMENTATION
#include "stb_ds.h"

struct hm {
	int key;
	int value;
};

struct hm *h = NULL;

void *read(void *arg) {
	for (;;) {
		ptrdiff_t t = 0;
		hmget_ts(h, 1, t);
	}
}

int main() {
	hmput(h, 1, 2);
	pthread_create(&(pthread_t){0}, NULL, read, NULL);

	for (;;) {
		ptrdiff_t t = 0;
		hmget_ts(h, 1, t);
	}
}
```

TSAN output:

```
==================
WARNING: ThreadSanitizer: data race (pid=11042)
  Write of size 8 at 0x0000037c2200 by main thread:
    #0 main /home/testuser/git/ds.c:25:3 (a.out+0x4de64d)

  Previous read of size 8 at 0x0000037c2200 by thread T1:
    #0 read /home/testuser/git/ds.c:15:3 (a.out+0x4de3cc)

  Location is global 'h' of size 8 at 0x0000037c2200 (a.out+0x0000037c2200)

  Thread T1 (tid=11044, running) created by main thread at:
    #0 pthread_create <null> (a.out+0x48ee8a)
    #1 main /home/testuser/git/ds.c:21:2 (a.out+0x4de5f1)

SUMMARY: ThreadSanitizer: data race /home/testuser/git/ds.c:25:3 in main
==================
```

**Expected behavior**
hmget*_ts macros should not reassign the original hash map pointer. Reassigning is only useful when the hashmap is NULL https://github.com/nothings/stb/blob/af1a5bc352164740c1cc1354942b1c6b72eacb8a/stb_ds.h#L1284

Can't think of a good fix other than the caller making a copy of the original pointer and passing that to stbds, like `struct hm *tmp = hm; hmget_ts(tmp, 1, t);` and ensuring that `hm` is not NULL to avoid a leak

The other fix is to make stbds expect a non-null hashmap but that will require changing the api a bit

git-bruh:
closes #1271

### DIFF
--- a/stb_ds.h
+++ b/stb_ds.h
@@ -302,7 +302,10 @@ NOTES
 
   * The following are the only functions that are thread-safe on a single data
     structure, i.e. can be run in multiple threads simultaneously on the same
-    data structure
+    data structure. hmget*_ts and shget*_ts require non-NULL hashmaps passed
+    to them to avoid leaks as the internal wrapper function reallocates the
+    pointer if it is NULL. The macros don't reassign the hashmap in order to
+    remain thread-safe.
         hmlen        shlen
         hmlenu       shlenu
         hmget_ts     shget_ts
@@ -379,6 +382,7 @@ CREDITS
     Andreas Molzer
     github:hashitaku
     github:srdjanstipic
+    github:git-bruh
     Macoy Madson
     Andreas Vennstrom
     Tobias Mansfield-Williams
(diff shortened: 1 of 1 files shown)


nothings:
in addition to not reassigning the pointer in the wrapper macros, the implementation function of the get*_ts should be changed to not reallocate on NULL at all, just assign the empty index to the temp and return. the code that allocates it should be moved from hmget_key_ts implementation to hmget_key.

Behavior-wise, this means that as written, current hmget_ts will crash on NULL and hmgetp_ts will return an invalid but not-NULL pointer pointer, but hmgeti_ts will be well-behaved. In addition to this change, the wrapper for hmgetp_ts can probably be made to return NULL, but that's not consistent with how the other hmgetp's work; I wonder if this should be changed to return NULL instead of returning a pointer to the hidden 'default' slot in the hash table array.

nothings:
Other than the above, this looks good to me.

nothings:
looks good, thank you for your contribution

PR #1278, opened 2022-01-23
Remove sz cast to int in stb_leakcheck_malloc.

kennethjohnsen:
Casting sz to an int before storing it in mi->size is flagged as a warning with -Wconversion.
Very new to the STB libraries, so forgive me if I'm missing something.


nothings:
yep, probably a historical artifact

PR #1290, opened 2022-02-17
Added fallthrough macro to suppress C++17 fallthrough warnings on MSVC and other compilers

wmcnamara:
This PR adds a STBI_FALLTHROUGH macro to fix warnings on Visual Studio about switch fallthrough.

The STBI_FALLTHROUGH macro is written to be defined if C++17 is enabled, and if it is enabled it will define STBI_FALLTHROUGH as [[fallthrough]], which stops the warnings.

### DIFF
--- a/l
+++ b/l
@@ -1,0 +1,14865 @@
+[33mcommit c529ebed3e81ca50bed77392fbc4ad097033ee97[m[33m ([m[1;36mHEAD -> [m[1;32mmaster[m[33m, [m[1;31morigin/master[m[33m, [m[1;31morigin/HEAD[m[33m)[m
+Author: Weston McNamara <westonfmcnamara@gmail.com>
+Date:   Thu Feb 17 10:01:17 2022 -0500
+
+    added fallthrough macro to suppress warning
+
+[33mcommit af1a5bc352164740c1cc1354942b1c6b72eacb8a[m
+Author: Sean Barrett <sean2@nothings.org>
+Date:   Fri Sep 10 00:48:29 2021 -0700
+
+    Update bug_report.md
+
+[33mcommit 37e21f17b2269fdc49cb0de83064d3af5f2962f6[m
+Author: Sean Barrett <sean2@nothings.org>
+Date:   Fri Sep 10 00:47:35 2021 -0700
+
+    Update bug_report.md
+
+[33mcommit b9e1d86edc4e526e494855bcb46d02f3269d58aa[m
+Author: Sean Barrett <sean2@nothings.org>
+Date:   Fri Sep 10 00:43:51 2021 -0700
+
+    Create config.yml
+
+[33mcommit b1826c9894c048e18cce45a289ee1db3fde7f40b[m
+Author: Sean Barrett <sean2@nothings.org>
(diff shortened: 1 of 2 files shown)


Wohlstand:
Please keep comment lines, they are required to shut up the GCC 7+ and CLang warnings (even with older standards than C++17, I do have C++11 at my projects, and this warning is always pops up if don't add the "`// fallthrough`" or "`/* fallthrough */`" comment line)

wmcnamara:
Thank you for catching this! I had no idea. I've added the comment to the end of the macro!

Wohlstand:
Lol, not that I actually meant, you just now made that both `[[fallthrough]]` and comment will appear at C++17 and some special conditions. I meant, comments should be always, at all standards. So:

```c
//C++17 fallthrough macro
#if (__cplusplus >= 201703L) || (defined(_MSVC_LANG) && (_MSVC_LANG >= 201703L) && (_MSC_VER >= 1913)) 
#define STBI_FALLTHROUGH [[fallthrough]];
#else
#define STBI_FALLTHROUGH
#endif

....
switch(x)
{
case 1:
case 2: STBI_FALLTHROUGH /* fallthrough */
case 3:
}
```

Wohlstand:
P.S. I doubt comments will be assed by macros, will need to run the preprocessor and see what it actually generated on output...

wmcnamara:
Oops, yeah I wasnt really thinking about that when I wrote it...

I'll go ahead and change that

Wohlstand:
I did an exact test:

With GCC:
изображение

With CLang:
изображение

So, adding that comment at macro has absolutely no sense

Wohlstand:
Now looks way better :fox_face: :+1:

moon-chilled:
It seems redundant to have both the comment and the explicit declaration.  Why not this instead?

```c
#elif defined(__GNUC__)
# define STBI_FALLTHROUGH __attribute__((fallthrough))
```

wmcnamara:
Thats definitely a much nicer solution.

@Wohlstand Does this fix the issue aswell on your end?

Wohlstand:
> It seems redundant to have both the comment and the explicit declaration. Why not this instead?

Old compilers would get hurt (such as GCC 5 and below), gonna check some to make sure all clear.

Wohlstand:
On GCC 4.8 the next warning appears:
```
# gcc -c fallthrough.c -o fallthrough.o
fallthrough.c: В функции «main»:
fallthrough.c:11:9: предупреждение: пустая декларация [по умолчанию включена]
         __attribute__((fallthrough));
         ^
```

"Empty declaration"

moon-chilled:
I think it is reasonable to give worse diagnostics on older compiler versions.  But yeah there should be a check for minimum gcc version.

Wohlstand:
So, the best and most universal solution would be just keeping the `/*fallthrough*/` comment which doesn't break old compilers and is approvable by modern compilers to not have the warning here. Speaking about MSVC compilers, they were the most painful compared to others (exclusively everything older than MSVC2015). There is still be the annoying case against friend templates in classes and the painful inability to take the pointer to the member of the sub-structure (everything works, but MSVC just fails).

Issue #1293, opened 2022-02-17
UBSAN: shift exponent is too large


pietroborrello:
**Describe the bug**
Several `UBSAN runtime error: shift exponent 32 is too large for 32-bit type 'unsigned int'` and similar

**To Reproduce**
Built stb according to [the oss-fuzz script](https://github.com/nothings/stb/blob/master/tests/ossfuzz.sh) with `CXXFLAGS='-O1 -fsanitize=address -fsanitize=array-bounds,bool,builtin,enum,float-divide-by-zero,function,integer-divide-by-zero,null,object-size,return,returns-nonnull-attribute,shift,signed-integer-overflow,unreachable,vla-bound,vptr'`

**UBSAN Output**

```
$ ./stbi_read_fuzzer ./id:000116,sig:06,src:001260,time:12860161,op:havoc,rep:16,trial:1503866
INFO: Seed: 1313754043
INFO: Loaded 1 modules   (6883 inline 8-bit counters): 6883 [0x5e1b33, 0x5e3616), 
INFO: Loaded 1 PC tables (6883 PCs): 6883 [0x573228,0x58e058), 
stbi_read_fuzzer: Running 1 inputs 1 time(s) each.
Running: id:000116,sig:06,src:001260,time:12860161,op:havoc,rep:16,trial:1503866
src/stb/tests/../stb_image.h:2065:27: runtime error: shift exponent 32 is too large for 32-bit type 'unsigned int'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior src/stb/tests/../stb_image.h:2065:27 in 
Executed id:000116,sig:06,src:001260,time:12860161,op:havoc,rep:16,trial:1503866 in 2 ms
***
*** NOTE: fuzzing was not performed, you have only
***       executed the target code on a fixed set of inputs.
***

```

**Crashing files** 
[ubsan-shift-too-large.zip](https://github.com/nothings/stb/files/8090129/ubsan-shift-too-large.zip)

rygorous:
I tested the provided repros agains Neil's patches and the bugs are confirmed fixed. Patches are in dev branch, fix will be in the next release.

rygorous:
Fixed in 2.28.

Marietto2008:
nope,it is not fixed...

![Istantanea_2023-06-30_02-23-45](https://github.com/nothings/stb/assets/755339/01be6847-0327-4644-a7aa-783e955ba4cb)

[...]

Issue #1292, opened 2022-02-17
UBSAN: integer overflow


pietroborrello:
**Describe the bug**
`UBSAN: runtime error: signed integer overflow: -126340289 * 17 cannot be represented in type 'int'`
and
`UBSAN: runtime error: signed integer overflow: -2147450975 + -32767 cannot be represented in type 'int'`

**To Reproduce**
Built stb according to [the oss-fuzz script](https://github.com/nothings/stb/blob/master/tests/ossfuzz.sh) with `CXXFLAGS='-O1 -fsanitize=address -fsanitize=array-bounds,bool,builtin,enum,float-divide-by-zero,function,integer-divide-by-zero,null,object-size,return,returns-nonnull-attribute,shift,signed-integer-overflow,unreachable,vla-bound,vptr'`

**UBSAN Output**
```
$ ./stbi_read_fuzzer ./id:000130,sig:06,src:002266+002478,time:16238914,op:splice,rep:16,trial:1492432

INFO: Seed: 1429753284
INFO: Loaded 1 modules   (6883 inline 8-bit counters): 6883 [0x5e1b33, 0x5e3616), 
INFO: Loaded 1 PC tables (6883 PCs): 6883 [0x573228,0x58e058), 
../cve_exp/work_LIBFUZZER_HELPER_STB_STBI_READ_FUZZER/out/stbi_read_fuzzer: Running 1 inputs 1 time(s) each.
Running: id:000130,sig:06,src:002266+002478,time:16238914,op:splice,rep:16,trial:1492432
src/stb/tests/../stb_image.h:2251:29: runtime error: signed integer overflow: -1073741919 * 2 cannot be represented in type 'int'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior src/stb/tests/../stb_image.h:2251:29 in 
src/stb/tests/../stb_image.h:2249:35: runtime error: signed integer overflow: -2147450975 + -32767 cannot be represented in type 'int'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior src/stb/tests/../stb_image.h:2249:35 in 
Executed id:000130,sig:06,src:002266+002478,time:16238914,op:splice,rep:16,trial:1492432 in 76 ms
***
*** NOTE: fuzzing was not performed, you have only
***       executed the target code on a fixed set of inputs.
***
```

**Crashing files**
[...]

Issue #1291, opened 2022-02-17
UBSAN: index out of bounds


pietroborrello:
**Describe the bug**
Several UBSAN violations possibly related to #1289 
The zip attached contains the testcases that trigger the violation

**To Reproduce**
Built stb according to [the oss-fuzz script](https://github.com/nothings/stb/blob/master/tests/ossfuzz.sh) with `CXXFLAGS='-O1 -fsanitize=address -fsanitize=array-bounds,bool,builtin,enum,float-divide-by-zero,function,integer-divide-by-zero,null,object-size,return,returns-nonnull-attribute,shift,signed-integer-overflow,unreachable,vla-bound,vptr'`

**Sample output**
```
INFO: Seed: 1010447616
INFO: Loaded 1 modules   (6883 inline 8-bit counters): 6883 [0x5e1b33, 0x5e3616), 
INFO: Loaded 1 PC tables (6883 PCs): 6883 [0x573228,0x58e058), 
stbi_read_fuzzer: Running 1 inputs 1 time(s) each.
Running: id:000092,sig:06,src:002772+001250,time:5407822,op:splice,rep:8,trial:1492432
src/stb/tests/../stb_image.h:1990:10: runtime error: index 257 out of bounds for type 'stbi_uc [257]'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior src/stb/tests/../stb_image.h:1990:10 in 
src/stb/tests/../stb_image.h:1991:4: runtime error: index 750 out of bounds for type 'stbi_uc [257]'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior src/stb/tests/../stb_image.h:1991:4 in 
src/stb/tests/../stb_image.h:2001:13: runtime error: index 256 out of bounds for type 'stbi__uint16 [256]'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior src/stb/tests/../stb_image.h:2001:13 in 
src/stb/tests/../stb_image.h:2000:17: runtime error: index 257 out of bounds for type 'stbi_uc [257]'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior src/stb/tests/../stb_image.h:2000:17 in 
src/stb/tests/../stb_image.h:1999:11: runtime error: index 264 out of bounds for type 'stbi_uc [257]'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior src/stb/tests/../stb_image.h:1999:11 in 
[...]

Issue #1289, opened 2022-02-17
AddressSanitizer: heap-use-after-free in stbi__jpeg_huff_decode


pietroborrello:
**Describe the bug**
UndefinedBehaviorSanitizer: undefined-behavior: index out of bounds + AddressSanitizer: heap-use-after-free in stbi__jpeg_huff_decode.

**To Reproduce**
Built stb according to [the oss-fuzz script](https://github.com/nothings/stb/blob/master/tests/ossfuzz.sh) with `CXXFLAGS='-O1 -fsanitize=address -fsanitize=array-bounds,bool,builtin,enum,float-divide-by-zero,function,integer-divide-by-zero,null,object-size,return,returns-nonnull-attribute,shift,signed-integer-overflow,unreachable,vla-bound,vptr'`

**ASAN Output**
```
$ ./stbi_read_fuzzer ./id:000233,sig:06,src:005305,time:31715435,op:havoc,rep:4,trial:1493419.jpeg

INFO: Seed: 1219869484
INFO: Loaded 1 modules   (6883 inline 8-bit counters): 6883 [0x5e1b33, 0x5e3616), 
INFO: Loaded 1 PC tables (6883 PCs): 6883 [0x573228,0x58e058), 
./stbi_read_fuzzer: Running 1 inputs 1 time(s) each.
Running: id:000233,sig:06,src:005305,time:31715435,op:havoc,rep:4,trial:1493419
src/stb/tests/../stb_image.h:1990:10: runtime error: index 257 out of bounds for type 'stbi_uc [257]'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior src/stb/tests/../stb_image.h:1990:10 in 
src/stb/tests/../stb_image.h:1991:4: runtime error: index 654 out of bounds for type 'stbi_uc [257]'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior src/stb/tests/../stb_image.h:1991:4 in 
src/stb/tests/../stb_image.h:2001:13: runtime error: index 256 out of bounds for type 'stbi__uint16 [256]'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior src/stb/tests/../stb_image.h:2001:13 in 
src/stb/tests/../stb_image.h:2000:17: runtime error: index 257 out of bounds for type 'stbi_uc [257]'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior src/stb/tests/../stb_image.h:2000:17 in 
[...]

PR #1297, opened 2022-02-23
Additional stb_image fixes for bugs from ossfuzz and issues 1289, 1291, 1292, and 1293

NeilBickford-NV:
Hi stb maintainers! I went through the first ten ossfuzz issues (up to and including ossfuzz issue 38394) and put together fixes for them. About half were duplicates of previous issues, but three were new and are fixed in this PR.

I also went through the test files in issues #1289, #1291, #1292, and #1293 and fixed ASan+UBSan reports from loading those issues' repro cases which weren't already fixed by #1223 or #1230.

Here are the bugs, some quick analyses, and their fixes:

https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=22620&q=proj%3Dstb&can=2 :

This is a PNM file with specifies a width of 3333333333. Parsing this integer overflows a signed 32-bit integer.

The fix I chose isn't the most elegant: I check to see if value * 10 + (*c - 0) would overflow a signed int, and then propagate an error up two levels. This also makes loading a 0-width or 0-height PNM file more explicitly an error (previously, this would produce an error for other reasons). However, I can think of a few other good ways of fixing this (e.g. simplifying the condition by limiting the maximum size a bit further, or changing `value` to an `unsigned int` and letting later error-handing handle it.) Please let me know if you would prefer any of these other ways!

https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=32803&q=proj%3Dstb&can=2 :

The reproducer file here is a PNG file with a second IDAT chunk with a `c.length` of 0x68088c86. The IDAT handler increases `idata_limit` to fit this amount of memory, and then reallocates space for `0xcce48000` bytes, causing an ossfuzz out-of-memory error (since this is greater than ossfuzz's limit of 2560 MB). The `stbi__getn` call later fails.

I'm not totally sure about my fix for this one, which is to reject IDAT sections larger than 1 GiB (the ossfuzz OOM threshold is 2560 MB). Comparing against the compressed data size would be ideal - but we don't have the data size in a callback context, right?

https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=36193&q=proj%3Dstb&can=2 :

The reproducer here is a JPEG file that manages to set `j->code_bits` to -8 in `stbi__grow_buffer_unsafe`! This results in a left shift of 32 bits on a 32-bit type, which is undefined behavior. This happens by getting into a situation where code_bits is 1, but the combined length s to read is 9.

This PR checks for this situation and returns an error if it happens.

Issue #1289:

Will post analysis in issue. This PR avoids this crash by bounds-checking `c`.

Issue #1291, file id_000154,sig_06,src_002783+000969,time_39921237,op_splice,rep_4,trial_1492432:

Will post analysis in issue. This pull request adds checks to both the `h->size[k++]` loop in `stbi__build_huffman` and the "DHT - define huffman table" code to verify that writes are in-bounds.

Since creating this pull request, I also put together fixes for several other files in issues 1292 and 1293:

Issue #1292, files id_000118,sig_06,src_003303,time_27343468,op_havoc,rep_16,trial_1496823 and id_000130,sig_06,src_002266+002478,time_16238914,op_splice,rep_16,trial_1492432:

These files all produce signed integer overflows (which are undefined behavior) when decoding the DC coefficient of a JPEG block or when accumulating delta codes (i.e. in the lines `data[0] = (shirt) (dc * dequant[0]);`, `data[0] = (short) (dc * (1 << j->succ_low));`, and `dc = j->img_comp[b].dc_pred + diff;`). This MR adds two functions, `stbi__addints_valid` and `stbi__mul2shorts_valid`, that check for overflow of addition of two 32-bit signed ints and multiplication of two signed shorts, respectively; if there's overflow here, we return an error. These two functions are maybe a bit more complex than ideal; maybe wrapping is OK here (in which case, casting to a signed type would make the overflow defined)?

Issue #1293, files id_000115,sig_06,src_003085,time_21982593,op_havoc,rep_8,trial_1497271, id_000157,sig_06,src_003484,time_46825823,op_havoc,rep_16,trial_1492432, and id_000212,sig_06,src_005033,time_18133274,op_havoc,rep_4,trial_1499760:

These are all relatively similar to ossfuzz issue 36193 above: the decoder exhausts its code buffer, then calls `stbi__grow_buffer_unsafe` (which I think implements NEXTBIT in ITU-T81), which immediately reaches a fill followed by a marker, leaving code_bits at 0. Any of a number of functions try to read some number of code bits from stbi__jpeg and decrease stbi__jpeg::code_bits, making it negative (e.g. -8 or -9)! The next time control returns to `stbi__grow_buffer_unsafe` and a byte is read that's not 0xff, the `j->code_buffer |= b << (24 - j->code_bits);` line left shifts by more than 32 bits.

I sort of manually patched each of the places where `stbi__grow_buffer_unsafe(j);` can be called followed by decreasing `j->code_bits` by making sure that the required number of bits were actually read. If not enough bits were read, I have the code act as if all 0s were read or return an error. This could be better - perhaps the best solution would be to implement NEXTBIT more closely and handle error propagation here? - but seems to work.

Thanks!

### DIFF
--- a/stb_image.h
+++ b/stb_image.h
@@ -4008,6 +4008,7 @@ static void *stbi__jpeg_load(stbi__context *s, int *x, int *y, int *comp, int re
    unsigned char* result;
    stbi__jpeg* j = (stbi__jpeg*) stbi__malloc(sizeof(stbi__jpeg));
    if (!j) return stbi__errpuc("outofmem", "Out of memory");
+   memset(j, 0, sizeof(stbi__jpeg));
    STBI_NOTUSED(ri);
    j->s = s;
    stbi__setup_jpeg(j);
@@ -4021,6 +4022,7 @@ static int stbi__jpeg_test(stbi__context *s)
    int r;
    stbi__jpeg* j = (stbi__jpeg*)stbi__malloc(sizeof(stbi__jpeg));
    if (!j) return stbi__err("outofmem", "Out of memory");
+   memset(j, 0, sizeof(stbi__jpeg));
    j->s = s;
    stbi__setup_jpeg(j);
    r = stbi__decode_jpeg_header(j, STBI__SCAN_type);
@@ -4046,6 +4048,7 @@ static int stbi__jpeg_info(stbi__context *s, int *x, int *y, int *comp)
    int result;
    stbi__jpeg* j = (stbi__jpeg*) (stbi__malloc(sizeof(stbi__jpeg)));
    if (!j) return stbi__err("outofmem", "Out of memory");
+   memset(j, 0, sizeof(stbi__jpeg));
    j->s = s;
    result = stbi__jpeg_info_raw(j, x, y, comp);
    STBI_FREE(j);
--- a/stb_image.h
+++ b/stb_image.h
@@ -1063,6 +1063,23 @@ static void *stbi__malloc_mad4(int a, int b, int c, int d, int add)
 }
 #endif
 
+// returns 1 if the sum of two signed ints is valid (between -2^31 and 2^31-1 inclusive), 0 on overflow.
+static int stbi__addints_valid(int a, int b)
+{
+   if ((a >= 0) != (b >= 0)) return 1; // a and b have different signs, so no overflow
+   if (a < 0 && b < 0) return a >= INT_MIN - b; // same as a + b >= INT_MIN; INT_MIN - b cannot overflow since b < 0.
+   return a <= INT_MAX - b;
+}
+
+// returns 1 if the product of two signed shorts is valid, 0 on overflow.
+static int stbi__mul2shorts_valid(short a, short b)
+{
+   if (b == 0 || b == -1) return 1; // multiplication by 0 is always 0; check for -1 so SHRT_MIN/b doesn't overflow
+   if ((a >= 0) == (b >= 0)) return a <= SHRT_MAX/b; // product is positive, so similar to mul2sizes_valid
+   if (b < 0) return a <= SHRT_MIN / b; // same as a * b >= SHRT_MIN
+   return a >= SHRT_MIN / b;
+}
+
 // stbi__err - error
 // stbi__errpf - error returning pointer to float
 // stbi__errpuc - error returning pointer to unsigned char
@@ -2135,6 +2152,7 @@ stbi_inline static int stbi__extend_receive(stbi__jpeg *j, int n)
    unsigned int k;
    int sgn;
    if (j->code_bits < n) stbi__grow_buffer_unsafe(j);
+   if (j->code_bits < n) return 0; // ran out of bits from stream, return 0s intead of continuing
 
    sgn = j->code_buffer >> 31; // sign bit always in MSB; 0 if MSB clear (positive), 1 if MSB set (negative)
    k = stbi_lrot(j->code_buffer, n);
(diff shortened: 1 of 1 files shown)


NeilBickford-NV:
By the way, I've made another branch at https://github.com/NeilBickford-NV/stb/tree/neilbickford/all-fixes which contains the fixes from this pull request and the other pull requests to fix fuzzer-found issues (#1223 and #1230), in case that's easier to look at or to merge in when the time comes to process everything. Thanks again!

NeilBickford-NV:
I've added some additional checks to fix issues reported in bugs 1292 and 1293; looks like there could still be at least one out-of-memory issue from fuzz-testing with libFuzzer.

NeilBickford-NV:
Update: Looks like the out-of-memory issue was in fact correctly reaching the 2^30 byte PNG IDAT section limit implemented in this merge request; the fuzzer set `c.length` to exactly 2^30 - 1, resulting in an allocation of 2^32 bytes. This went just over libFuzzer's default limit of 2048 Mb of memory, although it didn't go much higher than that. I guess checking against the file size when we're not reading data from streams might be a good idea, though!

NeilBickford-NV:
One more change in this latest commit: Running libFuzzer on stbi_read_fuzzer locally, I was seeing occasional out-of-bounds errors from UndefinedBehaviorSanitizer when accessing `stbi__bmask` in the line

```
   STBI_ASSERT((((j->code_buffer) >> (32 - h->size[c])) & stbi__bmask[h->size[c]]) == h->code[c]);
```

Unlike the previous bugs, I wasn't able to reproduce this one when loading an individual file, since it looks like it's due to reading from an uninitialized value. Specifically, it looks like it's possible for control to reach `stbi__jpeg_huff_decode` without setting the values inside the `stbi__huffman` struct, which would normally be set inside `stbi__build_huffman` (called from `stbi__process_marker`). So, my suspicion is that this was due to how `stbi__jpeg_load` doesn't clear the `stbi__jpeg` struct it allocates; `malloc` probably returned a memory block that was written by an earlier file during fuzz testing, which just happens to have the right values where calling `stbi__jpeg_huff_decode` without building the Huffman tables first indexes out of bounds.

Calling `memset(j, 0, sizeof(stbi__jpeg));` after allocating the `stbi__jpeg` struct appears to fix these intermittent errors. It's possible I've added it in more places than strictly necessary: it might not be necessary to zero-initialize it in `stbi__jpeg_test` and `stbi__jpeg_info`, and it might only be necessary to zero-initialize `stbi__jpeg::huff_dc` and `stbi__jpeg::huff_ac`. I decided to err on the side of caution here, but please let me know if you would like any modifications!

nothings:
it's perfectly fine to zero-initialize excessively

NeilBickford-NV:
Thanks, that's good to hear! I can have the fuzzer search for other cases where it winds up reading from uninitialized values and add fixes for those, if that would be good?

NeilBickford-NV:
Also, looks like the ossfuzz continuous integration check ran into the PNM crash from #1225 again - should I see if I can replace these 3 pull requests with a single pull request that combines them all? I think ossfuzz may well run cleanly on that. Thanks again!

nothings:
>  should I see if I can replace these 3 pull requests with a single pull request that combines them all?

Whatever is easiest or best for you; I tend to process PRs in batches anyway, so it doesn't matter too much to me.

NeilBickford-NV:
OK! I don't have too much of a preference either way and the PRs don't conflict with each other, so I'll leave it as is.

rygorous:
Code LGTM, changes merged into dev branch, will be in the next release.

PR #1299, opened 2022-03-02
stb_rect_pack: make rect_height_compare a stable sort

jsimmons:
stb_rect_pack was producing different results on different platforms due to differing qsort implementations.

This change uses the existing `was_packed` field to make the sort stable.


nothings:
ah, good call

sschaller:
Thanks for the fix! Also came across this while debugging inconsistent results across platforms.

PR #1306, opened 2022-03-06
Set up CodeQL scans and fixed several numeric overflows

artem-smotrakov:
I'd like to suggest setting CodeQL scans for `stb`. [CodeQL](https://codeql.github.com/) is a static-analysis engine that can help with detecting security and other issues. It can be easily run in a GitHub [workflow](https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/setting-up-code-scanning-for-a-repository#setting-up-code-scanning-using-actions). The suggested config runs scans on PRs and the main branch. Findings are going to be posted as comments in pull requests.

CodeQL reported several findings, mostly multiplication overflows and one suspicious check for pointer overflow. I've tried to fix them but this definitely needs a review because I have not writtten C code for a long time :)

I also noticed that `make -C tests all` fails because the tests still use `stb_perlin.h`. I've updated the tests, otherwise CodeQL jobs would fail.


nothings:
the perlin noise patent expired, so stb_perlin is coming back, so that's why nothing else involving it was changed

artem-smotrakov:
> the perlin noise patent expired, so stb_perlin is coming back, so that's why nothing else involving it was changed

Then, I think, `test_perlin.c` can be just temporarily removed from`Makefile`.

artem-smotrakov:
I've restored `test_perlin.c` and commented out `stb_perlin.h`.

Issue #1099, opened 2021-02-14
gcc/clang missing-field-initializers warnings for stb_image_write


flriancu:
Hi @nothings ,

GCC and clang report several `missing-field-initializers` warnings for stb_image_write.
Would you be interested in a PR for this? It would basically replace
```
stbi__write_context s = { 0 };
```
with
```
stbi__write_context s;
memset(&s, 0, sizeof(stbi__write_context));
```

Thanks!

rygorous:
I don't get these warnings on Windows with Clang 12, not even if I specifically ask for -Wmissing-field-initializers. Can you be more precise about which compiler version and target you're getting these warnings on?

flriancu:
@rygorous That was clang 9 or 10. If you can't reproduce with clang 12 nor with mingw gcc latest, I think we can close this issue. If time allows (unlikely), I'll test some more in the following days. If I can't get around to it, feel free to update this ticket as you see fit.

DctrNoob:
I also get this warning when compiling with clang 12.x under mac OS.

solidpixel:
I get warnings when compiling using clang++-9 and clang++-12 on Ubuntu 20.04 (WSL).

nothings:
I can make this happen on godbolt for clang 12.0.1 when compiling as C++ (but not as C) if I use -Wmissing-field-initializers, but not with -Wall. https://godbolt.org/z/5j8s87Ta7

@DctrNoob and @solidpixel, please clarify C vs C++ and commandline -W arguments.

nothings:
My guess is that the logic here is the {0} initializer is a C idiom so clang won't warn in C, but C++ has so many other ways to initialize structs that it's no longer considered an acceptable idiom there. If that's the sort of logic behind this (meaning they're not going to fix it), I guess it might be worth changing. However, if it doesn't happen even with -Wall I don't know if it's really worth it.

solidpixel:
Can confirm it's C++ (clang++) for my builds. 

Team policy is to go strict on warnings, so we're compiling with `-Wall -Wextra`. Testing locally on clang++-9 it's the `-Wextra` which is tripping this up, `-Wall` alone doesn't trigger it.

Given the original code is perfectly fine in C and we're slightly off-piste compiling it as C++, I've just worked around this locally using C++ zero-initializers.

DctrNoob:
@nothings I can confirm @solidpixel 's finding. I only get the warning when using `-Wextra`.

nothings:
[...]

PR #1311, opened 2022-03-13
stb_image_write: silence C++ warnings for = {0}

N-R-K:
Closes: https://github.com/nothings/stb/issues/1099

---

Grep finds a couple more instances of `{ 0 }` being used:

```console
$ grep --recursive "{.0.}"
stb_vorbis.c:   { 0 },
stb_voxel_render.h:   stbvox_mesh_face face_data = { 0 };
stb_voxel_render.h:   stbvox_mesh_vertex p1[4] = { 0 };
stb_ds.h:  stbds_string_arena                        sa      = { 0 };
stb_herringbone_wang_tile.h:   stbhw_config c = { 0 };
stb_herringbone_wang_tile.h:   stbhw__process p = { 0 };
stb_connected_components.h:   unsigned char num_adj[STBCC__MAX_CLUMPS_PER_CLUSTER] = { 0 };
stb_connected_components.h:   unsigned char connected[STBCC__MAX_EDGE_CLUMPS_PER_CLUSTER][STBCC__MAX_EDGE_CLUMPS_PER_CLUSTER/8] = { { 0 } };
stb_connected_components.h:   unsigned char disconnected[STBCC__MAX_EDGE_CLUMPS_PER_CLUSTER][STBCC__MAX_EDGE_CLUMPS_PER_CLUSTER/8] = { { 0 } };
```

Fixing them should be the same as this PR, should I go ahead and update those too?


nothings:
We didn't really resolve this in #1099. I'm not a big fan of using memset() because it's ugly, but I'm not really a big fan of ZERO_INIT because it introduces the #ifdef. I mean, I understand it's harmless, it's just ugly in its own way.

If we did take this PR, it needs to be STBIW_ZERO_INIT and etc. for each library, so they don't interfere if the user has "ZERO_INIT" in their code with some other meaning. (And simply `#undef`ing at the end isn't sufficient, since they could #define it before including us.)

N-R-K:
> I'm not a big fan of using memset() because it's ugly, but I'm not really a big fan of ZERO_INIT because it introduces the #ifdef. I mean, I understand it's harmless, it's just ugly in its own way.

I agree actually.

> If we did take this PR, it needs to be STBIW_ZERO_INIT and etc. for each library, so they don't interfere if the user has "ZERO_INIT" in their code with some other meaning.

That's true. I'll wait until you make up your mind on it before updating the PR then.

NB: I'm fine if this is rejected, only opened it because you agreed to use `memset()` which I think is more ugly than the macro.

N-R-K:
Updated to using memset(), compiles fine on both gcc and clang without any warnings with `-Wall -Wextra -Wmissing-field-initializers`.

Sokolmish:
There are also several places where such initializer is used:

- stb_ds.h in line [764](https://github.com/nothings/stb/blob/af1a5bc352164740c1cc1354942b1c6b72eacb8a/stb_ds.h#L764) (also, I don't know for what purpose this variable is used here).
- stb_image.h in lines [5026](https://github.com/nothings/stb/blob/af1a5bc352164740c1cc1354942b1c6b72eacb8a/stb_image.h#L5026) and [5815](https://github.com/nothings/stb/blob/af1a5bc352164740c1cc1354942b1c6b72eacb8a/stb_image.h#L5815).

PR #1319, opened 2022-04-13
stb_image.h: add QOI support

jbendtsen:
This commit adds support for loading QOI images with stb_image.h. Not sure if this is on the cards right now, given that QOI isn't too time-consuming to implement for one's own projects, and that several libraries already exist that support it. If not, feel free to ignore this PR - at the very least it was a fun exercise!


nothings:
We are not accepting any new image formats for stb_image for security reasons.

Issue #1321, opened 2022-04-19
Make stdint.h optional, allow to provide own type definitions


mupfdev:
**Is your feature request related to a problem? Please describe.**
I'm actively developing a game for a legacy system. The compiler does not provide `stdint.h`, which seems to be a requirement for `stb_image.h`.

**Describe the solution you'd like**
It would be great if the library would offer the possibility to use own type definitions.

nothings:
stb_image already HAS a path that uses its own type definitions. It is currently conditioned on compiling with Microsoft Visual C. You need to identify a preprocessor symbol we can use to identify your legacy compiler/platform.

mupfelofen-de:
Hi!
I don't know if it makes sense to introduce a separate symbol for the preprocessor for each platform. This could certainly be done more generically. But if that is what is wanted, I would be happy to give it a try.

nothings:
It doesn't make sense to me to put it on a generic #define that the user can flip, because then every user of the library has to write their own code to set it appropriately, and if they are cross-platform they may have to write conditional definitions. There are a lot more users then there are platforms, so it makes more sense for us to do the work once than for every user to do it themselves.

mupfelofen-de:
Touché! I'll try to make it work on Symbian and send a PR ASAP.

rygorous:
Fixed in 2.28.

PR #1322, opened 2022-04-19
Provide type definitions for Symbian OS/EPOC, #1321

mupfdev:
The Symbian S60v1 SDK does not provide `stdint.h`.
This additional preprocessor definition ensures that the required types are provided so that `stb_image.h` also compiles on this legacy system.

This modification is intended for use with the [N-Gage SDK.](https://github.com/ngagesdk)-project.

No additional credit required.

### DIFF
--- a/stb_image.h
+++ b/stb_image.h
@@ -635,7 +635,7 @@ STBIDEF int   stbi_zlib_decode_noheader_buffer(char *obuffer, int olen, const ch
    #endif
 #endif
 
-#if defined (_MSC_VER) || defined __SYMBIAN32__
+#if defined (_MSC_VER) || defined (__SYMBIAN32__)
 typedef unsigned short stbi__uint16;
 typedef   signed short stbi__int16;
 typedef unsigned int   stbi__uint32;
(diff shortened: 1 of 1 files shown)


nothings:
Can you make both defined() tests have parentheses so it's consistent?

mupfelofen-de:
There are a lot of open PRs, so I have one question: what happens next?
I have included `stb` as a sub-module at the moment. I guess it's worth pointing to my own fork for now?

nothings:
Yes, stb fixes get merged in big clumps when I feel like it, which is not very often, so waiting for fixes to be officially released is not advised.

mupfdev:
In the meanwhile, 5 months have passed. Any chance of getting this fix merged any time soon? I am currently still using my local version, which is sub-optimal.

rygorous:
One batch of merges a year is about typical, though sometimes it's taken us multiple years.

Sean and I are professional SW devs and the stb libs are only one side project among several for both of us. 2-3 weekends a year how much time we're realistically spending on these libs. We'll get to it eventually; if that doesn't work for your purposes, feel free to stay on your own fork or switch to something else.

rygorous:
Not using this PR because it's a one-line fix and using the PR merging flow for that is more trouble than it's worth, but it will be in the next release, thank you! I'll close this PR but leave the original issue around.

PR #1331, opened 2022-05-23
Use 64-bit integer value for sample offsets

### LINKED ISSUE #1329: stb_vorbis: Replace plain int with int64_t at offset/length related values
**Is your feature request related to a problem? Please describe.**
Having the plain signed 32-bit integer causes problems at OGG files with 198 kHz sample rate and over than 3 hours length (examples: several concert footages with the 5+ hours length), because signed 32-bit integer allows addressing ~3 hours of audio with 198 kHz sample rate. I highly recommend replacing it with the 64-bit integer to allow huge files to be played normally. Official Xiph Vorbis uses the `int64_t` to address sample offsets.

**Describe the solution you'd like**
Offset related values should use `int64_t` to extend the maximum range of sample addressing. As this library is never gets compiled as a shared library, there should no ABI breakages with such changes.

Wohlstand:
Fixes #1329

This PR also includes code for #1295 (to don't re-do the same work another time. If merge this first, please CLOSE THE #1295)

### DIFF
--- a/stb_vorbis.c
+++ b/stb_vorbis.c
@@ -165,6 +165,12 @@ extern void stb_vorbis_close(stb_vorbis *f);
 // NOT WORKING YET after a seek with PULLDATA API
 extern int stb_vorbis_get_sample_offset(stb_vorbis *f);
 
+//  this function returns the count of returned samples from the beginning of the
+//  file. Functions "stb_vorbis_get_samples_*", "stb_vorbis_seek_*()" will
+//  affect the returned value. Use this call to get the accurate sample position
+//  during playback.
+extern int stb_vorbis_get_playback_sample_offset(stb_vorbis *f);
+
 // returns the current seek point within the file, or offset from the beginning
 // of the memory buffer. In pushdata mode it returns 0.
 extern unsigned int stb_vorbis_get_file_offset(stb_vorbis *f);
@@ -862,6 +868,9 @@ struct stb_vorbis
 
    uint32 current_loc; // sample location of next frame to decode
    int    current_loc_valid;
+
+   int32  current_playback_loc; // sample location of played samples
+   int    current_playback_loc_valid;
 
   // per-blocksize precomputed data
 
@@ -3512,6 +3521,8 @@ static int vorbis_pump_first_frame(stb_vorbis *f)
    res = vorbis_decode_packet(f, &len, &left, &right);
    if (res)
       vorbis_finish_frame(f, len, left, right);
+   f->current_playback_loc = 0;
+   f->current_playback_loc_valid = TRUE;
    return res;
 }
 
@@ -4302,6 +4313,14 @@ int stb_vorbis_get_sample_offset(stb_vorbis *f)
       return -1;
 }
 
+int stb_vorbis_get_playback_sample_offset(stb_vorbis *f)
+{
+   if (f->current_playback_loc_valid)
+      return f->current_playback_loc;
+   else
+      return -1;
+}
+
 stb_vorbis_info stb_vorbis_get_info(stb_vorbis *f)
 {
    stb_vorbis_info d;
@@ -4913,13 +4932,16 @@ int stb_vorbis_seek_frame(stb_vorbis *f, unsigned int sample_number)
    }
    // the next frame should start with the sample
    if (f->current_loc != sample_number) return error(f, VORBIS_seek_failed);
+   f->current_playback_loc = sample_number;
    return 1;
 }
 
 int stb_vorbis_seek(stb_vorbis *f, unsigned int sample_number)
 {
-   if (!stb_vorbis_seek_frame(f, sample_number))
+   if (!stb_vorbis_seek_frame(f, sample_number)) {
+      f->current_playback_loc_valid = FALSE;
       return 0;
+   }
 
    if (sample_number != f->current_loc) {
       int n;
@@ -4929,6 +4951,9 @@ int stb_vorbis_seek(stb_vorbis *f, unsigned int sample_number)
       assert(f->channel_buffer_start + (int) (sample_number-frame_start) <= f->channel_buffer_end);
       f->channel_buffer_start += (sample_number - frame_start);
    }
+
+   f->current_playback_loc_valid = TRUE;
+   f->current_playback_loc = sample_number;
 
    return 1;
 }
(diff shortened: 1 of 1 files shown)


stb_vorbis.c
```diff
@@ -86,6 +86,25 @@
 extern "C" {
 #endif
 
+typedef unsigned char  uint8;
+typedef   signed char   int8;
+typedef unsigned short uint16;
+typedef   signed short  int16;
+typedef unsigned int   uint32;
+typedef   signed int    int32;
+#if defined(__x86_64) || defined(__x86_64__) || defined(__amd64) || defined(_M_X64) \
```
sezero, line 95: This is definitely wrong: `_M_X64` is an MSVC thing and long is always 32 bits in windows. The list can also be much simplified, to something like:
```c
#if defined(_LP64) || defined(__LP64__)
typedef unsigned long       uint64;
typedef   signed long        int64;
#elif defined(_WIN64)
typedef unsigned __int64    uint64;
typedef   signed __int64     int64;
#else /* assuming 32-bit arches: */
typedef unsigned long long  uint64;
typedef   signed long long   int64;
#endif
```

sezero: Those `_M_ARM64` etc are all MSVC. All non-windows 64 bit arches follow the LP64 model so they define `_LP64` or `__LP64__` or both, hence the shortcut. Only windows is the weirdo who follows LLP64, hence the `_WIN32` special case.

stb_vorbis.c
```diff
@@ -92,14 +92,18 @@ typedef unsigned short uint16;
 typedef   signed short  int16;
 typedef unsigned int   uint32;
 typedef   signed int    int32;
-#if defined(__x86_64) || defined(__x86_64__) || defined(__amd64) || defined(_M_X64) \
+#if defined(__x86_64) || defined(__x86_64__) || defined(__amd64) \
```
sezero, line 95: All these three already define `_LP64` or `__LP64__` or both, they are nonsense here.

stb_vorbis.c
```diff
@@ -92,14 +92,18 @@ typedef unsigned short uint16;
...
+    || defined(_LP64) || defined(__LP64__) \
     || defined(__ARM_ARCH_8__) || defined(__ARM_ARCH_8A) || defined(__ARM_ARCH_8A__) || defined(__ARM_ARCH_8R__) || defined(__ARM_ARCH_8M__) \
     || (defined(__TARGET_ARCH_ARM) && __TARGET_ARCH_ARM >= 8) \
     || (defined(__ARM_ARCH) && __ARM_ARCH >= 8) \
     || defined(_M_ARM64) || defined(__aarch64__) \
     || defined(__ppc64__) || defined(__powerpc64__) || defined(__64BIT__) /* On 64-bit architectures */
 typedef unsigned long  uint64;
 typedef   signed long   int64;
+#elif defined(_WIN64)
```
sezero, line 104: `_WIN64` should be `_WIN32`

stb_vorbis.c
sezero: Any checks other than `_LP64` or `__LP64__` can and should go away here.

sezero: Yes

Wohlstand: Should be fine now.

stb_vorbis.c
```diff
@@ -1469,7 +1511,7 @@ static int start_page_no_capturepattern(vorb *f)
       return error(f, VORBIS_unexpected_eof);
    // assume we _don't_ know any the sample position of any segments
    f->end_seg_with_known_loc = -2;
-   if (loc0 != ~0U || loc1 != ~0U) {
+   if (loc != ~(uint64)0U) {
```
sezero, line 1514: The `U` suffix is funny here. The constant can simply be `~((uint64)0)`

stb_vorbis.c
```diff
@@ -4427,7 +4479,7 @@ static int vorbis_search_for_page_pushdata(vorb *f, uint8 *data, int data_len)
             f->next_seg = -1;       // start a new page
             f->current_loc = f->scan[i].sample_loc; // set the current sample location
                                     // to the amount we'd have decoded had we decoded this page
-            f->current_loc_valid = f->current_loc != ~0U;
+            f->current_loc_valid = f->current_loc != ~(int64)0U;
```
sezero, line 4482: Like above, lose the `U` suffix. The constant can simply be `~((int64)0)`

Wohlstand: // better say "remove `U` suffix", the "lose" in this case sounds absolutely weird to me... "Hey, dude, lose your wallet on the bus seat! I wanna steal it!"

Done

stb_vorbis.c
```diff
@@ -4997,20 +5055,17 @@ unsigned int stb_vorbis_stream_length_in_samples(stb_vorbis *f)
       // parse the header
       getn(f, (unsigned char *)header, 6);
       // extract the absolute granule position
-      lo = get32(f);
-      hi = get32(f);
-      if (lo == 0xffffffff && hi == 0xffffffff) {
+      total = get64(f);
+      if (total == 0xffffffffffffffff) {
```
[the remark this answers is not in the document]
Wohlstand: Done

stb_vorbis.c
sezero: > // better say "remove `U` suffix", the "lose" in this case sounds absolutely weird to me... "Hey, dude, lose your wallet on the bus seat! I wanna steal it!"

Eh?? It's a common expression, never thought about it that way.

musicinmybrain: “Lose” for “intentionally get rid of,” especially as a suggestion or command, is a common US English slang idiom since at least the mid-20th century. Presumably it’s not quite universal among global English speakers.

sezero: Possibly, I haven't (yet) followed the code thoroughly.

stb_vorbis.c
Wohlstand: Fixed.

stb_vorbis.c
```diff
@@ -1434,6 +1457,19 @@ static int set_file_offset(stb_vorbis *f, unsigned int loc)
    #endif
 }
 
+static __forceinline uint64 get64raw(uint8 *f)
+{
+   uint64 x;
+   x = f[0];
+   x += f[1] << 8;
```
slouken, line 1462: You need a cast here as well. uint8 shifted left by 8 is 0.

nothings: It is false that you need casts here, uint8 shifts are promoted to int before shifting, so shift by 8 is not 0. (Technically shift by 24 does though, since it promotes to int and then you're shifting into the sign bit, which is why get32() already had a cast for the shift-by-24 step.)

slouken: Ah, thanks for the info.

PR #1333, opened 2022-05-28
Documentation added to Dynamic array initialization with known initial size

BlazesRus:
Documentation added to Dynamic array to make easier for someone to know how to use dynamic array macro

### DIFF
--- a/stb_ds.h
+++ b/stb_ds.h
@@ -404,6 +404,8 @@ CREDITS
 #define arraddn     stbds_arraddn // deprecated, use one of the following instead:
 #define arraddnptr  stbds_arraddnptr
 #define arraddnindex stbds_arraddnindex
+//Initialyze array with struct ElementType* ArrayName = arrsetlen(ArrayName, num_elems);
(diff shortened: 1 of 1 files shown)


nothings:
That doesn't work, as the array pointer needs to be NULL when you call arrsetlen, not undefined.

BlazesRus:
Fixed it changed it to:
//Initialyze array with struct ElementType* ArrayName = NULL;
//arrsetlen(ArrayName, num_elems);
//Equavalent to ANSI C99/C89 version of struct ElementType ArrayName[num_elems];

nothings:
I'm not sure how much value that explanation adds, as the standard way to create an array of N elements is push the N elements individually with `arrput`; `arrsetlen` works and is more efficient, but I'm not sure it's worth adding explicitly. Even if we did want to add this, that's probably not the right place for it, it should appear in the documentation section starting at line 69, either as an introductory thing or as part of the documentation for `arrsetlen` at line 134.

rygorous:
I don't think we want to take this, closing.

PR #1336, opened 2022-06-17
stb_include: use const-correct interface

JuanDiegoMontoya:
stb_include's interface should be const-correct to give the programmer guarantees about their arguments and to prevent the need for an explicit cast to pass `const char*`s (e.g. string literals) to read-only parameters.


gue-ni:
Are there any updates on this? Could this be merged?

PR #1339, opened 2022-06-26
stb_truetype: only free if vertices was allocated

ryuukk:


stb_truetype.h
```diff
@@ -3746,7 +3746,8 @@ STBTT_DEF unsigned char *stbtt_GetGlyphBitmapSubpixel(const stbtt_fontinfo *info
          stbtt_Rasterize(&gbm, 0.35f, vertices, num_verts, scale_x, scale_y, shift_x, shift_y, ix0, iy0, 1, info->userdata);
       }
    }
-   STBTT_free(vertices, info->userdata);
+   if (vertices)
+      STBTT_free(vertices, info->userdata);
```
N-R-K, line 3750: `free(3)` is defined to be a no-op when the argument is NULL. This would only be useful if a custom free is being used which doesn't respect the standard api.

And if that's the case, then there's also another unconditional free at line 3723 which will need to be checked as well.

N-R-K: With some more thought, if someone is using a custom-free which doesn't deal with NULL, then he should/can define it like this instead:

```c
#define STBTT_free(p, u) do { if (p) custom_free(p); } while (0)
```

rygorous:
As noted in N-R-K's review, we already require the "free" function to implement the standard behavior of doing nothing when passed a null pointer.

PR #1347, opened 2022-07-12
Fix typos

kianmeng:


ice0 (approved):

rygorous:
There is already another PR that does this so this is a duplicate; on a more general note, please do not submit grab-bag PRs that change multiple libraries at once, it's a pain to deal with.

PR #1350, opened 2022-07-24
Add MSVC address sanitizer support to stb_sprintf

aholkner:


jlaumon:
Any chance this gets merged eventually?

TheSandvichMaker:
To my testing, this doesn't work unless the declspec annotation is attached to the function _declaration_, rather than the definition. The `STBSP__ASAN` macro is currently only added to function definitions - so the PR as is doesn't work.

r-lyeh:
I did not know this PR existed and originally used a similar implementation than aholkner's. But it did not work for me.
Then I read what TheSandvichMaker proposed (good hint thanks) and tried that but it did not work either (lol).
At this point I am not sure if this Microsoft directive is a joke or I am an utterly idiot. Probably the latter.
I ended up patching manually the two places where the sanitizer complains and I can finally use ASAN in conjunction with stb_sprintf.

Havent really tested it that much but I guess some of you guys might start from here.

```C
            // Check if the next 4 bytes contain %(0x25) or end of string.
            // Using the 'hasless' trick:
            // https://graphics.stanford.edu/~seander/bithacks.html#HasLessInWord
            stbsp__uint32 v, c;
#if defined(_MSC_VER) && defined(__SANITIZE_ADDRESS__) && __SANITIZE_ADDRESS__
            // go safe path
            v = f[0];
            if( f[0] ) { v |= f[1] << 8;
            if( f[1] ) { v |= f[2] << 16;
            if( f[2] ) { v |= f[3] << 24; }}}
#else
            v = *(stbsp__uint32 *)f;
#endif
```

```C
#if defined(_MSC_VER) && defined(__SANITIZE_ADDRESS__) && __SANITIZE_ADDRESS__
   // go safe path
#else
   // scan over 4 bytes at a time to find terminating 0
   // this will intentionally scan up to 3 bytes past the end of buffers,
   // but becase it works 4B aligned, it will never cross page boundaries
   // (hence the STBSP__ASAN markup; the over-read here is intentional
   // and harmless)
   while (limit >= 4) {
      stbsp__uint32 v = *(stbsp__uint32 *)sn;
      // bit hack to find if there's a 0 byte in there
      if ((v - 0x01010101) & (~v) & 0x80808080UL)
         break;

      sn += 4;
      limit -= 4;
   }
#endif

   // handle the last few characters to find actual size
   while (limit && *sn) {
      ++sn;
      --limit;
   }
```

alektron:
I encountered the same issues. Here are my findings and what worked in the end.
First of all, this PR as well as #1477 and #1622 are of course correct in adding:
```cpp
#elif defined(_MSC_VER)
  #if defined(__SANITIZE_ADDRESS__) && __SANITIZE_ADDRESS__
  #define STBSP__ASAN __declspec(no_sanitize_address)  
#endif
```
But that is not enough since @TheSandvichMaker also seems to be correct:
> To my testing, this doesn't work unless the declspec annotation is attached to the function declaration, rather than the definition. The STBSP__ASAN macro is currently only added to function definitions - so the PR as is doesn't work.

So additionally the `STBSP__PUBLICDEC` macros have to be adjusted:
```cpp
#ifdef STB_SPRINTF_STATIC
#define STBSP__PUBLICDEC static STBSP__ASAN
#define STBSP__PUBLICDEF static STBSP__ASAN
#else
#ifdef __cplusplus
#define STBSP__PUBLICDEC extern "C" STBSP__ASAN
#define STBSP__PUBLICDEF extern "C" STBSP__ASAN
#else
#define STBSP__PUBLICDEC extern STBSP__ASAN
#define STBSP__PUBLICDEF STBSP__ASAN
#endif
#endif
```
It seems like `STBSP__ASAN` in `STBSP__PUBLICDEF` is not even required for MSVC but it also does not hurt. Since it is required for the other compilers we need to keep it.
Now it would be interesting if the changes in `STBSP__PUBLICDEC` can interfere with the other compilers. I did not check that.
In that case one would have to add a separate ifdef block for MSVC like in #1622.

Note however that this statement by @tim-rex in #1622 does not seem quite correct:
> Note that _declspec for MSVC must appear before the storage specifier

I am not sure about this in general but in the this specific case `STBSP__ASAN` after the storage specifier worked for me, while in front of it did not.

PR #1373, opened 2022-09-13
Update stb_truetype.h to make it usable for IoT/embedded devices as well

codewitch-honey-crisis:
Makes the library configurable to be suitable on IoT devices or otherwise devices without very much RAM (<256kB)

Changes only take place if configured as such. Otherwise the library behaves as normal wherever possible: Edit: I changed functions that returned pointers to return integer offsets instead that way they can be used with pointers *or* streams.

Any contributions i make are under "honey the codewitch" as that is my business and brand. I have a credit card and everything. So if you see that, well, I'm not just being silly for effect - I actually am silly.

I did note at the top that I forked 1.26 

I added my name at the top but not even under contributors because I didn't know if you were even interested in this. It's being used in LVGL going forward, whether or not you're interested, but I think it would be cool to have worked together to make this more accessible in more places and platforms! =)

### DIFF
--- a/stb_truetype.h
+++ b/stb_truetype.h
@@ -921,7 +921,7 @@ extern "C" {
     STBTT_DEF void stbtt_FreeShape(const stbtt_fontinfo* info, stbtt_vertex* vertices);
     // frees the data allocated above
 
-    STBTT_DEF stbtt_uint32 stbtt_FindSVGDoc(const stbtt_fontinfo* info, int gl);
+    STBTT_DEF unsigned int stbtt_FindSVGDoc(const stbtt_fontinfo* info, int gl);
     STBTT_DEF int stbtt_GetCodepointSVG(const stbtt_fontinfo* info, int unicode_codepoint, unsigned int* svgOfs);
     STBTT_DEF int stbtt_GetGlyphSVG(const stbtt_fontinfo* info, int gl, unsigned int* svgOfs);
     // fills svg with the character's SVG data.
@@ -1104,7 +1104,7 @@ extern "C" {
     // returns 1/0 whether the first string interpreted as utf8 is identical to
     // the second string interpreted as big-endian utf16... useful for strings from next func
 
-    STBTT_DEF stbtt_uint32 stbtt_GetFontNameString(const stbtt_fontinfo* font, int* length, int platformID, int encodingID, int languageID, int nameID);
+    STBTT_DEF unsigned int stbtt_GetFontNameString(const stbtt_fontinfo* font, int* length, int platformID, int encodingID, int languageID, int nameID);
 
     // returns the string (which may be big-endian double byte, e.g. for unicode)
     // and puts the length in bytes in *length.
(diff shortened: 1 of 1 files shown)


codewitch-honey-crisis:
Welp, I'm stumped. Where is main being referenced?

$ make all
gcc -I.. -Wno-pointer-to-int-cast -Wno-int-to-pointer-cast -DSTB_DIVIDE_TEST ../stb_vorbis.c test_c_compilation.c test_c_lexer.c test_dxt.c test_easyfont.c test_image.c test_image_write.c test_perlin.c test_sprintf.c test_truetype.c test_voxel.c -lm
gcc -I.. -Wno-write-strings -DSTB_DIVIDE_TEST -std=c++0x test_cpp_compilation.cpp -lm -lstdc++
/usr/lib/gcc/x86_64-linux-gnu/5/../../../x86_64-linux-gnu/crt1.o: In function `_start':
(.text+0x20): undefined reference to `main'
collect2: error: ld returned 1 exit status
Makefile:9: recipe for target 'all' failed
make: *** [all] Error 1
The command "make all" exited with 2.

nothings:
Don't worry about it, the tests are mostly for my purpose, not for your testing. (The details is they are standalone and require a -D on the commandline to determine which one to have a main. And in some cases the data required to run them isn't in the report.)

codewitch-honey-crisis:
Thanks! I thought it was something I did. My initial comment said I did not change the API. I fibbed. The truth is I forget I hand to change, like the FindSVGDoc function and one or two others that returned pointers to return integer offsets that could be used with either a pointer or stream.

nothings:
This is the kind of robust complexity that I generally try to avoid in the stb libraries. For example, libjpg supports streaming image, whereas stb_image requires the whole image to be loaded at once. I generally leave it to the more robust, official libraries to support these sorts of things, and the stb libraries are meant to be simpler to use, to not worry too much about memory management, etc.

codewitch-honey-crisis:
Fair enough. This library is *almost* perfect for IoT and embedded gadgets. This bridges the "almost" gap. So maybe it's worth considering in this case? check out https://lvgl.io 
Your code will be running on that going forward.
It already is being used in https://honeythecodewitch.com/gfx using a previous version I ported.

Most of your libraries are clearly for gaming engines. This is too. But trust me, in this case, it's much more, and fills a sorely needed gap in support across a variety of devices.

Do what you want, but I'd be remiss if I didn't try to twist your arm on this issue a little. =)

rygorous:
Don't take this the wrong way, but there are two big issues here:
1. as Sean mentioned, this is a huge change and looks to be adding a lot of functionality that is out of scope for what stb_truetype tries to do
2. independent of that, this PR completely changes indentation style all over the file which makes it hard to review what the actual diffs are.

Closing this PR as it is. If you have some specific things you'd like changed, feel free to open issues/discussion posts about these and we can talk out how much (if any) of it we're OK with being in the main library, and what we consider to be out of scope.

That doesn't mean we're categorically unwilling to take your code for functionality that we agree would be OK to have in the library, but we will not merge a 8100-line diff that rewrites essentially the entire file, because not only is that too big a change for one commit and too big to review as one unit, it would also conflict with essentially every other PR that touches stb_truetype which completely breaks our maintenance workflow.

For smaller fixes or functionality it's fine to make just a PR, but for something big like this just submitting a pile of code doesn't work. For all we know once a change is in we'll never hear from the original author again, but we're stuck maintaining the result indefinitely, so we're relatively picky.

PR #1374, opened 2022-09-13
Added null checking for `stbtt_GetFontBoundingBox`

PhucXDoan:


rygorous:
Is there a particular reason you need this? You _are_ supposed to give us pointers to all 4 values. You're free to ignore them after if you don't need them.

PR #1375, opened 2022-09-14
Added missing fclose in stb_include.h

cozis:
The "stb_include_load_file" function in "stb_include.h" leaks the file descriptor in case of malloc failure. I just added the missing fclose call before the early return statement

### DIFF
--- a/stb_include.h
+++ b/stb_include.h
@@ -33,6 +33,7 @@
 //
 // Fixes:
 //  Michal Klos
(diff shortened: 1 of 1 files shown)


paulwratt (approved):

ocornut:
FYI this is the same as #1310

nothings:
duplicate of #1310.

PR #1380, opened 2022-09-19
Adding support for octal and hexadecimal chars

danil-kondr2016:
This commit adds support of octal and hexadecimal codes of characters in string and char literals.

### DIFF
--- a/stb_c_lexer.h
+++ b/stb_c_lexer.h
@@ -31,13 +31,13 @@
 //     - only tested default-config paths by eyeballing output of self-parse
 //
 //     - haven't implemented multiline strings
-//     - haven't implemented octal/hex character constants
 //     - haven't implemented support for unicode CLEX_char
 //     - need to expand error reporting so you don't just get "CLEX_parse_error"
 //
 // Contributors:
 //   Arpad Goretity (bugfix)
 //   Alan Hickman (hex floats)
+//   Daniel Kondratenko (octal and hex chars)
 //
 // LICENSE
 //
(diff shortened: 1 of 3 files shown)


nothings:
if I'm reading the code right, there are multiple problems that should be straightforward to fix

- in octal parsing, the first if() statement doesn't update `*q`, so a one-character octal value won't be parsed properly (e.g. "\1foo")
- in octal parsing, the if()s aren't nested so digits later in a string can mess it up (e.g. try "\1f2")
- in hex parsing, the if()s aren't nested so digits later in a string can mess it up (e.g. try "\x1z1")
- hex character size is implementation-defined in C, so parsing isn't limited to 3 characters. should just do a full-sized hex parse, error if it overflows. The overflow size is the size of the literal used; for character literals that's `long int_number` and for string literals it's sizeof(char) so check the result fits in 8-bits (but use the full parse). Note that this overflow checking is NOT done elsewhere in stb_c_lexer, but we should improve that, so this is a good place to start.

style improvements:
- should only write to `*q` once at end for clarity in both octal and hex constants, might be clearest by advancing `p` as you read and then doing `*q = p`.
- although it's inconsistent with the value types, should probably parse the initial hex value in an `unsigned` type, since that's a more natural fit for hex constants

nothings:
Note that it looks like STB_C_LEXER_SELF_TEST just prints back the input string for string literals and char literals, which means you can't tell from that test that they're computing the wrong value in those cases. If the end location of a char literal is misparsed, that may or may not be recognized, but getting the end of the char literal wrong inside of a string literal will not be visible as well.

danil-kondr2016:
I have fixed errors which you told about.

PR #1383, opened 2022-09-20
Support implicit r and rg channel loading, only convert to luma then explicitly done.

tloch14:
In the age of PBR, the common case is not to load images with luma conversion. This PR makes it so that when supplying numerical integers 1...4 to the `desired_channels` argument, no additional operations are done. `STBI_grey` and `STBI_grey_alpha` have been moved to new identifiers where a special bit is set to identify them, while `STBI_r` and `STBI_rg` have taken their place. In this way, `STBI_grey` and `STBI_grey_alpha` now hints to the `desired_channels` argument that the user would like a luma conversion done.

Rational:
1.  Textures related to PBR assets often come with occlusion, roughness, and metallic packed into the r, g, and b channels. When you want to only grab the r and g channels (as asset authors often put all 1 or all 0 in the b channel if it is effectively unused), stb_image confusingly will do a luma conversion.
2. Performing a luma conversion in the case of 1 or 2 required channels goes against the law of least surprise.
3. Performing a luma calculation is additional work, and strictly speaking unrelated to the isolated goal of loading images. It's common to make such conversion operations at other times in asset creation.
4. OpenGL deprecated GL_LUMINANCE and co. in 2008 with OpenGL 3.0, then dropped them entirely a year later. The recommendation from OpenGL is to use R8, RG8, R16, or RG16 instead. This change follows that guideline.

### DIFF
--- a/stb_image.h
+++ b/stb_image.h
@@ -164,11 +164,13 @@ RECENT REVISION HISTORY:
 // An output image with N components has the following components interleaved
 // in this order in each pixel:
 //
-//     N=#comp     components
-//       1           grey
-//       2           grey, alpha
-//       3           red, green, blue
-//       4           red, green, blue, alpha
+//     N=#comp            components
+//       1                  red
+//       2                  red, green
+//       3                  red, green, blue
+//       4                  red, green, blue, alpha
+//       STBI_grey          grey
+//       STBI_grey_alpha    grey, alpha
 //
 // If image loading fails for any reason, the return value will be NULL,
 // and *x, *y, *channels_in_file will be unchanged. The function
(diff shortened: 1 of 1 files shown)


nothings:
This would break everyone using 1 & 2 currently. Not acceptable.

tloch14:
I thought that might be the response.

Would it be acceptable if STBI_r and STBI_rg were entirely new identifiers with new values?

nothings:
Yes, I think adding new enum values without changing the old would be okay on the API front. The only other question is how much code needs to change, and how much extra maintenance overhead that might induce. 

E.g., ignoring the API issue, I don't know if the submitted patch is actually correct. IIRC, PNG and maybe JPG will do some channel conversions automatically without ever calling `stbi__convert_format`. So it will be necessary to check those and make sure the new cases are handled (either as-you-go, or by loading normally and relying on stbi__convert_format; to be explicit, the point is that, again IIRC, PNG loader looks at desired_channels and tries to fix things up to fit, and it may expect only those 4 specific values to be present, and to be able to e.g. compare to the number of channels in the file. Note that the 1,2,3,4 convention in the API exactly matches how PNG's internal code for what channels are present, and the code probably relies on that. I don't know for sure, I wrote it 15 years ago.)

tloch14:
PNG code path is the one in particular I was using, so I'm fairly certain that uses `stbi__convert_format` with no additional work being done. But it does look like `load_jpeg_image` uses a separate code path from `stbi__convert_format`. 

I'll make the API changes and then look into the jpeg loader.

PR #1384, opened 2022-09-23
Fix PNM decoding 16bpp 3-channel to 4-channel

AdrianAtGoogle:
(this issue got discovered by fuzzing)


rygorous:
That fix landed as part of stb_image 2.28, closing.

PR #1389, opened 2022-10-09
stb_truetype: Missing a return in stbtt_GetGlyphBox

Scr3amer:
Hi !

Correct me if I am wrong but i think there is a missing return in stbtt_GetGlyphBox

Cheers


Scr3amer:
Hello guys !

Any update ? Is my understanding of the code correct and we are missing that return statement or do we ignore the return value on purpose ?

Cheers

PR #1396, opened 2022-10-21
simplify bucket search loops

BenFradella:
Do the bucket search loops that start at pos and loop back through the start in a single loop, rather than being broken out into pairs of identical loops.

This does actually fix a bug. Lines 1394,1395 were missing from the second loop, so the temp key was only set if we found a match in the first loop. I kept the contents of the first loop and deleted the second, so the temp key will get set for any position in the bucket now.

### DIFF
--- a/stb_ds.h
+++ b/stb_ds.h
@@ -382,6 +382,7 @@ CREDITS
     Macoy Madson
     Andreas Vennstrom
     Tobias Mansfield-Williams
+    Ben Fradella
 */
 
 #ifdef STBDS_UNIT_TESTS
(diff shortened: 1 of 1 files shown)


rygorous:
Bug fix is appreciated, will probably not take this exact implementation but thank you!

BenFradella:
Not sure how I closed the PR. Seems like it was triggered by a force push. That aside, I thought of a better way to do this than what I had before, so I figured I'd make that change and push it, if only for my own sake. Should be easier for the compiler to unroll the loops and whatnot

PR #1404, opened 2022-11-06
stb_image: Fix gif two_back memory address

cocoa-xu:
Hi, this should be a simple fix: since `out` is always the starting address, `two_back` should be calculated as follows,

```c
if (layers >= 2) {
    two_back = out + ((layers - 2) * stride);
}
```

otherwise, it would cause a segmentation fault.

Testing image [test_alpha.gif.zip](https://github.com/nothings/stb/files/9945209/test_alpha.gif.zip)

### DIFF
--- a/stb_image.h
+++ b/stb_image.h
@@ -113,7 +113,7 @@ RECENT REVISION HISTORY:
     Baldur Karlsson         Kevin Schmidt      JR Smith             github:Michaelangel007
                             Brad Weinberger    Matvey Cherevko      github:mosra
     Luca Sas                Alexander Veselov  Zack Middleton       [reserved]
-    Ryan C. Gordon          [reserved]                              [reserved]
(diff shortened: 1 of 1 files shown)


geneotech:
Thank you so much, this actually fixes a crash I had on a large GIF file. This PR should be accepted ASAP.

cocoa-xu:
Hi @nothings, sorry for the ping here. I noticed that there seems to be more people experiencing this issue, https://github.com/nothings/stb/issues/1504 and https://github.com/nothings/stb/issues/1688, and I believe this PR should have properly addressed the root cause. I'll be happy to write a unit test for it if there's a need for it!

641i130:
This resolved certain gifs causing seg faults on an application I work on for work. Thank you for your work @cocoa-xu !

PR #1414, opened 2022-12-05
stb_image: add struct version of stbi_info and an iPhone PNG flag

hikari-no-yume:
This makes it possible to handle the differences between iPhone and non-iPhone PNGs without having to independently know what kind of PNG it is, and without requiring stb_image to convert for you.

The documentation for iPhone PNGs is also rewritten.

(By the way, I'm wondering about making another, related contribution: instead of un-premultiplying alpha when loading an “iPhone” PNG, what if there would be an option to premultiply alpha when loading a non-“iPhone” PNG? Since game developers are a significant target audience for this library, that seems potentially useful. I don't immediately need that though, so I won't do it unless someone else is interested.)

### DIFF
--- a/stb_image.h
+++ b/stb_image.h
@@ -91,7 +91,7 @@ RECENT REVISION HISTORY:
  Optimizations & bugfixes                  Mikhail Morozov (1-bit BMP)
     Fabian "ryg" Giesen                    Anael Seghezzi (is-16-bit query)
     Arseny Kapoulkine                      Simon Breuss (16-bit PNM)
-    John-Mark Allen                        hikari_no_yume (stbi_info2)
+    John-Mark Allen                        hikari_no_yume (stbi_info2, is_iphone_png)
     Carmelo J Fdez-Aguera
 
  Bug & warning fixes
@@ -320,15 +320,19 @@ RECENT REVISION HISTORY:
 //
 // iPhone PNG support:
 //
-// We optionally support converting iPhone-formatted PNGs (which store
-// premultiplied BGRA) back to RGB, even though they're internally encoded
-// differently. To enable this conversion, call
-// stbi_convert_iphone_png_to_rgb(1).
+// iPhone-formatted PNGs, also known as CgBI PNGs, are supported. Unlike
+// standard PNGs, these store colors in BGR/BGRA order with premultiplied alpha.
 //
-// Call stbi_set_unpremultiply_on_load(1) as well to force a divide per
-// pixel to remove any premultiplied alpha *only* if the image file explicitly
-// says there's premultiplied data (currently only happens in iPhone images,
-// and only if iPhone convert-to-rgb processing is on).
+// By default, stb_image will load the colors in the same format they were
+// stored in. If you need to know which kind of PNG was loaded, and thus the
+// color format, call one of the stbi_info2 functions and check is_iphone_png.
+//
+// Alternatively, if you want stb_image to convert the color order to RGB/RGBA,
+// call stbi_convert_iphone_png_to_rgb(1).
+//
+// Call stbi_set_unpremultiply_on_load(1) as well to force a divide per pixel
+// pixel to remove any premultiplied alpha. Note that this currently only works
+// when color order conversion is also in use.
 //
 // ===========================================================================
 //
@@ -508,6 +512,7 @@ typedef struct
 
    int is_16_bit;
    int is_hdr;
+   int is_iphone_png;
 } stbi_info_res;
 
 STBIDEF stbi_info_res stbi_info2_from_memory(stbi_uc const *buffer, int len);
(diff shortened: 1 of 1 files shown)


hikari-no-yume:
I tested this on both a standard and a CgBI PNG, with the following C application:

```
#include "stb_image.h"

int main(int argc, char *argv[])
{  
    char *filename = argv[1]; 
   
    int x, y, channels_in_file;
    stbi_uc *res = stbi_load(filename, &x, &y, &channels_in_file, 4);
    int was_iphone = stbi_was_iphone_png();
    printf("%p, %d, %d, %d, %d\n", (void*)res, x, y, channels_in_file, was_iphone); 
}  
```

I also tested it with `STBI_NO_THREAD_LOCALS` defined.

nothings:
I think it makes more sense to be a query you run before decoding the image, a la stb_info, rather than doing it after the fact. That also avoids the clearing issue.

hikari-no-yume:
That's reasonable, I'll attempt to rework it. Personally I always am a bit suspicious of cases where you parse something multiple times because there's a risk of the two parses diverging in their result (I note there's an open issue about one of those), but it seems like the CgBI case is trivial enough that it shouldn't be much of an issue.

hikari-no-yume:
I can think of two ways to rework it:

* There's already the `stbi_is_hdr` and `stbi_is_16_bit` function families, so if I follow that precedent, I could add a `stbi_is_iphone_png` family.
* Add an extended version of `stbi_info` that also returns a flags field. Then you could use it like:
  ```c
  int res = stbi_info(filename, &x, &y, &comp, &flags);
  if (res && (flags & STBI_flag_is_iphone_png)) { … }
  ```

I think I prefer the latter approach because it avoids parsing the file an extra time if you want other info, and this way there won't need to be any more “is” functions added in future.

Do you have any opinion on this?

nothings:
I think my prefernce would be an extended info function, but instead of a flags field it should return a struct with all the info (including width & height). I *doubt* we'll run out of flags or need to add fields that aren't flags, but this way we're covered in that unlikely eventuality, and we don't have to bring bitmasking operations in it, each flag can just be a boolean (well, because it's stb_image, just make them ints or uint8s or such).

hikari-no-yume:
Ah, that's a great idea! Any preference for directly returning a struct, initialising a struct via a pointer parameter, or returning a pointer to an allocated struct?

(Side-note: it might be nice if there was a variant of `stbi_load` that also gave you the info struct, but I wouldn't do that in the same change.)

nothings:
I guess might as well return a struct, since it lets the library client declare and call all at once.

I wouldn't mess with stbi_load, because unfortunately there's a lot of functions there, and we'd have to make a variant for all of them, and that's just expanding the API too much IMO, compared to this which is just two new functions (file and memory).

hikari-no-yume:
I reworked it along those lines now. Feel free to nitpick on naming details. I'm not sure I love `stbi_info2` and `stbi_info_res`…

hikari-no-yume:
Tiny app I used to test the new version:

```c
#include "stb_image.h"

int main(int argc, char **argv)
{
    if (argc != 2) abort();
    char *filename = argv[1];

    int old_x, old_y, old_comp;
    int old_supported = stbi_info(filename, &old_x, &old_y, &old_comp);
    int old_is_16_bit = stbi_is_16_bit(filename);
    int old_is_hdr = stbi_is_hdr(filename);

    stbi_info_res new = stbi_info2(filename);

    printf("supported: %d %d\n", old_supported, new.supported);
    if (new.supported) {
        printf("- x: %d %d\n", old_x, new.x);
        printf("- y: %d %d\n", old_y, new.y);
        printf("- comp: %d %d\n", old_comp, new.comp);
        printf("- is_iphone_png: %d\n", new.is_iphone_png);
    }
    printf("is_16_bit: %d %d\n", old_is_16_bit, new.is_16_bit);
    printf("is_hdr: %d %d\n", old_is_hdr, new.is_hdr);
}
```

thedmd:
This is how I end up expanding my copy of `stb_image`, by introducing more general `stbi_image_info` structure:
```c
enum
{
    STBI_type_unknown = 0,

    STBI_type_jpeg   = 1,
    STBI_type_png    = 2,
    STBI_type_bmp    = 3,
    STBI_type_tga    = 4,
    STBI_type_psd    = 5,
    STBI_type_hdr    = 6,
    STBI_type_pic    = 7,
    STBI_type_gif    = 8,
    STBI_type_pnm    = 9,
};

typedef struct stbi_image_info
{
    int type;
    int x, y;
    int comp;
    int dpm_x, dpm_y;
} stbi_image_info;
```

In my case I needed DPI information and occasionally file type.

> I wouldn't mess with stbi_load, because unfortunately there's a lot of functions there, and we'd have to make a variant for all of them

It was a chore, but ability to quickly expand `stb_image` to return that one extra bit of data made it worth doing.
I did that on in my copy for functions I use, where API returning `int *x, int *y, int *comp` was implemented over one that accept pointer to `stbi_image_info` instead.

For example:
```c
STBIDEF stbi_uc *stbi_load_from_callbacks(stbi_io_callbacks const *clbk  , void *user, int *x, int *y, int *channels_in_file, int desired_channels);
```

Got counterpart:
```c
STBIDEF stbi_uc *stbi_load_from_callbacks_info(stbi_io_callbacks const *clbk  , void *user, int desired_channels, stbi_image_info *info);
```

I would like advocate to pursue this direction of making `stb_image` expandable if possible.

hikari-no-yume:
Hmm, for what it's worth, it'd be trivial for me to add a type enum to the info struct my patch adds.

hikari-no-yume:
Is there anything I could do to help this get somewhere? If it feels too complicated, I'm happy to simplify it for example.

PR #1443, opened 2023-01-30
stb_image: fix implicit conversion warnings

### LINKED ISSUE #1444: stb_image: warnings MSVC: Implicity conversions/possible loss of data
**Describe the bug**
After some recent changes to stb_image.h file, when compiling with MSVC set to treat all warnings as errors, this library fails to compile at these lines with this message: `warning C4244: 'argument': conversion from '...' to '...', possible loss of data`
```
- stb_image.h(2224,32)
- stb_image.h(2281,35)
- stb_image.h(3425,49)
```

**Expected behavior**
Compilation with no warnings, the easiest fix would be to explicitly cast these variables.

Linked my pr as I am trying to fix it.

diamante0018:
Hello, first-time contributor here.
This pull request fixes 3 warnings that I get when using this library on the MSVC platform.
I have set my MSVC compiler to treat these warnings as errors so this is necessary for me to continue using this library without making any changes to the solution file on visual studio/turning off warnings.
Thanks.

Closes #1444

### DIFF
--- a/stb_image.h
+++ b/stb_image.h
@@ -3384,12 +3384,12 @@ static int stbi__decode_jpeg_header(stbi__jpeg *z, int scan)
    return 1;
 }
 
-static int stbi__skip_jpeg_junk_at_end(stbi__jpeg *j)
+static stbi_uc stbi__skip_jpeg_junk_at_end(stbi__jpeg *j)
 {
    // some JPEGs have junk at end, skip over it but if we find what looks
    // like a valid marker, resume there
    while (!stbi__at_eof(j->s)) {
-      int x = stbi__get8(j->s);
+      stbi_uc x = stbi__get8(j->s);
       while (x == 255) { // might be a marker
          if (stbi__at_eof(j->s)) return STBI__MARKER_none;
          x = stbi__get8(j->s);
@@ -3422,7 +3422,7 @@ static int stbi__decode_jpeg_image(stbi__jpeg *j)
          if (!stbi__process_scan_header(j)) return 0;
          if (!stbi__parse_entropy_coded_data(j)) return 0;
          if (j->marker == STBI__MARKER_none ) {
-         j->marker = (unsigned char) stbi__skip_jpeg_junk_at_end(j);
+         j->marker = stbi__skip_jpeg_junk_at_end(j);
             // if we reach eof without hitting a marker, stbi__get_marker() below will fail and we'll eventually return 0
          }
          m = stbi__get_marker(j);
--- a/stb_image.h
+++ b/stb_image.h
@@ -2221,7 +2221,7 @@ static int stbi__jpeg_decode_block(stbi__jpeg *j, short data[64], stbi__huffman 
    if (!stbi__addints_valid(j->img_comp[b].dc_pred, diff)) return stbi__err("bad delta","Corrupt JPEG");
    dc = j->img_comp[b].dc_pred + diff;
    j->img_comp[b].dc_pred = dc;
-   if (!stbi__mul2shorts_valid(dc, dequant[0])) return stbi__err("can't merge dc and ac", "Corrupt JPEG");
+   if ((dc > SHRT_MAX) || (dequant[0] > SHRT_MAX) || !stbi__mul2shorts_valid((short) dc, (short) dequant[0])) return stbi__err("can't merge dc and ac", "Corrupt JPEG");
    data[0] = (short) (dc * dequant[0]);
 
    // decode AC components, see JPEG spec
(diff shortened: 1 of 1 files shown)


nothings:
None of these seem like the correct way to fix the issue. For example, for security reasons, the code wants to check that multiplying two shorts gives a valid result. If one of the inputs isn't a short, but you cast it to a short for just that function call, you could truncate it smaller than it is, the security test passed, but then the actual multiply is done with a larger number, breaking the security check.

diamante0018:
So, if I understand this correctly, instead of casting you would prefer I clamp the value to be in between short min/max for that function call
(And unsigned char for the final warning)?
Is there a place in this repository where it's done already so I may have a look at it?

nothings:
I'm pretty much saying that @rygorous (who wrote this code) should take a look at it and figure out what's going on. Presumably either `dc` ought to be stored in a `short`, or `stbi__mul2shorts_valid` should take ints, or something.

diamante0018:
Roger, feel free to close this pr when it is fixed properly or I will come back to it in a bit to see when it has been fixed to close it myself 🙂

rygorous:
I didn't write it, Neil did, but I'll have a look later.

stb_image.h
```diff
@@ -2221,7 +2221,7 @@ static int stbi__jpeg_decode_block(stbi__jpeg *j, short data[64], stbi__huffman
    if (!stbi__addints_valid(j->img_comp[b].dc_pred, diff)) return stbi__err("bad delta","Corrupt JPEG");
    dc = j->img_comp[b].dc_pred + diff;
    j->img_comp[b].dc_pred = dc;
-   if (!stbi__mul2shorts_valid(dc, dequant[0])) return stbi__err("can't merge dc and ac", "Corrupt JPEG");
+   if (!stbi__mul2shorts_valid((short) dc, dequant[0])) return stbi__err("can't merge dc and ac", "Corrupt JPEG");
```
NeilBickford-NV, line 2224: Oh yep, this was my fault in #1297 ! Thank you for the fix. How about checking that `dc` is valid to cast to a short, and casting the `stbi__uint16 as well, like this?
```
if ((dc > SHRT_MAX) || (dequant[0] > SHRT_MAX) || !stbi__mul2shorts_valid((short) dc, (short)dequant[0])) return stbi__err("can't merge dc and ac", "Corrupt JPEG");
```

stb_image.h
```diff
@@ -2278,7 +2278,7 @@ static int stbi__jpeg_decode_block_prog_dc(stbi__jpeg *j, short data[64], stbi__
       if (!stbi__addints_valid(j->img_comp[b].dc_pred, diff)) return stbi__err("bad delta", "Corrupt JPEG");
       dc = j->img_comp[b].dc_pred + diff;
       j->img_comp[b].dc_pred = dc;
-      if (!stbi__mul2shorts_valid(dc, 1 << j->succ_low)) return stbi__err("can't merge dc and ac", "Corrupt JPEG");
+      if (!stbi__mul2shorts_valid((short) dc, 1 << j->succ_low)) return stbi__err("can't merge dc and ac", "Corrupt JPEG");
```
NeilBickford-NV, line 2281: Similarly, maybe this could check as well that `dc` can be casted to a short before doing so, like this?
```
if ((dc > SHRT_MAX) || !stbi__mul2shorts_valid((short) dc, 1 << j->succ_low)) return stbi__err("can't merge dc and ac", "Corrupt JPEG");
```

diamante0018:
Thanks for the feedback, let me know if the last cast to `unsigned char` is fine.
Also I'm not familiar with the code style of this project but I tried my best to follow it, please let me know if it is okay.

NeilBickford-NV:
Thank you, that looks good to me from a security perspective - `stbi__skip_jpeg_junk_at_end()` always returns a value between 0 and 0xff inclusive, so another solution is it could return a `stbi_uc` instead, which would avoid the cast to unsigned char.

By the way, Sean and Fabian, please feel free to @ me any time - I'm always happy to help with things!

diamante0018:
You are right! I changed the return type of stbi__skip_jpeg_junk_at_end and removed the cast entirely.
Thanks.

NeilBickford-NV (approved):
The changes look good to me!

Ah, I don't have permission to merge anything in, though, and am not a project member, sorry - that would be @nothings and @rygorous . I understand it usually takes some time for things to be merged in - what I've usually done in the past for projects that require a patched version of `stb_image` is I've pointed Git submodules to the fork used for the merge request (or patched the copy if the project is using a copy of the file), then pointed them back to `nothings:master` once the patch has been merged.

diamante0018:
Thanks for the tip, I will do that I think. Or just suppress the warning before including the headers.

ChrisThrasher:
I'd like to bump this PR since I'm dealing with the same problems in #1444. @nothings Is this good to merge?

diamante0018:
Woot

rygorous:
A fix for this is already in the dev branch (and has been since May 2nd, commit https://github.com/nothings/stb/commit/8c3aa0548766e6fef7da184107d35fd81315c1a5), it's just waiting on the next time we do a release.

diamante0018:
Better late then never. May I suggest that some MSVC tests are added to STB workflow(s) to see if it compiles with all warnings enabled.
When I tried to recently compile stb (a week ago or so) it still failed.

PR #1447, opened 2023-02-08
stb_ds: Support environments without realloc

digicannon:
For environments that have `malloc` and `free` but not `realloc`.  `STBDS_REALLOC` can be implemented with something like this thanks to the addition of `size_in_use`:
```
void * realloc(void * old_ptr, size_t new_size, size_t size_in_use) {
    void * new_ptr = malloc(new_size);
    memcpy(new_ptr, old_ptr, size_in_use);
    return new_ptr;
}
```

Also made the arguments for STBDS_REALLOC/FREE match in the definition and documentation.

### DIFF
--- a/stb_ds.h
+++ b/stb_ds.h
@@ -41,8 +41,8 @@ COMPILE-TIME OPTIONS
      hash table insertion about 20% slower on 4- and 8-byte keys, 5% slower on
      64-byte keys, and 10% slower on 256-byte keys on my test computer.
 
-  #define STBDS_REALLOC(context,ptr,size) better_realloc
-  #define STBDS_FREE(context,ptr)         better_free
+  #define STBDS_REALLOC(context,ptr,new_size,size_in_use) better_realloc
+  #define STBDS_FREE(context,ptr)                         better_free
 
      These defines only need to be set in the file containing #define STB_DS_IMPLEMENTATION.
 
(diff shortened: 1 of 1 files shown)


rygorous:
This is not going to get merged.

Niche platforms and environments are something we can't test reliably and already take up a disproportionate amount of our time. If anything, we're going to support fewer affordances for platforms lacking basic C functionality going forward, not add more of them.

If your platforms' allocator has neither realloc nor a way to figure out the size of an allocation gives it base pointer, there is already an easy way to support it using the existing API: just put a tag value with the size in front in your `realloc` implementation. That way, the code that deals with the peculiarities of your platform lives on your side where you can test and debug it. By contrast, if we merged this, 99.9% of stb_ds users (ourselves included) would be on a platform that ignores `size_in_use` which means it's only a matter of time until it breaks.

nothings:
Haha, no, I said by email I'd be ok with this, although it needs to be changed to not break existing code.

nothings:
Although the only reason I said it would be ok to add it is because I didn't realize there was already an STBDS_REALLOC etc., so I thought size_in_use would be ok to support if we were adding the feature, but I suppose if there already is one I need to reconsider whether it's worth it.

nothings:
> If your platforms' allocator has neither realloc nor a way to figure out the size of an allocation gives it base pointer, there is already an easy way to support it using the existing API: just put a tag value with the size in front in your `realloc` implementation. That way, the code that deals with the peculiarities of your platform lives on your side where you can test and debug it. By contrast, if we merged this, 99.9% of stb_ds users (ourselves included) would be on a platform that ignores `size_in_use` which means it's only a matter of time until it breaks.

I guess this is a good argument, so yeah, nevermind, not going to accept this.

PR #1450, opened 2023-02-09
Fix reading bmp V3 info headers with `BI_BITFIELDS` compression

Cirras:
Previously these bitmaps would unconditionally discard all 4 bitmasks, then erroneously overread another 12 bytes if the compression was `BI_BITFIELDS`.

See attached BMP with `BITMAPV3INFOHEADER` and `BI_BITFIELDS` compression.
It should be loaded properly by paint, and cause a "bad BMP" error in stb_image.
[v3_info_header_with_bitfields.zip](https://github.com/nothings/stb/files/10698337/v3_info_header_with_bitfields.zip)

### DIFF
--- a/stb_image.h
+++ b/stb_image.h
@@ -5478,13 +5478,8 @@ static void *stbi__bmp_parse_header(stbi__context *s, stbi__bmp_data *info)
       stbi__get32le(s); // discard vres
       stbi__get32le(s); // discard colorsused
       stbi__get32le(s); // discard max important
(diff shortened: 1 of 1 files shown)


Cirras:
Totally understand that the maintainers have their own priorities, but I'd really appreciate it if this one could be looked at and hopefully merged.
It would allow me to remove some gnarly hacks in a LibGDX application that (indirectly) relies on sbt for loading bitmaps.

Issue #1452, opened 2023-02-21
NULL pointer dereference in the stb_image.h


0xdd96:
I discovered a [null pointer dereference](https://github.com/libsixel/libsixel/issues/73) vulnerability in the [stb_image.h](https://github.com/libsixel/libsixel/blob/490ec15087e37d8e1395e4dbfb99fc543c5bae5d/src/stb_image.h#L6446-L6458) of libsixel, which also exists in this project. 

Specifically, if the `stbi__pic_load_core` function returns 0 (line 6528), `result` will be released (line 6529) and set to 0 (line 6530). This null pointer will be dereferenced in `stbi__convert_format`, which would crash the application.

https://github.com/nothings/stb/blob/5736b15f7ea0ffb08dd38af21067c314d6a3aae9/stb_image.h#L6523-L6535

NeilBickford-NV:
Tracking CVE numbers: this is https://nvd.nist.gov/vuln/detail/CVE-2023-43898 ((#1521), and its equivalent in Servo's Rust port is https://rustsec.org/advisories/RUSTSEC-2023-0021.html.

PR #1454, opened 2023-02-25
Fixes null pointer dereference in https://github.com/nothings/stb/issues/1452

NBickford-NV:
Hi stb maintainers!

I just saw issue #1452, and put together this pull request to fix it. When stbi__pic_load_core returns `NULL`, this code now frees the allocated image and returns 0 immediately, instead of passing a null pointer to `stbi__convert_format()`.

Thanks!

### DIFF
--- a/stb_image.h
+++ b/stb_image.h
@@ -6527,7 +6527,7 @@ static void *stbi__pic_load(stbi__context *s,int *px,int *py,int *comp,int req_c
 
    if (!stbi__pic_load_core(s,x,y,comp, result)) {
       STBI_FREE(result);
-      result=0;
+      return 0;
    }
    *px = x;
    *py = y;


musicinmybrain:
Thank you! I have [just applied this](https://src.fedoraproject.org/rpms/stb/c/b0c5c6bad2626e09418c21252a87707fd08d3481?branch=rawhide) as a downstream patch in Fedora Linux.

NeilBickford-NV:
That's amazingly fast, thank you Ben! (I figured I'd send you an email later this morning since I remember you saw and applied the earlier stb_image patches, but you beat me to it!)

NeilBickford-NV:
Tracking CVE numbers: this is a patch for https://nvd.nist.gov/vuln/detail/CVE-2023-43898 ((#1521).

NBickford-NV:
Looks like this is also a patch for https://nvd.nist.gov/vuln/detail/CVE-2021-45340, which is how this bug appeared in libsixel (see https://github.com/libsixel/libsixel/issues/73 and https://github.com/nothings/stb/pull/1736).

musicinmybrain:
@NBickford-NV You’ve studied this code more deeply than I am likely to have time to. Do you think https://github.com/nothings/stb/pull/1736 is a wise change *in addition to* this PR? Is it easy enough to prove that there is nowhere else that `stbi__convert_format` could end up being called with a null `data` pointer?

NBickford-NV:
@musicinmybrain Turns out, that's a really good question! I tried proving this and failed at `stbi__load_gif_main`. #1736 also protects against that. So while #1454 fixes one bug, #1736 is a more general solution.

My attempted proof that this is the only place where `stbi__convert_format` could be called with a null `data` pointer is below. This fails in case 6.
* `stbi__convert_format` is called in [] places:
  1. In `stbi__do_png`, can `(unsigned char *)result` be `NULL`? If it's `NULL`, then `p->out` is `NULL`. That means `z->out` in `stbi__parse_png()` was `NULL` and `stbi__parse_png()` returned a non-zero value. There's a few things that could write to `z->out`:
    a. `stbi__create_png_image`; if it left `a->out` `NULL`, then it failed (so `stbi__parse_png()` returns 0) -- so this can't happen.
    b. `stbi__expand_png_palette`; this only sets `a->out` if it fails (so `stbi__parse_png()` returns 0) -- so this can't happen.
    c. So `result` can't be `NULL`.
  2. In `stbi__bmp_load`, can `out` be `NULL`? Control only reaches `stbi__convert_format` there if it doesn't exit at `if (!out) return stbi__errpuc("outofmem", "Out of memory");`, so this can't happen.
  3. In  `stbi__tga_load`, `tga_data` can't be null due to a null check after allocation, like in `stbi__bmp_load`.
  4. In `stbi__psd_load`, `out` can't be null due to a null check after allocation, like in `stbi__bmp_load`.
  5. In `stbi__pic_load`, it used to be that `result` could be 0 if `stbi__pic_load_core` failed; this PR patches that. Since `stbi__pic_load_core` doesn't modify the `result` pointer itself (although it isn't passed in as an `stbi_uc* const`), `result` can only be null if it was null before `stbi__pic_load_core`. That can't happen because of the null check after the call to `stbi__malloc_mad3`.
  6. In `stbi__load_gif_main`, can `out` be null? This might be possible; we have to pass `stbi__gif_test`, but then have `stbi__gif_load_next` return either `0` (error) or `s` (end of animated gif). That's probably possible to do; make a file that starts with `GIF8[79]a` and then has a corrupt first frame.
  7. In `stbi__gif_load`, `u` can't be null because control's guarded by an `if(u)` check.
  8. In `stbi__pnm_load`, `out` can't be null due to a null check after allocation, like in `stbi__bmp_load`.

NBickford-NV:
Okay, after deeper analysis, I think case 6 might be OK after all, but it's complex.

`stbi__convert_format` only crashes if `out` is null and both the width and height are greater than 0; if either are 0, it never reads from the source pointer. In `stbi__load_gif_main`, the width is `layers * g.w`, but `layers` is only nonzero if `u` is non-null at least once. And if `u` is non-null at least once, then control reaches `out = (stbi_uc*)stbi__malloc( layers * stride );` and then the `if (!out)` check.

That is: if `out` is null, then the width must be 0, and so there's no crash. If the width were 0, then `out` would be a non-null pointer.

musicinmybrain:
Thank you for that very detailed analysis! I’m shipping https://github.com/nothings/stb/pull/1736 as an additional downstream patch in Fedora’s [`stb` package](https://src.fedoraproject.org/rpms/stb) and referencing this discussion.

hzeller:
Given the analysis, I wonder how we get @nothings attention to get both this #1454 and #1736 merged to fix a whole class of bugs.

They are also sufficiently trivial that the review is simple.

nothings:
> Given the analysis, I wonder how we get @nothings attention

I read every email I get, and I get an email for every comment in the repo. It's just a matter of me committing the time to a release cycle.

PR #1457, opened 2023-03-10
Fix: stb_sprintf - gcc defines __riscv

misaka00251:
This allow stb to build on RISC-V machine.


hack3ric:
This is actually wrong - `__riscv` doesn't mean riscv64, it should also check `__riscv_xlen == 64`. I'll submit an updated pull request.

PR #1462, opened 2023-03-23
Update stb_rect_pack.h Change the type of stbrp_rect.id from int to void* for more flexible api

Anaskorichi999:
Update stb_rect_pack.h Change the type of stbrp_rect.id from int to void* for more flexible api


nothings:
this would break anyone using the current id. you can make this change locally but it doesn't make sense to accept.

Anaskorichi999:
What do you think of adding this
struct stbrp_rect {
    stbrp_coord x,y;
    int id,w,h,was_packed;
    void *userdata;
};

nothings:
It just doesn't seem very necessary? You can make a parallel array of void*, or make an array of void* indexed by id if you need that functionality.

Issue #1458, opened 2023-03-11
stb_image: unused functions warning when using `STBI_ONLY_PNG`


rajveermalviya:
**Describe the bug**

when using `STBI_ONLY_PNG` and `cc -Wall`, compiler generates following warnings:
```shell
stb_image.h:1068:12: warning: unused function 'stbi__addints_valid' [-Wunused-function]
static int stbi__addints_valid(int a, int b)
           ^
stb_image.h:1076:12: warning: unused function 'stbi__mul2shorts_valid' [-Wunused-function]
static int stbi__mul2shorts_valid(short a, short b)
           ^
2 warnings generated.
```
https://godbolt.org/z/o535xjnWs

**To Reproduce**

```c
#define STB_IMAGE_IMPLEMENTATION
#define STBI_ONLY_PNG
#include "stb_image.h"
```

**Compiler Info**

```shell
$ clang --version
clang version 15.0.7 (Fedora 15.0.7-1.fc37)
Target: x86_64-redhat-linux-gnu
Thread model: posix
InstalledDir: /usr/bin
```

rygorous:
These are not going to get fixed.

The `STBI_NO_*` and `STBI_ONLY_*` defines are a historical mistake that we still keep in the library purely to not break existing users, but there are 9 binary flags here that work out to 512 possible combinations (511 if you exclude the nonsensical "support no formats" configuration), and that is simply too much of a maintenance burden to keep compiling warning-free across multiple versions of the 3 major compilers we support, especially considering how low-value the feature is.

Individually each such flag combination works out to a few `#if` directives, but cumulatively it leads to all shared code looking like https://github.com/nothings/stb/blob/5736b15f7ea0ffb08dd38af21067c314d6a3aae9/stb_image.h#L1020 which is just a brittle nightmare. Unused static functions might elicit warnings but they are completely harmless. Copy-pasting these `#if`s around everywhere hurts readability and maintenance, and getting them wrong in the other direction and excluding code that shouldn't be (and what should and shouldn't be excluded can easily change for only some combinations of flags e.g. after a security patch) will cause the library to not compile, which is a much more serious issue.

[...]

PR #1472, opened 2023-04-20
This prevents few warnings that appear when compiling with clang

oviano:
Examples of these warnings:

```
stb_image_write.h:776:13: error: 'sprintf' is deprecated: This function is provided for compatibility reasons only. Due to security
      concerns inherent in the design of sprintf(3), it is highly recommended that you use snprintf(3) instead. [-Werror,-Wdeprecated-declarations]
      len = sprintf(buffer, "EXPOSURE=          1.0000000000000\n\n-Y %d +X %d\n", y, x);
```

```
./thirdparty/stb/stb_image_write.h:518:32: error: missing field 'context' initializer [-Werror,-Wmissing-field-initializers]
   stbi__write_context s = { 0 };
                               ^
```

### DIFF
--- a/stb_image_write.h
+++ b/stb_image_write.h
@@ -206,7 +206,9 @@ STBIWDEF void stbi_flip_vertically_on_write(int flip_boolean);
    #define _CRT_NONSTDC_NO_DEPRECATE
    #endif
 #elif defined(__clang__)
+   #pragma clang diagnostic push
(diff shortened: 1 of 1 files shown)


nothings:
We generally do not like to put pragma warning directives into the libraries. See https://github.com/nothings/stb/issues/1458#issuecomment-1465084305 for some explanations.

PR #1476, opened 2023-05-16
stb_image: pbm support

plzombie:
stb_image already supports binary ppm and pgm files. So I think it will be good to add support for pbm too.
I also added some test files and added pbm in *tests/ossfuzz.sh*, don't know if it's enough or not.


plzombie:
The last commit https://github.com/nothings/stb/pull/1476/commits/1ab7418d80b72c2587e64cdce767dde56bd62ca7 is for 16-bit ppm files. As I understand it's always stored in big-endian format so conversion is needed

PR #1477, opened 2023-05-16
Add STBSP__ASAN for msvc compiler

septag:


Condzi:
I also had an issue with ASAN and the trick stbsp uses for finding '\0'. However, I found out that I also need to modify STBSP_PUBLICDEC / PUBLICDEF macros by adding STBSP__ASAN to them, like so:
```cpp
#ifdef STB_SPRINTF_STATIC
#define STBSP__PUBLICDEC static STBSP__ASAN
#define STBSP__PUBLICDEF static STBSP__ASAN
#else
#ifdef __cplusplus
#define STBSP__PUBLICDEC extern "C" STBSP__ASAN
#define STBSP__PUBLICDEF extern "C" STBSP__ASAN
#else
#define STBSP__PUBLICDEC extern STBSP__ASAN
#define STBSP__PUBLICDEF STBSP__ASAN
#endif
#endif
```
I'm not sure why they aren't there in the first place. Without this added, my code still was giving me `global-buffer-overflow`.

Manuzor:
I came here because I ran into `global-buffer-overflow` as well, even with the fixes in this PR, but the final missing piece was adding `STBSP__ASAN` to the public definitions, so thanks @Condzi for that info!

By the way, I think this PR is a duplicate of #1350 although it contains an explicit MSVC version check.

PR #1478, opened 2023-05-26
Fix an issue with out of range value being generated

danemadsen:
Fixes raysan5/raylib#3071

This wan an issue caused by stb_perlin_fbm_noise3 where values outside of the range of -1 to 1 were occasionally generated causing underflow / overflow which resulted in spikes


nothings:
Occasionally generating a value outside of the range of [-1,1] is the expected behavior. Normalizing as you suggest would radically change the output for all users. Rescaling to a guaranteed range is best done in the caller.

PR #1497, opened 2023-07-13
Fix memory leaks if stbi__convert failed

JarLob:


NeilBickford-NV:
Elaborating on the pull request title in case it helps make review easier: it looks like this caused a memory leak because `stbi__convert_*` was called like this:

```c
result = stbi__convert_8_to_16((stbi_uc *) result, *x, *y, req_comp == 0 ? *comp : req_comp);
```

`stbi__convert_8_to_16` or `stbi__convert_16_to_8` destroy the input and return a new buffer - but if allocation fails, they report an error and return 0. So, in that failure case, `result` is allocated (by `stbi__load_*`), but never freed.

This might also fix a segfault: if `stbi__vertically_flip_on_load` is `true` and `stbi__convert_*` fails, then the null `result` would be passed to `stbi__vertical_flip`, which tries to dereference it (if `*y > 1`).

PR #1498, opened 2023-07-17
Fix memory leaks in stb_vorbis

JarLob:


sezero:
I think these issues are already handled in https://github.com/nothings/stb/pull/1174

PR #1515, opened 2023-09-14
Undefine L C R defines after use

danielmaciel:
Hi, just some minor #undef additions to avoid conflicts.


nothings:
You're not supposed to #include stb_vorbis.c, it's a C file and not an H file, so there can't be any conflicts.

PR #1517, opened 2023-09-23
Fix stbsp__uintptr on riscv64

hack3ric:
[Last pull request](https://github.com/nothings/stb/pull/1457) had done it wrong. This is an updated patch.


hack3ric:
Hi, could this change be accepted?

Issue #1602, opened 2024-02-02
stb_image_resize2: assert/crash on resizing 2x2 to 1x1 images with wrap


aras-p:
**Describe the bug**
When resizing very small images (like 2x2 into 1x1) with `STBIR_EDGE_WRAP` and `STBIR_FILTER_DEFAULT`, there's a debug assert and a possible later crash.

stb_image_resize2.h version v2.04.

**To Reproduce**

This code:
```
#include <stdint.h>
#define STB_IMAGE_RESIZE_IMPLEMENTATION
#include "stb_image_resize2.h"

int main()
{
	int srcWidth = 2, srcHeight = 2, dstWidth = 1, dstHeight = 1;
	uint8_t* src = new uint8_t[srcWidth * srcHeight * 4];
	memset(src, 0xAB, srcWidth * srcHeight * 4);
	uint8_t* dst = new uint8_t[dstWidth * dstHeight * 4];
	stbir_resize(src, srcWidth, srcHeight, 0,
		dst, dstWidth, dstHeight, 0,
		STBIR_RGBA, STBIR_TYPE_UINT8,
		STBIR_EDGE_WRAP, STBIR_FILTER_DEFAULT);
	delete[] src;
	delete[] dst;
}
```

Triggers `STBIR_ASSERT( info->ring_buffer_num_entries <= info->alloc_ring_buffer_num_entries );` at line 7028. `info->ring_buffer_num_entries` is 6, `info->alloc_ring_buffer_num_entries` is 5.

Changing wrap mode to clamp, or filter to box does not trigger the assert/crash.

**Expected behavior**
I'd expect no asserts and no crashes.

jeffrbig2:
I will scope this out, this weekend!

jeffatrad:
Also, I should probably eventually special case 16x16 and smaller downsamples - that could probably be quite a bit faster without going through all the general machinery.  Although these are already fast, I suppose.

nothings:
I wouldn't bother special-casing them; it's hard to imagine a scenario where they'd be a critical path if they're already reasonably fast.

jeffrbig2:
OK, fixed in #1603

nothings:
Fixed by #1605.

PR #1603, opened 2024-02-04
Fix 2 pixel to 1 pixel with wrap and fix output buffer for output callback

jeffrbig2:


nothings:
Fixes #1602

nothings:
Can you explain why "no double reflect" is testing `> w*3` rather than `> w*2`, when the latter seems to make more sense?

jeffrbig2:
We allow you to reflect 1x off the left, then 1x for the pixel buffer of course, and then 1x on the right.    2x2 with some filters could get off the 1x on the left, so that you would have have wrap a 2nd time (which I just don't support)..
 
  -Jeff

jeffrbig2:
Can walk you through on zoom if helpful...
 
  -Jeff

jeffrbig2:
I guess that's a shitty comment...
 
  -Jeff

PR #1605, opened 2024-02-04
Fix 2 pixel to 1 pixel with wrap and fix output buffer for output callback

jeffrbig2:


jeffrbig2:
Wait, I'm updating the comments...

nothings:
Fixes #1602.

nothings:
Just to be clear, you changed the one function to take `row0` and `row1`, but the only thing you do with `row0` is an assert?

jeffrbig2:
Yes. Eventually, I may use that parameter for other clipping checks...

nothings:
Should it be in an `STBIR__UNUSED()`?

jeffrbig2:
Oh gad, in release mode, you'd get that dumb warning?

nothings:
I haven't seen it, but I assume you'd get it on some compiler with some settings (which we don't control since it's compiled as source with the user's settings).

PR #1606, opened 2024-02-05
Fix "string_len" typo in stb_c_lexer.h

lackhoa:
Looks like a copy-paste mistake.


NBickford-NV:
Hi! If I'm looking at the Files Changed view correctly, it looks like this is already in main as a result of merging in https://github.com/nothings/stb/pull/1147/files ?

lackhoa:
@NBickford-NV Yeah, it's ben fixed. Thanks!

PR #1610, opened 2024-02-14
Fix compile error on riscv64

rwmjones:
Adding @davidlt who wrote the patch originally

We are testing the patch on RISC-V here:
http://fedora.riscv.rocks/koji/packageinfo?packageID=28035

### DIFF
--- a/stb_sprintf.h
+++ b/stb_sprintf.h
@@ -230,7 +230,7 @@ STBSP__PUBLICDEC void STB_SPRINTF_DECORATE(set_separators)(char comma, char peri
 #define stbsp__uint16 unsigned short
 
 #ifndef stbsp__uintptr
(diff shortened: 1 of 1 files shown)


ilyakurdyukov:
This is insane. Whitelisting 64-bit architectures in 2024 should be considered a serious crime against portability. In times when most architectures are 64-bit. How many years have compilers had predefined macros like these?

```
#define _LP64 1
#define __LP64__ 1
#define __INTPTR_WIDTH__ 64
```

And this whitelisting nonsense only for the uintptr, which has been available in `stdint.h` since C99, for 25 years!

And yet, people who can't find their architecture on the list ask to add it. They don't ask to fix this trashfire once and for all.

I wanted to create a bug about this, but judging by the number of open and unanswered issues and PRs, it looks like this project is closer to abandonware than to anything else.

PR #1615, opened 2024-02-20
Make fallthroughs explicit in stb_image.h

r-barnes:
The `-Wimplicit-fallthrough` flag can be used to check for implicit fallthroughs, which are a common source of errors.

I've added and used `STBI_FALLTHROUGH` which maps to `__attribute__((fallthrough))`, which the compiler uses to determine whether a fallthrough is explicit or not.

### DIFF
--- a/stb_image.h
+++ b/stb_image.h
@@ -372,7 +372,14 @@ RECENT REVISION HISTORY:
 
 #define STBI_VERSION 1
 
-#define STBI_FALLTHROUGH __attribute__((fallthrough))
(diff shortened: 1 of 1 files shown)


nothings:
This needs to work on all compilers, so the define needs to be an ifdef which tests for suitable compilers, with an else that definites it to nothing.

iTrooz:
Hey ! Is there any news on this PR ?

nothings:
I quit my job this week and should soon have plenty of free time to start updating the repo more regularly.

Issue #1446, opened 2023-02-03
stb_image_write: warning C4996: ... Consider using sprintf_s ...


etavardt:
During a MS VS C++ build, I get the following warning.

    stb_image_write.h(776): warning C4996: 'sprintf': This function or variable may be unsafe. 
    Consider using sprintf_s instead. To disable deprecation, use _CRT_SECURE_NO_WARNINGS.
    See online help for details.

**Expected behavior**

    I expect to see warnings that are relevant to my code without adding _CRT_SECURE_NO_WARNINGS to my build. 
    If I did add it I would not see Warnings for my own code as well and I do wish to see them.

etavardt:
My work around is to place:
#define __STDC_LIB_EXT1__ 
along with:
#define STB_IMAGE_WRITE_IMPLEMENTATION
just before the:
#include <stb/stb_image_write.h>

I don't know if this is a good idea or not but it got rid of the warning.  I'll know better once I start implementing my save function.

FYI: keep up the good work.

N-R-K:
> I expect to see warnings that are relevant to my code without adding _CRT_SECURE_NO_WARNINGS to my build. 
> If I did add it I would not see Warnings for my own code as well and I do wish to see them.

`stb` libraries try to compile without warnings - so this is not a "solution" but as a temporary workaround - it's _typically_ possible to silence warnings for a section of code via some sort of pragma. GCC and Clang accept the following:

```c
#pragma GCC diagnostic push /* also works on clang */
#pragma GCC diagnostic ignored "-Wunused-function"
#define HEADERLIB_IMPLEMENTATION
#include "headerlib.h"
#pragma GCC diagnostic pop
```

I'd assume msvc would have similar construct - but I don't use msvc so cannot help you further.

- - -

> ` #define __STDC_LIB_EXT1__`

AFAIU [that macro](https://port70.net/~nsz/c/c11/n1570.html#6.10.8.3) should be defined by the implementation (i.e libc) if they support Annex K. I don't believe it's something the user is meant to define themselves (?)

etavardt:
You make a good point about not using  `#define __STDC_LIB_EXT1__`

So I went with this:

    #define STB_IMAGE_WRITE_IMPLEMENTATION
    #pragma warning( push )
    #pragma warning(disable:4996)
    #include <stb/stb_image_write.h>
    #pragma warning( pop )

[...]

PR #1618, opened 2024-03-04
Fix _CRT_SECURE_NO_WARNINGS deprecation notice when compiling with msvc

birds3345:
This fix takes care of the _CRT_SECURE_NO_WARNINGS deprecation notice when compiling with msvc. The preprocessor directive did not account for msvc compilers as they don't define STDC_LIB_EXT1.


birds3345:
Forgot to mention that this is in stb_image_write.h

birds3345:
fixes #1446

PR #1619, opened 2024-03-04
FIX: use snprintf instead of sprintf

ismagilli:
Modern compilers warn that the `sprintf` function is deprecated. I have added a call to `snprintf` function.

Note. Function `snprintf` appeared only in C99, so I added a check that the program compiles to the C99 standard and above. The `gcc` and `clang` compilers define the constant `__STDC_VERSION__`. The `defined(__STDC_VERSION__)` check is necessary because the `__STDC_VERSION__ ` constant may not defined. 

From `man snprintf`:
```
the snprintf() and vsnprintf() functions conform to ISO/IEC 9899:1999 (“ISO C99”)
```


ismagilli:
Update commit text:
```diff
- FIX: use snprintf instead of sprintf
+ FIX: image_writer: use snprintf instead of sprintf
```

PR #1622, opened 2024-03-15
Avoid address sanitizer on MSVC

tim-rex:
Extending on #393
Since MSVC now supports address sanitizer, it appears to be throwing the same issues as other platforms

This PR adds compiler checks and appropriate declspecs for MSVC

Note that _declspec for MSVC must appear **before** the storage specifier, hence the additional set of STBSP__PUBLICDEF's
There's probably a cleaner way to structure that


Manuzor:
See also #1477 as well as #1350.

Condzi:
You might add UBSAN as well, since it also flags the tricks used in sprintf library. `__attribute__((no_sanitize("undefined")))`

PR #1623, opened 2024-03-15
support memory context for arrays

hugoArregui:


nothings:
hash table had a context pointer because the hash table already has a lot of size overhead. however, the array size overhead is a lot smaller, so adding a context pointer to it adds a significant amount.

hugoArregui:
@nothings would this makes sense if I add a `#ifdef STBDS_ARRAY_CUSTOM_CONTEXT` and only include the pointer in the structure then? If you have another idea I can try it as well (I though this may be useful for you as well because of the TODO comment in the file)

nothings:
Oh, I forgot about the todo. I guess it's ok then.

PR #1638, opened 2024-05-01
Fix warning on use of sprintf

rasky:
Using snprintf avoid compilation warnings/errors on modern UNIX systems. Specifically, this triggered on macOS 14.4.1.


rasky:
Duplicate of #1619

PR #1647, opened 2024-06-06
Fix RGB->BGR channel flipping. First 4 pixels were being skipped.

ryanrsrs:
Caused first 4 pixels of every line to have bad colors. Not noticable unless the image has bright colors along the left edge.

### DIFF
--- a/stb_image_resize2.h
+++ b/stb_image_resize2.h
@@ -326,6 +326,7 @@
       Jorge L Rodriguez: Original 1.0 implementation
       Aras Pranckevicius: bugfixes
       Nathan Reed: warning fixes for 1.0
+      Ryan Salsbury: bugfixes
 
    REVISIONS
       2.07 (2024-05-24) fix for slow final split during threaded conversions of very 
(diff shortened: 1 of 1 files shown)


jeffrbig2:
Gad, the scalar paths of the channel flipping weren't being tested.  Sorry, will update.  What platform were you on that the SIMD path wasn't selected, btw...?

ryanrsrs:
"armv7l" i.e. 32-bit ARM Raspberry Pi Zero 2 W. 

Code was compiled with CFLAGS including: -march=native -Os -DSTBIR_USE_FMA -mfpu=neon-vfpv4 -mfp16-format=ieee

Here is my makefile: https://github.com/ryanrsrs/console-jpeg/blob/main/Makefile

I tried with and without -DSTBIR_NO_SIMD, but I think this scalar path was always being used. Should it be NEON?

jeffrbig2:
Oh right, that path has no simd path now.  Let me do your fix and add one...

PR #1667, opened 2024-07-24
stb_image: support horizontally flipped TGA images

illwieckz:
The TGA format is a bottom-left format by default: the first byte of the first column is expected to be displayed at the bottom-left of the screen, this behavior is inherited from devices painting the screen from the bottom left to the top right.

The TGA format provides two bits in the image descriptor byte to paint the data from other sides. The 4th bit tells the device to paint the screen from right-to left and the 5th bit tells the device to paint the screen from top to bottom.

So basically:

- 00: from bottom-left to top-right
- 01: from top-left to bottom-right
- 10: from bottom-right to top-left
- 11: from top-right to bottom-left

Previously stb_image only read the 5th bit and then only supported the loading of vertically flipped images, stb_image was ignoring the 4th bit coding the horizontal flip. Now both flipping directions are supported for both raw and RLE storage.

### DIFF
--- a/stb_image.h
+++ b/stb_image.h
@@ -5884,8 +5884,8 @@ static void *stbi__tga_load(stbi__context *s, int *x, int *y, int *comp, int req
    int tga_height = stbi__get16le(s);
    int tga_bits_per_pixel = stbi__get8(s);
    int tga_comp, tga_rgb16=0;
-   int tga_inverted = stbi__get8(s);
-   // int tga_alpha_bits = tga_inverted & 15; // the 4 lowest bits - unused (useless?)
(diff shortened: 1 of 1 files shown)


nothings:
Given that it sounds like this literally never occurs in the wild, it would be better to make the change as small as possible. Is it possible to use the second loop which swaps the pixels for all cases, instead of having two versions?

Ernesto-Alves67:
Help me understand. Why are so many pull request? Missing people to actually aprove them? I'm asking because i want to colaborate with some big and relevant project. And that one interrested me.

PR #1682, opened 2024-09-10
stb_ds: Fix `d[i] << 24` undefined behaviour due to overflow

darkuranium:
Because of int promotion, `d[i] << 24` overflows an int, resulting in UB.

Found via ubsan in clang 18:
```
stb/stb_ds.h:1124:67: runtime error: left shift of 255 by 24 places cannot be represented in type 'int'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior stb/stb_ds.h:1124:67
```

### DIFF
--- a/stb_ds.h
+++ b/stb_ds.h
@@ -1079,8 +1079,8 @@ static size_t stbds_siphash_bytes(void *p, size_t len, size_t seed)
     } while (0)
 
   for (i=0; i+sizeof(size_t) <= len; i += sizeof(size_t), d += sizeof(size_t)) {
-    data = d[0] | (d[1] << 8) | (d[2] << 16) | (d[3] << 24);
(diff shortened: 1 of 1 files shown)


sleeptightAnsiC:
same as https://github.com/nothings/stb/pull/1188 (?)

IhrWerbepartner:
Ran into this today as well.

nothings:
> Ran into this today as well.

Do you say "ran into", do mean you encountered incorrect behavior, or do you mean it tripped ubsan?

IhrWerbepartner:
I compiled with `zig cc`, and it tripped ubsan. But i managed to reproduce it with gcc and ubsan enabled as well. I have not investigated if this caused incorrect behavior though.

nothings:
Yes, if it triggered ubsan for one person it is not noteworthy it triggers for others, and it is not something that needed confirmation or narrowing down, I just wanted to verify there was no actual buggy behavior, just an unexploited UB.

PR #1684, opened 2024-09-19
seemingly pointless if branch

LemoNode:
I cant see what this if branch is supposed to do, the j gets reset to zero in the for loop below, the output buffer gets memset to zero right below as well and nothing, other than the output buffer that gets set to 0 right afterwards anyway,  is changed in the hex_blamp function.
The prev_dt isnt really doing anything either, maybe im missing someting but i deleted this if statement and everything seems to be working just fine for me, would love to know if this is leftover code or if this is just incomplete

### DIFF
--- a/stb_hexwave.h
+++ b/stb_hexwave.h
@@ -433,19 +433,6 @@ STB_HEXWAVE_DEF void hexwave_generate_samples(float *output, int num_samples, He
 
    // convert parameters to times and slopes
    hexwave_generate_linesegs(vert, hex, dt);
-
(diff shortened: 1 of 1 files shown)


nothings:
What do you think `// if frequency changes` means?

LemoNode:
well yes but what is it actually modifying though?

nothings:
Yeah, you're right, the output memset should be before the if.

LemoNode:
Oh I see, i'll just delete the request anyway since the problem was something else, thank you!

PR #1699, opened 2024-10-06
fix: stbi_write_hdr_to_func undefined when STBI_WRITE_NO_STDIO defined

craftablescience:
The implementation of `stbi_write_hdr_to_func` is not compiled when `STBI_WRITE_NO_STDIO` is defined, even though the function signature is still present in the header. This appears to be a typo/oversight, and I have changed the location of the #ifndef guard so it matches the other image formats. I have tested this code and the header now compiles with `STBI_WRITE_NO_STDIO` while using `stbi_write_hdr_to_func` where it did not before.


craftablescience:
Fixed the spacing to match code style

PR #1707, opened 2024-10-17
stb_image: Define JPEG-specific functions only when JPEG is needed

### LINKED ISSUE #1648: sttb_image: Unused functions when STBI_NO_JPEG is defined
When defining `STBI_NO_JPEG`, the `GCC` compiler starts to show warnings:

`
stb_image.h:1078:12: warning: 'int stbi__mul2shorts_valid(int, int)' defined but not used [-Wunused-function]
 1078 | static int stbi__mul2shorts_valid(int a, int b)
stb_image.h:1070:12: warning: 'int stbi__addints_valid(int, int)' defined but not used [-Wunused-function]
 1070 | static int stbi__addints_valid(int a, int b)
`

So, these functions must be wrapped into `#ifndef STBI_NO_JPEG` (or something else).

ccatgirl:
There are two functions that are only used for JPEG reading.
If JPEG is not wanted (`STBI_NO_JPEG`) they are still defined and cause errors when compiling with `-Wall -Werror -pedantic`.
This pull request fixes that and also closes #1648.
Maybe it's better to put the code for those two functions into a different location?

### DIFF
--- a/stb_image.h
+++ b/stb_image.h
@@ -100,7 +100,7 @@ RECENT REVISION HISTORY:
  Bug & warning fixes
     Marc LeBlanc            David Woo          Guillaume George     Martins Mozeiko
     Christpher Lloyd        Jerry Jansson      Joseph Thomson       Blazej Dariusz Roszkowski
-    Phil Jordan                                Dave Moore           Roy Eltham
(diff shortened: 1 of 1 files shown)


ynks:
+1 also getting this warnings

PR #1708, opened 2024-10-18
Fix STBIR__FREE_AND_CLEAR causing null reference member access

xchellx:
`stb_image_resize2.h:6734:5: runtime error: member access within null pointer of type 'stbir__info' (aka 'struct stbir__info')`
ASAN detects this (where Valgrind didn't).

### DIFF
--- a/stb_image_resize2.h
+++ b/stb_image_resize2.h
@@ -6689,7 +6689,7 @@ static void stbir__get_split_info( stbir__per_split_info* split_info, int splits
 
 static void stbir__free_internal_mem( stbir__info *info )
 {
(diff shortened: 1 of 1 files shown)


nothings:
That's pretty subtle and seems to subvert whatever the copying to `p` is for. I think a better fix would be to replace the final `STBIR__FREE_AND_CLEAR(info)` with:

```
STBIR_FREE(info)
```
since there's no good reason to clear that pointer (it's a stack local and this is essentially the last line of the function).

nothings:
Just to be clear, does that fix the ASAN warning?

xchellx:
After a typo fix (forgot second parameter) it does indeed fix the ASAN error.

PR #1724, opened 2024-12-04
Fix Clang Warning

YGXXD:
image
Apple Clang suggests using the safer function snprintf instead of sprintf. And I tried to fix this warning.


nothings:
Proposed implementation isn't portable to all supported compilers/platforms.

YGXXD:
You're right, snprintf is the standard after c99

PR #1742, opened 2025-02-08
stb_vorbis.c: don't depend on limits.h, define UINT_MAX ourselves

Des-Nerger:
> * Do add your name to the list of contributors.

Not worth it.


Des-Nerger:
Oh, my bad. I didn't actually need this change: the Zig toolchain I use allows things like:
```c
#include <stdint.h>
#include <stddef.h>
#include <limits.h>
#include <stdbool.h>
```
without a need to enable `libc`. So I guess I'd better to close this PR and delete my corresponding branch.

PR #1748, opened 2025-02-18
stb_image: add support for CgBI version 0x06 PNG format

rwl4:
The CgBI format is Apple's variant of PNG used in early iOS versions, and now used in projects like touchHLE. While stb_image already supported version 0x02, version 0x06 requires different handling:
- Read CgBI chunk version (0x06 uses raw DEFLATE data without zlib headers)
- Add zlib wrapper (0x78 0x01 header + dummy Adler32) for version 0x06
- Handle color transformation based on version

This allows stb_image to load CgBI PNGs from older iOS games and apps that use version 0x06. Tested with multiple CgBI PNGs from early iOS applications.

### DIFF
--- a/stb_image.h
+++ b/stb_image.h
@@ -114,7 +114,7 @@ RECENT REVISION HISTORY:
     Josh Tobin              Neil Bickford      Matthew Gregan       github:poppolopoppo
     Julian Raschke          Gregory Mullen     Christian Floisand   github:darealshinji
     Baldur Karlsson         Kevin Schmidt      JR Smith             github:Michaelangel007
-                            Brad Weinberger    Matvey Cherevko      github:mosra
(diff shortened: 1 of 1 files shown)


nothings:
I'm sure this is of interesting to TouchHLE, but it doesn't seem like it has broader utility. Maybe it's better to just make this a fork?

rwl4:
My rationale was that stb_image has the `stbi_convert_iphone_png_to_rgb` feature half complete, so I figured it would make sense to complete it. It's fine as a fork too though. It's up to you.

nothings:
I don't love having the iphone support at all, it just feels clumsy and ineffective anyway, but it was added before we got more strict about the scope of the library and it didn't seem reasonable to remove it.

PR #1761, opened 2025-03-19
Fix typo.

wacopha:


Hex4dec:
so close, its actually Mitchell-Netravali

PR #1811, opened 2025-06-21
Realloc sized for stb_ds

jonathandw743:
STBDS_REALLOC_SIZED for stb_ds (similarly to stb_image).
Good for using different kinds of allocators like arenas.

### DIFF
--- a/stb_ds.h
+++ b/stb_ds.h
@@ -41,12 +41,14 @@ COMPILE-TIME OPTIONS
      hash table insertion about 20% slower on 4- and 8-byte keys, 5% slower on
      64-byte keys, and 10% slower on 256-byte keys on my test computer.
 
-  #define STBDS_REALLOC(context,ptr,size) better_realloc
-  #define STBDS_FREE(context,ptr)         better_free
+  #define STBDS_MALLOC(context,size)                       better_malloc
+  #define STBDS_FREE(context,ptr)                          better_free
+  #define STBDS_REALLOC(context,ptr,size)                  better_realloc
+  #define STBDS_REALLOC_SIZED(context,ptr,oldsize,newsize) better_realloc (which required old size)
 
      These defines only need to be set in the file containing #define STB_DS_IMPLEMENTATION.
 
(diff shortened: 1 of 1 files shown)


jonathandw743:
> I'm not opposed in principle, but why add STBDS_MALLOC?

It's not entirely necessary but:

Its how it is done in in stb_image and stb_image_write - parity between different files seems like a good idea.
Most allocators have separate alloc and realloc functions where realloc does "more stuff". Allowing the simpler alloc to be used where it can be is slightly more efficient.

One downside is slightly increased complexity of the code - one extra macro.
This is not much of an issue for the user as if they do happen to have an allocator with realloc but no alloc, they can just define STBDS_MALLOC as realloc(0, ...). If this is an issue, I could make defining STBDS_MALLOC optional and provide a default definition using STBDS_REALLOC_SIZED.

The macro I think is most redundant is in fact STBDS_REALLOC as if the user doesn't want to use the oldsz provided by STBDS_REALLOC_SIZED, they can just... not. This could be seen as more ergonomic for the user, but IMHO the user having to know that there are two macros and knowing which one is used if both are defined is more costly than the slightly cleaner code. I just included this for parity, but if it's not important then I can remove it.

nothings:
Right, removing STBDS_REALLOC is a non-starter because that breaks back-compatibility for a feature most people don't need (old size).

stb_ds was written long after stb_image and stb_image_write, and omitting MALLOC as being redundant was an intentional choice when it was written, not an oversight. given the traditional definition of realloc as being malloc when oldptr=NULL, having to wire up a malloc just seems like unnecessary friction. if anything, the no-malloc interface should be backported to stb_image et al. 

(You could also theoretically eliminate free, since Posix and ISO C both also define realloc as being free if size is 0, but there was some reason I didn't exploit this, although I don't remember now; mayybe either some other platforms it doesn't work, or just the edge case of freeing a NULL ptr being treated as malloc(0) not free(NULL).)

jonathandw743:
> having to wire up a malloc just seems like unnecessary friction

Would you like me to remove malloc entirely or provide a default definition?

PR #1813, opened 2025-06-26
Simplify stb_divide.h implementation and improve C compatibility

maxpesh:
- Simplify implementation allowing to remove unnecessary overflow/underflow checking.
- Fix bug in defining C_INTEGER_DIVISION_TRUNCATES when not explicitly defined and compiled with C99 or later.
- Change C++ // style comments to C /* */ style comments for compatibility with pre-C99 language standard.

### DIFF
--- a/stb_divide.h
+++ b/stb_divide.h
@@ -1,87 +1,88 @@
-// stb_divide.h - v0.94 - public domain - Sean Barrett, Feb 2010
-// Three kinds of divide/modulus of signed integers.
-//
-// HISTORY
-//
-//   v0.94              Fix integer overflow issues
-//   v0.93  2020-02-02  Write useful exit() value from main()
-//   v0.92  2019-02-25  Fix warning
-//   v0.91  2010-02-27  Fix euclidean division by INT_MIN for non-truncating C
-//                      Check result with 64-bit math to catch such cases
-//   v0.90  2010-02-24  First public release
-//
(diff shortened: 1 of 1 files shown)


nothings:
Are `//` comments something you actually need to avoid, or is it just on principle? Because all stb libraries use that, so it's probably not worth changing unless you have an actual scenario. (You might, that's why I'm asking!)

maxpesh:
`//` comments were introduced in C99, so compiling with a pre-C99 standard will result in a compilation error.
`gcc -x c stb_divide.h -DSTB_DIVIDE_IMPLEMENTATION -DSTB_DIVIDE_TEST -DC_INTEGER_DIVISION_TRUNCATES -DSTB_DIVIDE_TEST_64='long long' -DSTB_DIVIDE_TEST_FLOOR -std=c90 -O3`
Fails with error:
```
stb_divide.h:1:1: error: C++ style comments are not allowed in ISO C90
    1 | // stb_divide.h - v0.94 - public domain - Sean Barrett, Feb 2010
      | ^
```

nothings:
I'm well aware of that. But in the real world of de facto standards, many C compilers before C99 actually supported `//` comments.

So, again I ask, is this something you actually need, or just something you're changing on principle?

maxpesh:
No, I don't have any actual scenario other than the compilation error I had.

PR #1818, opened 2025-07-04
Add loading of older Silicon Graphics Iris image type

ScottJohnson2718:
Silicon Graphics Iris file format loader
This is an old format but my work requires I use them. Presagis Creator and the simulation industry still use these textures heavily.
There is no writer submitted, guessing that users would write out to a more modern format. 
Tested by downloading the free version of Presagis Creator, which comes with a gallery of textures. A test application loaded all the textures from 
C:\Presagis\Suite16\Creator\gallery\Textures
and wrote them as PNG. Then the results were viewed in the Windows Explorer by showing large icons.

Note, this is my first pull request ever.

### DIFF
--- a/stb_image.h
+++ b/stb_image.h
@@ -5758,29 +5758,29 @@ static int stbi__sgi_test(stbi__context *s) {
     return r == 0x01DA;
 }
 
-static void stbi__sgi_rle_decode(stbi_uc *dst, int dst_length, stbi_uc *src, int rle_length) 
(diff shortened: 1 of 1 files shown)


nothings:
Sorry to reject your first ever pull request, but we do not accept new image formats for stb_image, and have not for more than five years. You can of course maintain a fork yourself with this added, and likely it won't be too hard to merge fixes from this repo to your fork.

PR #1827, opened 2025-08-07
stb_image:  fix BMP HeaderV3 incorrect parse

yixy-only:
The size of HeaderV3 is 56. Compared with the 40 bytes of HeaderV1, the additional 16 bytes are 4 color masks. The original code accidentally discarded these 16 bytes, resulting in the loss of mask information, resulting in the "bad BMP" error.

After testing, the modified code runs well!

image


yixy-only:
This is a BMP image using HeaderV3 and can be used for testing. 
[image-v3.bmp](https://github.com/user-attachments/files/22000926/image-v3.bmp)

PR #1828, opened 2025-08-11
convert 16-bit pnms from big- to platform-endian

wkjarosz:
pnm files are big-endian, so while 8-bit files work fine, 16-bit files require converting from big- to native-endian.

This PR applied this conversion directly after loading the pixel data in `stbi__pnm_load`.

You can test on these two files (an 8-bit and 16-bit version), which now both display correctly:
[8-and-16-bit-test.zip](https://github.com/user-attachments/files/21717541/8-and-16-bit-test.zip)


wkjarosz:
I see now that there https://github.com/nothings/stb/pull/1710 fixes the same bug. My apologies for not searching before posting.

Issue #1838, opened 2025-09-14
Security Vulnerability Report - Request for Private Contact


Toddkk02:
Hello Sean,

I am a security researcher and have discovered multiple critical memory corruption vulnerabilities in stb_image.h through systematic fuzzing analysis.

The vulnerabilities include:
- Double free conditions (CWE-415) 
- Segmentation faults with heap corruption
- Reproducible crashes with minimal test cases

I have prepared complete proof-of-concept code and detailed technical analysis.

I understand this project discusses security issues publicly, but given the critical nature of these vulnerabilities, what is your preferred method for sharing the technical details?

I'm committed to responsible disclosure and can provide all information immediately.

Thank you for maintaining this valuable library.

Best regards

nothings:
This project discusses security issues publically. Like right here in public issues on github. No matter how "critical" the vulnerability.

Toddkk02:
i edited the security issue thanks for the response

Toddkk02:
Hi @NBickford-NV,

Found a double-free in the GIF loader during fuzzing with AFL++. Here's the reproduction:

PoC: [poc.c](https://github.com/user-attachments/files/22380345/poc.c)
Trigger: [crash_trigger.gif](https://github.com/user-attachments/assets/000e388f-3edd-409b-9253-83046ac80317)

The bug hits in `stbi__load_gif_main_outofmem` at line 6958. AddressSanitizer shows memory gets allocated at line 7010, freed via realloc at 6994, then freed again in the error handler.

60-byte trigger file makes it easy to reproduce. Compile with `-fsanitize=address -lm` and run `./poc crash_trigger.gif`.

Let me know if you need any additional details for the fix.

NBickford-NV:
Actually, looks like in my patched branch I get a double-free on `delays`! JarLob's other patch added `if (delays && *delays) *delays = 0;` to some of the other paths but not this one.

PR #1839, opened 2025-09-18
Fixes the first stbi_load_gif issue in #1838 by clearing `delays` on all outofmem paths

NBickford-NV:
After calling `stbi_load_gif`, apps should free the `delays` pointer it passed in if `delays` is non-null. This means that after freeing `delays` in the `stbi__load_gif_main_outofmem` error path, stb_image_load should also set `delays` to `NULL`. Otherwise if we allocate `delays` once but we get to `stbi__load_gif_main_outofmem` on a later frame, then the app sees a non-null `delays` pointer and tries to free it again.

With this change, running `clang++ poc.c -fsanitize=address && ./a.out 490442704-000e388f-3edd-409b-9253-83046ac80317.gif` from #1838 now correctly reports "Failed to load GIF" instead of segfaulting.

Forks with JarLob's double-free fix in #1549 were also affected; it implemented this logic on some paths, but not all, and #1838 happens to reach one of the other paths. Moving the `delays = NULL` logic to `stbi__load_gif_main_outofmem` on those forks allows us to save a few lines of code.

In addition, since `realloc` first frees the input pointer ([C specification](https://www.open-std.org/jtc1/sc22/wg14/www/docs/n1570.pdf) 7.22.3.5 item 1), we shouldn't pass the input to `stbi__load_gif_main_outofmem` after a realloc -- then we'd get a different double-free. If we remove the temporary pointer and do `p = realloc(p, ...)` then I think we get the right behavior and save a few lines of code.

Thanks!

### DIFF
--- a/stb_image.h
+++ b/stb_image.h
@@ -6956,7 +6956,10 @@ static void *stbi__load_gif_main_outofmem(stbi__gif *g, stbi_uc *out, int **dela
    STBI_FREE(g->background);
 
    if (out) STBI_FREE(out);
-   if (delays && *delays) STBI_FREE(*delays);
+   if (delays && *delays) {
+      STBI_FREE(*delays);
+      *delays = NULL;
+   }
    return stbi__errpuc("outofmem", "Out of memory");
 }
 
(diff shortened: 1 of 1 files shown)


NBickford-NV:
My colleague Pyarelal Knowles noticed that I made a mistake in simplifying the calls to `realloc` in `stbi__load_gif_main` and accidentally added a memory leak; if `realloc` returns `NULL` because it failed, then C specification 7.22.3.5 item 3 says it doesn't free the input pointer:

> If memory for the new object cannot be
> allocated, the old object is not deallocated and its value is unchanged.

I've reverted my changes to `stbi__load_gif_main` to the previous version, which I think was correct, so now the only change is we clear `delays` in outofmem. Tagging @sezero since this PR was integrated into SDL_image in https://github.com/libsdl-org/SDL_image/commit/658e4423772cef1b552432547691e08387002dd9 .

PR #1847, opened 2025-10-20
stb_ds: add arrclear

txgk:
I build my projects with compiler warnings enabled by default, so when I use `arrsetlen(a,n)` with `n=0`, I get this:

```c
stb_ds.h:539:50: warning: comparison of unsigned expression in ‘< 0’ is always false [-Wtype-limits]
  539 | #define stbds_arrsetlen(a,n)   ((stbds_arrcap(a) < (size_t) (n) ? stbds_arrsetcap((a),(si...
      |                                                  ^
stb_ds.h:407:21: note: in expansion of macro ‘stbds_arrsetlen’
  407 | #define arrsetlen   stbds_arrsetlen
      |                     ^~~~~~~~~~~~~~~
test.c:7:9: note: in expansion of macro ‘arrsetlen’
    7 |         arrsetlen(numbers, 0);
      |         ^~~~~~~~~
```

I understand that this warning is only relevant for the `n=0` case (warning description makes it pretty clear) and it's not dangerous at all, but it's a disturbing warning output which comes up on every clean build nevertheless.

I'd really want to avoid having to make exceptions for such warnings with pragmas, compiler arguments and what not.

So I've come up with `arrclear` macro. It does what it says, just like in STL's `vector`; and it doesn't raise the warning.

If you feel me on this one, consider having this macro in the `stb_ds.h`, otherwise never mind :^)

Best regards, Grigory

### DIFF
--- a/stb_ds.h
+++ b/stb_ds.h
@@ -136,6 +136,11 @@ DOCUMENTATION
           Changes the length of the array to n. Allocates uninitialized
           slots at the end if necessary.
 
+      arrclear:
+        void arrclear(T* a);
+          Changes the length of the array to 0. Doesn't destroy the allocated
+          elements if any are present. Equivalent to arrsetlen(a, 0).
+
       arrsetcap:
         size_t arrsetcap(T* a, int n);
           Sets the length of allocated storage to at least n. It will not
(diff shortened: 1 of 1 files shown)


nothings:
duplicate of #1175, but a good reminder I need to merge that

minikratka-ai (approved):

Issue #1844, opened 2025-10-12
stb_image_resize2: vertical_filter_kernel appears to be incorrectly set 


TianZerL:
**Describe the bug**
In the provided [code snippet](https://github.com/nothings/stb/blob/fede005abaf93d9d7f3a679d1999b2db341b360f/stb_image_resize2.h#L7836), the `stbir__set_sampler` call for the vertical sampler incorrectly uses `resize->horizontal_filter_kernel` instead of `resize->vertical_filter_kernel`. This likely results in the vertical filter kernel not being applied as intended, potentially causing incorrect image scaling behavior in the vertical direction.

**Expected behavior**
The vertical sampler should use `resize->vertical_filter_kernel` to ensure the correct kernel is applied during vertical scaling. The horizontal and vertical filters should operate independently with their respective configured kernels.

Issue #1841, opened 2025-09-29
[stb_image_resize2] Assertion failure in stbir_resize_extended_split when processing certain scale factors


hgarrereyn:
Hi, I encountered an internal assertion when using the `stb_image_resize2` API in a way that is legal as far as I can tell.

Specifically, the following testcase sets up a new resize operation from `16x9` -> `2x11`. It then invokes `stbir_build_samplers_with_splits` to prepare a multithreaded resize operation, and tries to run the last split first with `stbir_resize_extended_split`. However, this hits an internal assertion:
```
Assertion "in_first_scanline >= split_info->ring_buffer_first_scanline" in /fuzz/install/include/stb_image_resize2.h:6304
``` 

While this testcase is single threaded, it seems likely that in the intended usage (invoking `stbir_resize_extended_split` in multiple threads) could lead to a race here triggering this bug only sometimes.

(found via automated fuzzing)

See the following testcase:

**testcase.cpp**
```cpp
#include <cstddef>
#include <cstdint>
extern "C" {
#include "/fuzz/install/include/stb_image_resize2.h"
}
int main(){
    // Input/Output: RGB uint8, downscale W (16->2), upscale H (9->11)
    STBIR_RESIZE *resize = new STBIR_RESIZE();
    int in_w=16, in_h=9, out_w=2, out_h=11;
    stbir_resize_init(resize, nullptr, in_w, in_h, 0,
                              nullptr, out_w, out_h, 0,
                              STBIR_RGB, STBIR_TYPE_UINT8);

    // Provide buffers
    size_t in_bytes  = (size_t)in_w * in_h * 3;
    size_t out_bytes = (size_t)out_w * out_h * 3;
    unsigned char *in_buf  = new unsigned char[in_bytes];
    unsigned char *out_buf = new unsigned char[out_bytes];
    for (size_t i=0;i<in_bytes;i++) in_buf[i] = (unsigned char)i;
    stbir_set_buffer_ptrs(resize, in_buf, 0, out_buf, 0);

    // Build with splits and then execute only a later split (start=last, count=1)
    int splits = stbir_build_samplers_with_splits(resize, 4);
    if (splits >= 2) {
        int start = splits - 1;
        int count = 1;
        // This call aborts with: assertion
        //   in_first_scanline >= split_info->ring_buffer_first_scanline
        // inside stbir__vertical_gather_loop
        stbir_resize_extended_split(resize, start, count);
    }

    delete[] in_buf;
    delete[] out_buf;
[...]

PR #1849, opened 2025-10-22
stb_image_resize 2.16 - fixed the easy-to-use APIs to allow inverted bitmaps (negative strides), fix vertical filter kernel callback, fix threaded gather buffer priming (and assert)

jeffrbig2:


jeffrbig2:
Fixes #1844 #1824 #1841

Issue #1860, opened 2025-11-25
Heap Buffer overflow in stbi__convert_format16


kirit1193:
**Description**
There seems to be a heap buffer overflow in stb_image when loading 16-bit PNGs with channel conversion. If you load a grayscale PNG and request RGBA output. A 23171x23171 grayscale PNG requesting 4-channel output overflows to ~190KB instead of 4GB, then the conversion loop writes way past the buffer.

The existing size check at line 5121 doesn't catch this because it checks the source channel count, not the requested output channels.

**To Reproduce**
```c
#include <stdio.h>
#define STB_IMAGE_IMPLEMENTATION
#include "stb_image.h"

int main(int argc, char** argv) {
    int w, h, c;
    unsigned short *img = stbi_load_16(argv[1], &w, &h, &c, 4);  // request RGBA
    if (!img) return 1;
    stbi_image_free(img);
    return 0;
}
```

```
gcc -fsanitize=address test.c -o test
./test poc-004.png
```

ASAN output:
```
==14637==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x7ed40212604c
WRITE of size 2 at 0x7ed40212604c thread T0
    #0 stbi__convert_format16 stb_image.h:1837
    #1 stbi__do_png stb_image.h:5280
    #2 stbi__png_load stb_image.h:5299
    #3 stbi__load_main stb_image.h:1169
    #4 stbi_load_16 stb_image.h:1232

0x7ed40212604c is located 4 bytes after 194632-byte region allocated here:
    #0 malloc
    #1 stbi__malloc stb_image.h:987
    #2 stbi__convert_format16 stb_image.h:1820
```

Attaching the poc with a txt extension.

[poc-004.txt](https://github.com/user-attachments/files/23741089/poc-004.txt)

Issue #1529, opened 2023-10-19
Integer overflow in `stbi__convert_8_to_16`


JarLob:
`stbi__malloc` in [`stbi__convert_8_to_16`](https://github.com/nothings/stb/blob/5736b15f7ea0ffb08dd38af21067c314d6a3aae9/stb_image.h#L1204-L1218) [1] may overflow. However for successful exploitation `img_len` must be bigger than zero [2]. Any big enough `img_len` multiplied by 2 and casted to `size_t` on a 64 bit platform results in an unsigned number not smaller than the `img_len`.

```cpp
static stbi__uint16 *stbi__convert_8_to_16(stbi_uc *orig, int w, int h, int channels)
{
   int i;
   int img_len = w * h * channels;
   stbi__uint16 *enlarged;

   enlarged = (stbi__uint16 *) stbi__malloc(img_len*2); // [1] int overflow
   if (enlarged == NULL) return (stbi__uint16 *) stbi__errpuc("outofmem", "Out of memory");

   for (i = 0; i < img_len; ++i) // [2]
      enlarged[i] = (stbi__uint16)((orig[i] << 8) + orig[i]); // replicate to high and low byte, maps 0->0, 255->0xffff

   STBI_FREE(orig);
   return enlarged;
}
```

#### Impact

It doesn't look like a potential security issue, but the signed integer overflow behavior is undefined according to C/C++ standard.

ericoporto:
@JarLob could add somewhere like in the title of these or the PRs WHAT stb libraries you are contributing the code/issue to? I only use stb_vorbis from here (and indirectly through SDL_Sound) and am having a hard time differentiating from all issues you opened - in the same day!

PR #1862, opened 2025-11-25
Add fix for https://github.com/nothings/stb/issues/1860 by checking malloc multiplication before allocating

NBickford-NV:
This should fix https://github.com/nothings/stb/issues/1860 .

Previously, `stbi__convert_format16` didn't check if the multiplication `req_comp * x * y * 2` would overflow before calling `stbi__malloc`. Thankfully, the fix is a quick change to use `stbi__malloc_mad4` instead of a raw `stbi__malloc`. (`stbi__convert_format8` already checked this. The floating-point path checks this inside `stbi__hdr_load`.)

I also added a safety check to code in `stbi__convert_8_to_16` that looks like it could run into a similar issue. (We know that `req_comp * x * y` is valid here because we've been passed a valid buffer. That does imply that we could save a few operations by doing an `stbi__malloc_mad2` inside `stbi__convert_format16`, though.)

Thanks!

### DIFF
--- a/stb_image.h
+++ b/stb_image.h
@@ -1035,7 +1035,7 @@ static int stbi__mad3sizes_valid(int a, int b, int c, int add)
 }
 
 // returns 1 if "a*b*c*d + add" has no negative terms/factors and doesn't overflow
-#if !defined(STBI_NO_LINEAR) || !defined(STBI_NO_HDR) || !defined(STBI_NO_PNM)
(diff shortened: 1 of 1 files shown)


NBickford-NV:
(It looks like the fix in `stbi__convert_8_to_16` is another way of fixing https://github.com/nothings/stb/issues/1529; @JarLob previously wrote https://github.com/NBickford-NV/stb/commit/f100bfc302c0e095856c71a174714cce0a22e30a .)

Issue #1861, opened 2025-11-25
Out-of-bounds read in PNG palette expansion


kirit1193:
**Description**
stbi__expand_png_palette() doesn't validate that pixel indices are within the palette size. With this input, the palette buffer is stack-allocated at 1024 bytes, so it doesn't crash - it just reads uninitialized stack memory and puts it in the output pixels.

Basically a tiny PNG that leaks whatever is on the stack into the decoded image data.

**To Reproduce**
```c
#include <stdio.h>
#define STB_IMAGE_IMPLEMENTATION
#include "stb_image.h"

int main(int argc, char** argv) {
    int w, h, c;
    unsigned char *img = stbi_load(argv[1], &w, &h, &c, 0);
    if (!img) return 1;

    int leaked = 0;
    for (int i = 0; i < w * h * c; i++)
        if (img[i] != 0x00) leaked++;  // palette only has black

    printf("Non-zero bytes: %d / %d\n", leaked, w * h * c);
    stbi_image_free(img);
    return 0;
}
```

```
gcc test.c -o test
./test poc-005.png
```

Output:

```
Non-zero bytes: 150 / 300
```

Half the pixels contain leaked stack data. Valgrind confirms:

```
==4119== Conditional jump or move depends on uninitialised value(s)
==4119==  Uninitialised value was created by a stack allocation
==4119==    at 0x11BC1E: stbi__parse_png_file (stb_image.h:5079)
==4119== ERROR SUMMARY: 150 errors from 1 contexts
```

Attaching the png that triggers this in txt format:

[poc-005.txt](https://github.com/user-attachments/files/23741355/poc-005.txt)

NBickford-NV:
Thanks @kirit1193 ! I've put together a potential fix for this one in https://github.com/nothings/stb/pull/1863 . Looks like AddressSanitizer doesn't catch this, but MemorySanitizer does -- if you get the chance to run a fuzzer with MemorySanitizer enabled, it would be much appreciated by me.

PR #1863, opened 2025-11-25
Fixes issue #1861 by zero-initializing `palette`.

NBickford-NV:
In Issue #1861 , out-of-bounds palette indices that are greater than or equal to `pal_len` read from uninitialized elements of `palette` in `stbi__expand_png_palette`. This is an issue since data that was written to the stack previously could include sensitive info.

As one approach to fixing this, this merge request zero-initializes `palette` (and `tc16` as well, since I think it might be possible for something similar to happen in `stbi__compute_transparency16` if `img_out_n` isn't equal to `img_n`), so that these reads return 0 instead of uninitialized data. (We know that we can't read past the end of `palette` in `stbi__expand_png_palette` since `palette` has length 1024, but `*palette` has type `stbi_uc`, so the maximum element we can read is `(1 << 8) * 4 + 3 == 1023`.)

A different approach would be to add a check that each element of the palette is less than `pal_len` and exit early if so. However, my guess is the cost of a 1024-element memset is probably less on average than checking the palette index of every pixel, even if the branches in the latter fix were always predicted correctly, because it prevents vectorization inside `stbi__expand_png_palette`.

Thank you!

### DIFF
--- a/stb_image.h
+++ b/stb_image.h
@@ -5077,9 +5077,9 @@ static void stbi__de_iphone(stbi__png *z)
 
 static int stbi__parse_png_file(stbi__png *z, int scan, int req_comp)
 {
-   stbi_uc palette[1024], pal_img_n=0;
+   stbi_uc palette[1024]={0}, pal_img_n=0;
    stbi_uc has_trans=0, tc[3]={0};
-   stbi__uint16 tc16[3];
+   stbi__uint16 tc16[3]={0};
    stbi__uint32 ioff=0, idata_limit=0, i, pal_len=0;
    int first=1,k,interlace=0, color=0, is_iphone=0;
    stbi__context *s = z->s;


iliaal:
+1 — hit by [iliaal/fastchart](https://github.com/iliaal/fastchart) 1.1.1, a PHP extension that rasterizes user-supplied SVG through plutosvg, which vendors stb_image. Paletted PNGs via data: URIs reach `stbi__parse_png_file` with attacker-controllable pixel indices; without this fix, out-of-range indices leak uninitialized stack bytes into the renderable output that the PHP caller then downloads (info-leak primitive). fastchart now carries the same two-line patch locally; happy to drop it once this lands.

The reasoning about memset vs branch cost in the PR description matches what I saw in practice — the cost is in the noise even on tight inner loops, and the defensive-init shape stays safe under future refactors that might regress an index check.

Issue #1888, opened 2026-02-07
stb_vorbis floor1 reserves too much memory


lassade:
Looking at libvorbis file [floor1.c](https://github.com/xiph/vorbis/blob/2d79800b6751dddd4b8b4ad50832faa5ae2a00d9/lib/floor1.c#L152) and [backends.h:57](https://github.com/xiph/vorbis/blob/2d79800b6751dddd4b8b4ad50832faa5ae2a00d9/lib/backends.h#L57) the total amount of values can't be bigger than VIF_POSIT+2 = 65 but Xlist and related allocates max_partitons_count * max_dim_size + 2 or 31 * 8 + 2 = 250.

Another small things
- partition_class_list uses 32 instead of 31
- values could be a uint8

PR #1889, opened 2026-02-14
Small improvements an fixes on stb_vorbis

lassade:
- Reduces the upper bounds of some arrays across the code as explaned here #1888 as well some other ones
- Fixes compiler bug when builtind with `-Werror` on line 1404, the statement `f->stream_start + loc < f->stream_start` is always false
- Fixes `max_part_read` when computing how much temporary memory is needed, it's a non issue because `imdct_mem` is always bigger than `classify_mem` by a lot I don't see a cenario when it doesn't. the fix is pretty easy: the max `actual_size` is `f->blocksize_1` and not `f->blocksize_1/2` this happens on residue type 2 shown at `decode_residue` line 2111

### DIFF
--- a/stb_vorbis.c
+++ b/stb_vorbis.c
@@ -34,8 +34,10 @@
 //    github:audinowho   Dougall Johnson     David Reid
 //    github:Clownacy    Pedro J. Estebanez  Remi Verschelde
 //    AnthoFoxo          github:morlat       Gabriel Ravier
+//    github:lassade
 //
 // Partial history:
+//    1.23    - 2021-07-11 - various small fixes
 //    1.22    - 2021-07-11 - various small fixes
 //    1.21    - 2021-07-02 - fix bug for files with no comments
 //    1.20    - 2020-07-11 - several small fixes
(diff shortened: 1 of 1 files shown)


stb_vorbis.c
```diff
@@ -719,9 +721,9 @@ typedef struct
    uint8 class_subclasses[16]; // varies
    uint8 class_masterbooks[16]; // varies
    int16 subclass_books[16][8]; // varies
-   uint16 Xlist[31*8+2]; // varies
-   uint8 sorted_order[31*8+2];
-   uint8 neighbors[31*8+2][2];
+   uint16 Xlist[65]; // varies
+   uint8 sorted_order[65];
+   uint8 neighbors[65][2];
```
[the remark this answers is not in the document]
lassade: I don't know why but it is in the [spec](https://xiph.org/vorbis/doc/Vorbis_I_spec.html#x1-1010007.2.2) in at the last paragraph it says "Vector [floor1_x_list] is limited to a maximum length of 65 elements;"

stb_vorbis.c
```diff
@@ -4161,7 +4163,7 @@ static int start_decoder(vorb *f)
       int i,max_part_read=0;
       for (i=0; i < f->residue_count; ++i) {
          Residue *r = f->residue_config + i;
-         unsigned int actual_size = f->blocksize_1 / 2;
+         unsigned int actual_size = f->blocksize_1;
```
AliceLR, line 4166: Duplicate of #1487

lassade: this loop looks for the maximum ammount of memory needed, in my test data I don't have a single vorbis file that doesn't have a residue type 2 so using blocksize_1 is good enough

PR #1902, opened 2026-03-06
Fix implicit conversion float to double

bjornwah:
error: implicit conversion from 'float' to 'double' to match other operand of binary expression [-Werror=double-promotion]


bjornwah:
This macro seems only to be called with floating point args.

Would it be better to make it a function? Macros are a bit fuzzy with regard to typing.

PR #1927, opened 2026-04-03
Fix multiple memory safety issues in stb_image and stb_vorbis

boergeson:
Fixes for several integer overflow, out-of-bounds, and logic bugs found during a security review.

**stb_image.h:**
- Add overflow checks to `stbi__convert_16_to_8` and `stbi__convert_8_to_16` (lines 1193, 1209)
- Use `stbi__malloc_mad4` in `stbi__convert_format16` -- the 8-bit sibling was already using `stbi__malloc_mad3` but the 16-bit version was missed (line 1820)
- Fix `two_back` pointer in animated GIF loader: `out - 2*stride` points before the buffer, should be `out + (layers-2)*stride` (line 7023). Crashes under ASAN.
- Validate palette index in `stbi__expand_png_palette` -- the palette length was being ignored via `STBI_NOTUSED(len)`, allowing out-of-bounds reads from uninitialized stack memory (line 4988)
- Fix `stbi__process_marker` failure returning 1 (success) instead of 0 (line 3448)

**stb_vorbis.c:**
- Fix sign error in bounds clamp in `codebook_decode_deinterleave_repeat`: `- c_inter` should be `+ c_inter` (line 1891)

PoCs and a full writeup are in the `poc/` directory on this branch. See the companion issue for the complete list of findings.


AliceLR:
The stb_vorbis.c change is a duplicate of #1490.

Issue #1928, opened 2026-04-03
many memory safety bugs in stb_image and stb_vorbis


boergeson:
I did  look into  stb_image.h v2.30 and stb_vorbis.c v1.22 and I found a of integer overflows,  out of bounds accesses,  and logic errors .. I have PoCs for many of them and submitted a PR with fixes for the ones that are easy fix and you can REPRODUCE them if you check the fork

## abt the PoCs

there is working PoCs for bugs 1-12 in a fork: https://github.com/boergeson/stb/tree/fix/memory-safety-issues/poc to check them out

palette leak (bug 5) is the most visual -- you can see the stack data in the output pixel values. The GIF two_back (bug 3) crashes immediately under ASAN

## PR

I submitted a fix PR for bugs 1, 2, 3, 5, 8, and 12: https://github.com/nothings/stb/pull/1927

# the bugs

### 1. stbi__convert_format16 missing overflow check (line 1820)

The 8-bit format converter was hardened with `stbi__malloc_mad3`:
```c
// line 1763 -- safe
good = (unsigned char *) stbi__malloc_mad3(req_comp, x, y, 0);
```

But the 16-bit version still uses a raw malloc:
```c
// line 1820 -- no overflow check
good = (stbi__uint16 *) stbi__malloc(req_comp * x * y * 2);
```

With large dimensions (e.g. 32768x32768 RGBA), `req_comp * x * y * 2` overflows to 0. The subsequent conversion loop writes 8 GB into a zero-size buffer.

Reachable via 16-bit PNG, PSD, or PNM with a requested component  count different from source

### 2. stbi__convert_16_to_8 / stbi__convert_8_to_16 (lines 1193, 1209)

It's same pattern: `w * h * channels` is computed as a plain int multiplication with no overflow check, then passed to stbi__malloc. The convert_8_to_16 version also multiplies by 2, doubling the overflow risk.

### 3. Animated GIF two_back pointer underflow (line 7023)

```c
two_back = out - 2 * stride;
```

`out` is start of the allocated buffer. `out - 2 * stride` points before heap allocation. When frame useing disposal method 3, line 6818 reads from this pointer:

```c
memcpy( &g->out[pi * 4], &two_back[pi * 4], 4 );
```

crashes Under AddressSanitizer,, correct formula should be `out + (layers - 2) * stride`.

PoC: a 3-frame animated GIF with disposal method 3 on the third frame

### 4. Animated GIF layers*stride overflow (lines 6991-7021)

[...]

PR #1954, opened 2026-05-21
stb_image: zero-init paletted-PNG palette buffer

iliaal:
`stbi__parse_png_file` declares `palette[1024]` on the stack. The PLTE chunk fills the first `pal_len*4` bytes. `stbi__expand_png_palette` then reads `palette[idx*4 .. idx*4+4]` for every pixel index in `[0,255]` without bounds-checking `idx` against `pal_len`. A paletted PNG referencing `idx >= pal_len` reads uninitialized stack into the output raster (return addresses, canaries, prior locals). Bug #5 in #1928.

```diff
-   stbi_uc palette[1024], pal_img_n=0;
+   stbi_uc palette[1024]={0}, pal_img_n=0;
```

Complements #1927, doesn't compete with it. #1927 adds `idx >= len` checks in `stbi__expand_png_palette` and rejects malformed PNGs. This PR zero-inits the buffer instead. With #1927 alone, out-of-range indices error out. With this alone, they render black. With both, the leak is gone even if the index check regresses later. Cost: one C99 brace-init at function entry; the 1024 stack bytes were already there. Matches the existing `tc[3]={0}` style two lines down.

Repro: PoC at `poc/poc_palette_leak.c` from #1927. Without this patch, 161/255 pixels contain stack bytes. With it, those pixels are zero.

Hit by [fastchart](https://github.com/iliaal/fastchart) 1.1.1, a PHP extension that rasterizes user SVG through plutosvg, which vendors stb_image. fastchart 1.1.1 carries this as a local patch; sending upstream so plutosvg and other consumers pick it up.

v2.30, master HEAD. No API change. No allocator paths touched.

### DIFF
--- a/stb_image.h
+++ b/stb_image.h
@@ -5077,7 +5077,7 @@ static void stbi__de_iphone(stbi__png *z)
 
 static int stbi__parse_png_file(stbi__png *z, int scan, int req_comp)
 {
-   stbi_uc palette[1024], pal_img_n=0;
+   stbi_uc palette[1024]={0}, pal_img_n=0;
    stbi_uc has_trans=0, tc[3]={0};
    stbi__uint16 tc16[3];
    stbi__uint32 ioff=0, idata_limit=0, i, pal_len=0;


NBickford-NV:
I think this is a subset of https://github.com/nothings/stb/pull/1863 .

iliaal:
@NBickford-NV right, this is a strict subset of #1863 (same single-char `palette[1024]={0}` change; #1863 also includes `tc16[3]={0}` for the 16-bit transparency case, which I missed). My search missed it before posting. Closing here in favour of #1863.

iliaal:
Closing in favour of #1863 (linked above).

PR #1957, opened 2026-05-30
Fix for GCC 3.0 and below in C99 mode

Un1q32:
GCC 3.0 and 2.95 hadn't implemented this particular part of the C99 standard, so if you try to use stb_ds along with `-std=c9x` you'll get cryptic errors about lvalues.  This just makes those particular compilers fallback to the C89 path which works fine.


Un1q32:
Closed for the same reason as #1966

PR #1959, opened 2026-06-02
stb_image_write: fix linker errors in HDR image functions when STBI_WRITE_NO_STDIO is defined

sanjay-nambiar:
Calling any of the HDR image write functions with `STBI_WRITE_NO_STDIO` defined results in an undefined symbol linker error.  This is because all the HDR write function definitions are guarded with `STBI_WRITE_NO_STDIO` macro when only the function that writes directly to a file - `stbi_write_hdr` should be under the include guard.

Fixing this requires removing one usage of `sprintf` used when writing an HDR image header. Implemented a simple unsigned int to string function to replace  `sprintf`, borrowing code from `stb_sprintf`.

### DIFF
--- a/stb_image_write.h
+++ b/stb_image_write.h
@@ -141,6 +141,7 @@ CREDITS:
       github:ignotion
       Adam Schackart
       Andrew Kensler
+      Sanjay Nambiar
 
 LICENSE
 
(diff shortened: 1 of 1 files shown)


nothings:
Do not borrow from stb_sprintf; that function is not designed to be simple, but to be fast, and we don't need speed here. A much shorter function will do.

sanjay-nambiar:
Updated the function to be a much simpler one. Thanks!

Issue #1947, opened 2026-05-10
Heap buffer overflow via size_t-to-int truncation in stb_vorbis setup_malloc (codebook multiplicands)


Tomer-PL:
## Summary

A crafted 205-byte Ogg Vorbis file triggers a heap buffer overflow in `stb_vorbis.c` when processed by any application using stb_vorbis for Vorbis decoding. The root cause is an implicit `size_t`-to-`int` truncation when passing allocation sizes to the internal `setup_malloc` function.

**Affected:** stb_vorbis v1.22 (and likely earlier versions), any application linking stb_vorbis on 64-bit platforms.

## Root Cause

`setup_malloc` (line 950) takes `int sz`:

```c
static void *setup_malloc(vorb *f, int sz)
{
   sz = (sz+7) & ~7;
   // ...
   return sz ? malloc(sz) : NULL;
}
```

Callers compute sizes using `size_t` arithmetic (due to `sizeof` returning `size_t`), but the result is implicitly truncated to `int` at the function call boundary:

```c
// line 3880 (codebook multiplicands pre-expansion, lookup_type=1)
c->multiplicands = (codetype *) setup_malloc(f,
    sizeof(c->multiplicands[0]) * c->entries * c->dimensions);
```

The left-to-right multiplication promotes `c->entries` and `c->dimensions` (both `int`) to `size_t`, producing a correct `size_t` result. This `size_t` is then truncated to `int` when passed to `setup_malloc`.

## Exploit

A Vorbis codebook with `entries = 16,519,105` and `dimensions = 65`:

```
sizeof(float) * 16,519,105 * 65
= 4,294,967,300 (correct size_t)
= 4 (truncated int32: 4,294,967,300 mod 2^32)
```

After alignment: `malloc(8)` — space for 2 floats. The subsequent multiplicands expansion loop writes 31 floats (124 bytes) into this 8-byte buffer — a **116-byte heap overflow**.

The inner loop is limited to 31 iterations by the `div > UINT_MAX / lookup_values` check, but that's still enough for a significant overflow.

## Reproduction

```bash
# Using whisper.cpp as the test harness (any stb_vorbis consumer works)
# Build with ASan
cmake -B build -DCMAKE_C_FLAGS="-fsanitize=address" -DCMAKE_CXX_FLAGS="-fsanitize=address"
cmake --build build

# Run with the 205-byte crafted OGG
[...]

Issue #1168, opened 2021-07-15
stb_vorbis: Undefined behavior found by libFuzzer/UBSan.


AliceLR:
These issues cropped up during fuzzing sessions for libxmp. I considered submitting a patch but I'm not familiar enough with Vorbis to know how to handle the first and the second is fairly minor.

[OGG_lookup1_values.zip](https://github.com/nothings/stb/files/6821929/OGG_lookup1_values.zip)
```
5bb536ce97eb9f582be48fe6825d04f7f245f02c.ogg

src/depackers/vorbis.c:1244:14: runtime error: 2.14525e+155 is outside the range of representable values of type 'int'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior src/depackers/vorbis.c:1244:14 in 
src/depackers/vorbis.c:1248:14: runtime error: 2.14525e+155 is outside the range of representable values of type 'int'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior src/depackers/vorbis.c:1248:14 in 

825e51922407acfb71fbc40e804234be91dc5d91.ogg

src/depackers/vorbis.c:1244:14: runtime error: inf is outside the range of representable values of type 'int'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior src/depackers/vorbis.c:1244:14 in 
src/depackers/vorbis.c:1248:14: runtime error: inf is outside the range of representable values of type 'int'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior src/depackers/vorbis.c:1248:14 in 

etc...
```

[OGG_fast_ftoi.zip](https://github.com/nothings/stb/files/6821937/OGG_fast_ftoi.zip)
```
65537dda5e67e6d6aa4b032e70b45e90e7f1322d.ogg

src/depackers/vorbis.c:5171:15: runtime error: signed integer overflow: -1026200468 - 1136656384 cannot be represented in type 'int'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior src/depackers/vorbis.c:5171:15 in 

7530d375fa4fef8e627674df712762dd3a2ab5d0.ogg

src/depackers/vorbis.c:5171:15: runtime error: signed integer overflow: -1037235904 - 1136656384 cannot be represented in type 'int'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior src/depackers/vorbis.c:5171:15 in 

etc...
```

[...]

PR #1960, opened 2026-06-03
[stb_vorbis] Increase setup_malloc size parameter to uint64.

### LINKED ISSUE #1933: stb_vorbis: integer overflow in codebook multiplicands allocation
stb_vorbis.c:3880:
  setup_malloc(f, sizeof(float) * c->entries * c->dimensions)

entries and dimensions come from the vorbis stream. entries=1048577, dimensions=4096: product overflows int32, setup_malloc gets 16KB, loop writes 16GB.

Fix: check for overflow before allocation, cast to size_t.

AliceLR:
Marking as a draft for now since it's poorly tested, but this should fix #1933, #1947, and the second case in the comment https://github.com/nothings/stb/issues/1168#issuecomment-1082625780. However, it's not clear if this is the fix you actually want (particularly re: the definition of `uint64` and various type-corrections).

@nothings I've noticed several dubious submissions duplicating parts of bug reports and patches I submitted years ago. Is there anything I can do to help?

### DIFF
--- a/stb_vorbis.c
+++ b/stb_vorbis.c
@@ -640,13 +640,28 @@ enum STBVorbisError
 #define MAX_BLOCKSIZE_LOG  13   // from specification
 #define MAX_BLOCKSIZE      (1 << MAX_BLOCKSIZE_LOG)
 
-
+#if defined B_BEOS_VERSION
+   #include <SupportDefs.h>
+#elif defined __amigaos4__
+   #include <exec/types.h>
+#elif defined _arch_dreamcast /* KallistiOS */
+   #include <arch/types.h>
+#else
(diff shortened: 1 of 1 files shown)


AliceLR:
* I've replaced the simple `typedef` for `uint64` with a more robust version copied from [libxmp's common.h](https://github.com/libxmp/libxmp/blob/490bd1df55544ce6c328456be6dc6fb6ea0edc0c/src/common.h#L90-L115). This fragment is probably non-copyrightable.
* I removed the creeping type corrections and reduced this patch down to what should be the bare minimum to fix the affected `setup_malloc`, `setup_temp_malloc`, and `setup_temp_free` calls.

PR #1966, opened 2026-06-10
GCC 2.7 C++ fix

Un1q32:
G++ 2.7 isn't complete enough to support the template solution, but it also doesn't error on implicit pointer casting.  It will throw a warning, but at least it will compile.

### DIFF
--- a/stb_ds.h
+++ b/stb_ds.h
@@ -687,7 +687,7 @@ enum
    STBDS_SH_ARENA
 };
 
-#ifdef __cplusplus
+#if defined(__cplusplus) && (!defined(__GNUC__) || __GNUC__ >= 3 || (__GNUC__ == 2 && __GNUC_MINOR__ >= 8))
 // in C we use implicit assignment from these void*-returning functions to T*.
 // in C++ these templates make the same code work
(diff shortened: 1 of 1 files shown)


nothings:
GCC 2.7 from 1995?

Un1q32:
Yes

nothings:
Is this being actively used for development in some environment? There are dozens of versions of gcc and it would become a maintenance burden if they all caused minor tweaks.

Un1q32:
It's not being used like, daily, but I use it on occasion when playing with older systems.  Having to build a newer compiler on said systems would be annoying.  That being said, I can just carry the patch myself if you don't want it, it's not a big deal.

Un1q32:
For an idea of how many tweaks you would need, afaik this and https://github.com/nothings/stb/pull/1957 would cover all issues for all versions of GCC from 2.7 to latest (at least for stb_ds).  Both are one line changes.

nothings:
> For an idea of how many tweaks you would need, afaik this and #1957 would cover all issues for all versions of GCC from 2.7 to latest (at least for stb_ds). Both are one line changes.

There are around 17 libraries in this repository. Let's go with you carrying this patch for yourself.

PR #1971, opened 2026-06-30
stb_rect_pack: fix corner case in best-fit waste calculation

jadedpasta:
Under the best-fit heuristic, the waste calculation for a candidate x-coordinate may be incorrect when the initial node has `y == 0`. In this case, `node->y == min_y == 0` and the first loop iteration enters the `else` block, which does not account for the possibility of a reduced first-iteration `visited_width`.

To fix this, handle the first-iteration case (`visited_width == 0`) explicitly.

This case was found by comparing my own implementation vs stb_rect_pack on a randomized skyline. The test case that triggered the issue was the following:

```
width: 12, height: 10
nodes:
    .{ .x = 0, .y = 5 }
    .{ .x = 1, .y = 0 }
    .{ .x = 5, .y = 5 }
    .{ .x = 7, .y = 1 }
    .{ .x = 9, .y = 3 }
    .{ .x = 12, .y = 1073741824 }
input rectangle: .{ .w = 9, .h = 2 }
candidate xpos: 3

            =
            =
            =
   ---------=
   ---------=
+    ++     =
|    |+     =
|    |+  +++=
|    |+  |++=
|    |+++|++=
=============

```

### DIFF
--- a/stb_rect_pack.h
+++ b/stb_rect_pack.h
@@ -1,4 +1,4 @@
-// stb_rect_pack.h - v1.01 - public domain - rectangle packing
+// stb_rect_pack.h - v1.02 - public domain - rectangle packing
 // Sean Barrett 2014
 //
 // Useful for e.g. packing rectangular textures into an atlas.
@@ -38,9 +38,11 @@
 //  Bugfixes / warning fixes
 //    Jeremy Jaussaud
 //    Fabian Giesen
+//    Robert Hrusecky
 //
 // Version history:
 //
+//     1.02  (          )  fix corner case in best-fit waste calculation 
 //     1.01  (2021-07-11)  always use large rect mode, expose STBRP__MAXVAL in public section
 //     1.00  (2019-02-25)  avoid small space waste; gracefully fail too-wide rectangles
 //     0.99  (2019-02-07)  warning fixes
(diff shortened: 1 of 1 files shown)


nothings:
1. Was this bug report written by AI?
2. Why is the ASCII art broken?
3. What do you mean tested on a randomized skyline? Is it possible for stb_rect_pack.h to even generate a node with y=0 unless that node is last, or are you fixing a case that can never happen?

jadedpasta:
1. This was not written by AI. Sorry if the ASCII art gave that impression.
2. I'm not sure why you think it appears broken. I hand typed it out to match the set of nodes given. "+" and "|" represent the skyline, "=" represent the packing area boundaries, and "-" represents the candidate position.
3. Oh you're totally right, I guess this particular situation can't be generated by packing rectangles, since from an empty area and with the current available heuristics, rectangles should initially always get placed left to right at y=0. My code generates the skyline with random segments (not from packing).

I guess this is just a theoretical correction then, not an actual issue. However it does seem to point out that this algorithm is relying on a non-trivial assumption about what the skyline can be. Feel free to close if you don't care about it. Sorry for the noise.

nothings:
not a bug

PR #1984, opened 2026-08-02
update LLM rejection link

ruby-R53:
it's moved

* Do add your name to the list of contributors. (Don't worry about the formatting.) I'll try to remember to add it if you don't, but I sometimes forget as it's an extra step.


nothings:
Thank you!

PR #1988, opened 2026-08-19
stb_sprintf: treat a negative '*' precision as omitted

AbhinavMir:


AbhinavMir:
@nothings, could you review this?

